Issue #19, opened 2019-08-12
Finalize JDT LS Quarkus extension with ConfigRoot


angelozerr:
Finalize JDT LS Quarkus extension which collect Quarkus properties from ConfigRoot annotation:

 * [x] clean code & add javadoc comments
 * [x] manage map (basicly with {*})
 * [x] manage list
 * [x] write tests
 * [x] collect  enumeration (like enum)
 * [x] collect phase to show in the completion only the "runtime" properties
 * [x] collect javadoc

angelozerr:
Fixed with https://github.com/redhat-developer/quarkus-ls/commit/06a333b77934e08d365b184f92be90454394612c

PR #28, opened 2019-08-13
Finalize JDT LS Quarkus extension with ConfigRoot

angelozerr:
See #19


angelozerr:
This PR improves the compute of Quarkus properties by adding:

 * location (JAR)
 * get Javadoc
 * get phase
 * tests

Here a demo:

QuarkusCompletion2

xorye:
Great PR, the documentation is definitely more informative.

It seems like the newlines are inconsistent.

Example 1:
image
There is a newline in `quarkus-core-deployment-0.20.0.jar`, after `deployment-` and before `0.20.0.jar`.
There is a newline after "Source:"

Example 2:
image
There is a newline after "Type:", whereas Example 1 didn't.

However, I don't know if this is something we can easily control as it seems like if the value is too long, a newline gets inserted automatically? 

In addition, I noticed that for some properties, the descriptions end in a period ".", whereas some properties' descriptions end without a period. Example 1 ends with a period whereas Example 2 does not. For this, I think we should be consistent and always put a period, or not put one at all. What do you think?

angelozerr:
@xorye for newline, I'm afraid that we can do nothing -(

Please create an issue with that in order to discuss about this problem with @fbricon.

> Example 2 does not.

It's because of javdoc https://github.com/quarkusio/quarkus/blob/master/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSourceRuntimeConfig.java#L56

we cannot do nothing about thi sproblem.

PR #32, opened 2019-08-14
Fix hover test on Windows

angelozerr:
On Windows OS, the line separator is \r\n. This PR fix that. The hard coded \n comes from the Javadoc of Quarkus properties (see META-INF/quarkus-javadoc.properties files from JAR)


angelozerr:
@xorye please review this simple PR which fixes hover test on Windows OS

xorye:
I have tested them on Windows 10, and it works.

Issue #1, opened 2019-08-01
Completion support for key in application.properties


angelozerr:
Completion support for key in application.properties:

```
greeting.message = hello
gree | = 
```

 - [x] show Quarkus properties in completion
 - [x] open this completion kind only if it's a key (not a value)
 - [ ] eat the content until '=' is found when completion is applied
 - [x] snippet support to set the cursor after the =
 - [x] snippet choice support for enum and boolean
 - [x] show documentation by merging 
    * JavaDoc (as markdown if client support it)
    * source (class+field)
    * type (string, boolean, etc)
 - [x] add a QuarkusAssert class to test completion

PR #34, opened 2019-08-14
Snippet choice support for enum and boolean.

angelozerr:
Snippet choice support for enum and boolean.

See #1

### DIFF
--- a/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCompletions.java
+++ b/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCompletions.java
@@ -9,6 +9,9 @@
 *******************************************************************************/
 package com.redhat.quarkus.services;
 
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -18,6 +21,8 @@ import org.eclipse.lsp4j.CompletionList;
 import org.eclipse.lsp4j.InsertTextFormat;
 import org.eclipse.lsp4j.MarkupKind;
 import org.eclipse.lsp4j.Position;
+import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4j.TextEdit;
 import org.eclipse.lsp4j.jsonrpc.CancelChecker;
 
 import com.redhat.quarkus.commons.ExtendedConfigDescriptionBuildItem;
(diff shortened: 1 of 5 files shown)


angelozerr:
This PR fixes https://github.com/redhat-developer/vscode-quarkus/issues/25 and provide enum completion (only when key completion is applied).

QuakusCompletionChoixe

xorye:
Great PR, I very much like the text edits to deal with https://github.com/redhat-developer/vscode-quarkus/issues/25

I have some comments:


The snippet choice works great, but when I select one of the options (in this case, `BLOCK`), and then If I hit "Enter" to go to the next line, completion does not show up as I type. There is a grey highlight covering the text that I am typing. I think this highlight is preventing me from getting completion options.\
\
\

The grey highlight goes away when I hit "Tab" however. Once the grey highlight is gone, completion works as I type.\
\
\

However, despite having grey highlights, the completion options still show up if you invoke completion with "Ctrl + Space". The problem is that the grey highlights prevent completion options showing as you type.\
\
I'm not sure if this is something that can be fixed. Is this something we should be concerned about?

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCompletions.java
```diff
@@ -88,26 +98,62 @@ private static void collectPropertyKeySuggestions(TextDocument document, Propert
...
+			insertText.append(' '); // TODO: this space should be configured in format settings
+			insertText.append('=');
+			insertText.append(' '); // TODO: this space should be configured in format settings
+
+			if (enums != null) {
+				// Enumerations
+				if (snippetsSupported) {
+					SnippetsBuilder.choice(1, enums, insertText);
+				} else {
+
```
xorye: Just double checking, was there supposed to be something here?

angelozerr:
Thanks @xorye for your feedback, I will study more your issue.

fbricon:
When you hit enter, you're still in "snippet" mode in vscode, as you're inside a placeholder, in that mode, completion is not called. If you hit `Esc` after selecting your value, it'll work. Long story short, this is a VS Code problem, orthogonal to this PR.

angelozerr:
@fbricon can we merge this PR?

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCompletions.java
angelozerr: Good catch. I fix it and add tests for snippetSupport or not.

quarkus.ls/com.redhat.quarkus.ls/src/test/java/com/redhat/quarkus/services/ApplicationPropertiesHoverTest.java
```diff
@@ -26,35 +26,35 @@
  */
 public class ApplicationPropertiesHoverTest {
 
-	@Test
```
fbricon: why?

angelozerr: Sorry with my mistake, it's fixed.

Issue #26, opened 2019-08-13
Validate application.properties: unknown properties


angelozerr:
Validate application.properties

 * Quarkus properties which doesn't exists.

before doing this issue, the issue https://github.com/redhat-developer/quarkus-ls/issues/9 should be fixed to collect all properties (from JAR and sources).

Pay attention with map quarkus properties which looks like:

`quarkus.mymap."any.value" = `. The cache stores `quarkus.mymap.{*}` as key map and not  `quarkus.mymap."any.value"`

fbricon:
You might use application.properties to define non-quarkus properties so I'm not sure about this one. 
Might only need to validate the quarkus namespace

angelozerr:
> You might use application.properties to define non-quarkus properties so I'm not sure about this one.

Indeed perhaps it's a bad idea.

> Might only need to validate the quarkus namespace

If you are using @ConfigProperty there are not quarkus namespace (not managed for the moment, I must do that in https://github.com/redhat-developer/quarkus-ls/issues/30)

More it seems that there are some Quarkus properties which hard coded, see 
https://github.com/redhat-developer/quarkus-ls/issues/29

I think more and more it's a bad idea or perhaps set the unknown properties as warning?

PR #36, opened 2019-08-15
Add test diagnostic

### LINKED ISSUE #3: Validate application.properties: syntax error
Validate application.properties:

 * syntax error : ex : key name which have not '=' only if it's not inside a line comment.

Add CodeAction to add the  '='

xorye:
Fixes #3 

### What was implemented
This PR introduces a new diagnostic with message: "Missing '=' after key" and code: "missing_equals_after_key".

The error range covers the whole line:
image

There are multiple error ranges if there is more than one line with a missing equals sign:
image

Considers comments:
image

Code action in action:

### What was not implemented
This PR does not provide functionality allowing the client to toggle on/off validation. Validation is on by default.
This PR does not create an extensions registry like lsp4xml. An extensions registry might be helpful for implementing code actions.
This PR does not contain a file that defines error codes for error diagnostics like lsp4xml. 
I will implement these depending on feedback.

### How this PR detects a missing equals sign
For every line in `application.properties`, I continually call `PropertiesScannerUtils.getTokenType()` to determine all the tokens in a particular line. If a line contains a `KEY` token but does not contain an `EQUALS` token, I create a diagnostic. This is done all in the [`QuarkusDiagnostics.java`](https://github.com/xorye/quarkus-ls/blob/fd83e4fe5e4af39c5ea20658dd613e622db3e9a7/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusDiagnostics.java) file.

### Notes about the code action
The current code action just adds an equals sign at the end of the line with the error. This is only useful in the case where the line only contains a key. If the line contains both key and a value, but no equals sign, the code action places the equals sign at the wrong place:

Which is not what we want. If we wanted to place the equals sign in between the key and the value, we would have to determine which part is the key and which part is the value. 

However, it might just be easier for users if they added the missing equals sign themselves, since its very quick to do. A code action that places an equals sign at the wrong place, would only frustrate users.

### DIFF
--- a/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/ls/QuarkusLanguageServer.java
+++ b/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/ls/QuarkusLanguageServer.java
@@ -66,7 +66,7 @@ public class QuarkusLanguageServer implements LanguageServer, ProcessLanguageSer
 		serverCapabilities.setDocumentHighlightProvider(false);
 		serverCapabilities.setDocumentSymbolProvider(false);
 		serverCapabilities.setWorkspaceSymbolProvider(false);
-		serverCapabilities.setCodeActionProvider(false);
+		serverCapabilities.setCodeActionProvider(true);
 		serverCapabilities.setDocumentFormattingProvider(true);
 		serverCapabilities.setDocumentRangeFormattingProvider(true);
 		serverCapabilities.setRenameProvider(false);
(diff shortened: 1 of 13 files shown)


angelozerr:
@xorye it's a very cool PR, but when I see your work and other services (completion, hover) I think we should have a PropertyModel (like DOMDocument with XML) to clean all services.

I'm working on that https://github.com/redhat-developer/quarkus-ls/issues/37 My idea is to adapt my work on https://github.com/ec4j/ec4j/blob/master/core/src/main/java/org/ec4j/core/parser/EditorConfigParser.java for simple properties.

xorye:
I agree, I think #37 would be very helpful

angelozerr:
@xorye I would like to initialize validation by implementing https://github.com/redhat-developer/quarkus-ls/issues/26 In this initialisation, my idea is to loop for each property of the properties model and call validateProperty. We will able to manage any validation property to avoid looping for properties for each validation case. In other please wait for I implement https://github.com/redhat-developer/quarkus-ls/issues/26 to implement your validation case. Thanks!

angelozerr:
> @angelozerr If I understand correctly, your idea is to loop over each property in the properties model and call validateProperty, which validates for all validation cases?

Yes it's my idea, but we will see if there are no limitation with that when we will implement validation for other cases.

Issue #35, opened 2019-08-15
Documentation shows comment delimiters if Javadoc is on one line


fbricon:
Compare :
<img width="709" alt="Screen Shot 2019-08-15 at 8 48 34 PM" src="https://user-images.githubusercontent.com/148698/63118441-20f25080-bf9e-11e9-9e46-f4194204c054.png">

to :
<img width="591" alt="Screen Shot 2019-08-15 at 8 50 14 PM" src="https://user-images.githubusercontent.com/148698/63118526-5139ef00-bf9e-11e9-8ad1-63333e8e8533.png">

this is a regression as you can see from https://github.com/redhat-developer/quarkus-ls/pull/28#issuecomment-520873072

angelozerr:
Indeed i know this problem. For the moment i return just the javadoc by using jdt javadoc range if sources is available. My work is not finsished, i need to

* remove slash and star of javadoc
* concert javadoc on markdown if client support it

Please rename this issue like "improve javadoc extraction"

fbricon:
But this worked in https://github.com/redhat-developer/quarkus-ls/pull/28#issuecomment-520873072:
![image](https://user-images.githubusercontent.com/20326645/62951741-78cf6280-bdb8-11e9-8d61-d36611e98d3f.png)

angelozerr:
It worked because we are in case where we have not the sources. In this case docs comes from the quarkus-javadoc.properties file stored in the jar.

Issue #40, opened 2019-08-19
Get Javadoc from java sources or attached Javadoc.


angelozerr:
Today javadoc is extracted only from It extract Javadoc only from META-INF/quarkus-javadoc.properties

Javadoc should be extracted when it is available:

 * from Java sources (when JAR is linked to source)
 * from attached Javadoc

After extracting the Javadoc, the `/**` should be removed and HTML tag (ex : `<code>`, etc) 
 * should be removed if doc must be rendered as plaintext 
 * replace with markdown if doc mus b erenderer as markdown

PR #41, opened 2019-08-19
Get javadoc from only META-INF/quarkus-javadoc.properties

angelozerr:
This PR fixes https://github.com/redhat-developer/quarkus-ls/issues/35

It extract Javadoc only from META-INF/quarkus-javadoc.properties and not from java sources or attached Javadoc. To have a best renderer in markdown it should be better to transform HTML Javadoc to markdown. See issue https://github.com/redhat-developer/quarkus-ls/issues/40


fbricon:
any tests?

angelozerr:
Tests already exists. See https://github.com/redhat-developer/quarkus-ls/blob/master/quarkus.jdt/com.redhat.quarkus.jdt.test/src/main/java/com/redhat/quarkus/jdt/core/JDTQuarkusManagerTest.java#L43

PR #43, opened 2019-08-19
Extract Quarkus Properties for Map

### LINKED ISSUE #42: Extract Quarkus Properties for Map
Ex:

`quarkus.log.category."com.lordofthejars".level=DEBUG`

See https://lordofthejars.github.io/quarkus-cheat-sheet/#_logging

angelozerr:
Extract Quarkus Properties for Map

Fixes #42


angelozerr:
Here a demo with completion for config Map:

ConfigMapDemo

Please note that hover doesn't work for config map.

@fbricon do you want I support that in this issue?

fbricon:
we can add hover support in a separate PR.

According to https://lordofthejars.github.io/quarkus-cheat-sheet/#_logging, quarkus.log.category."key".level's default value should be `INFO`, but `inherit` is returned

angelozerr:
> According to https://lordofthejars.github.io/quarkus-cheat-sheet/#_logging, quarkus.log.category."key".level's default value should be INFO, but inherit is returned

Indeed and we should have choice for INFO, but as Level is not enum, it doesn't work.

For inherit, we could study the problem in an another issue, no? Could we merge this PR?

PR #68, opened 2019-09-02
Support completion for profile

### LINKED ISSUE #55: Support completion for profile
See https://quarkus.io/guides/application-configuration-guide#configuration-profiles for profil documentation.

Profil completion should be activated when properties starts with `%`.

There are 2 cases:

* completion is triggered before the `'.'` character (ex : `%de|`). In this case only list of profils should be returned:

  * dev
  * prod
  * test

What about custom profils (ex: `%staging`)? perhaps we could add it in `settings.json` ?

* completion is triggered after the `'.'` character (ex : `%dev.|`). In this case Quarkus properties should be returned with profil:

  * `%dev.quarkus.http.port`

angelozerr:
Support completion for profile

Fixes #55


angelozerr:
This PR manages profile extracation inside the properties model (see PropertyKey#getProfile()) which is more clean.

Completion is trigger with the following characters:

 * `%`
 * `.`
 * `=`

It provides completions:

* for profiles: dev, prod, test (ex : `%|`) and custom profiles (ex: staging). Custom profiles are retrieved by searching existings property from application.properties which defines a custom profile.
 * for Quarkus properties with profile (ex :` %dev.quarku|s` )

Here a demo:

ProfileDemo

fbricon:
Awesome work!

PR #83, opened 2019-09-11
Show extension name in Hover

### LINKED ISSUE #82: Show extension name in Hover
Show extension name in Hover

angelozerr:
Show extension name in Hover

Fixes #82


fbricon:
what would be the hover output if extensionname is null?

angelozerr:
> what would be the hover output if extensionname is null?

no case for the moment, but when source will be implemented we will have the case and in this case extension name is not displayed in hover.

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/utils/DocumentationUtils.java
```diff
@@ -71,19 +71,12 @@ public static MarkupContent getDocumentation(ExtendedConfigDescriptionBuildItem
...
-		String location = item.getLocation();
-		if (location != null) {
-			int slashIndex = location.lastIndexOf('/');
-			if (slashIndex != -1) {
-				location = location.substring(slashIndex + 1, location.length());
-			}
-			addParameter("Location", location, documentation, markdown);
+		// Extension name
+		String extensionName = item.getExtensionName();
+		if (extensionName != null) {
```
xorye: Can we also just do `addParameter("Extension", item.getExtensionName(), documentation, markdown);`, since `addParameter()` checks for null value [here](https://github.com/angelozerr/quarkus-ls/blob/5ef83b4f8e338d15200676567d05b58592060e4a/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/utils/DocumentationUtils.java#L97)?

angelozerr: Good catch! It should be fixed now.

angelozerr:
@fbricon do you think we could merge this PR?

PR #85, opened 2019-09-12
Collect Quarkus properties from Java source

### LINKED ISSUE #9: Collect Quarkus properties from Java source
Today Quarkus properties are collected from application library. It should be very good to collect the properties from the source.

The hard part is to track the change of Java source.

angelozerr:
Collect Quarkus properties from Java source

Fixes #9


angelozerr:
This PR requires the vscode-quarkus PR https://github.com/redhat-developer/vscode-quarkus/pull/79

Here a demo to test this feature:

QuarkusPropertiesFromJavaSourceDemo

xorye:
Very cool PR.

When I comment out/remove a ConfigProperty and save, I get duplicate completion items. If I do this multiple times, I get even more duplicate completion items. Is this a problem with this PR?
gif showing duplicate completion items

fbricon:
Yeah it's prolly caused by this PR

xorye:
Maybe this one is more of an enhancement request, but it would be nice if generic classes would display their paramter type in the "Type" section of the documentation.

For example, if `greeting.name` expected an `Optional<String>`, it would be nice if it was in the documentation. Currently, we only see `Optional`:
image

angelozerr:
> For example, if greeting.name expected an Optional<String>, it would be nice if it was in the documentation. Currently, we only see Optional:

I agree with you, I tried to support that but it seems with Java sources type are not resolved and when I consume [JDT JavaModelUtils](https://github.com/redhat-developer/quarkus-ls/blob/master/quarkus.jdt/com.redhat.quarkus.jdt.core/src/main/java/com/redhat/quarkus/jdt/core/JDTQuarkusManager.java#L698) it remove the type declared in generic
https://github.com/eclipse/eclipse.jdt.ui/blob/master/org.eclipse.jdt.core.manipulation/core%20extension/org/eclipse/jdt/internal/corext/util/JavaModelUtil.java#L507

Please create an issue about that.

angelozerr:
> Is this a problem with this PR?

That's strange it worked for me. Le me try again. Yes it's a problem with this PR. I must write test about this usecase.

angelozerr:
> Very cool PR.

Thanks!

> When I comment out/remove a ConfigProperty and save, I get duplicate completion items. If I do this multiple times, I get even more duplicate completion items. Is this a problem with this PR?

Good catch! It should work better now

PR #90, opened 2019-09-16
Remove existing Quarkus properties in completion items

### LINKED ISSUE #47: Don't provide existing Quarkus properties in completion
If application.properties contains some Quarkus properties: like:

```
quarkus.datasource.url=jdbc:postgresql:quarkus_test
quarkus.datasource.driver=org.postgresql.Driver

```

Completion should not provide thoses properties.

xorye:
Fixes #47 

To keep track of existing key names, I am using a set of strings.

[This check](https://github.com/xorye/quarkus-ls/blob/0692081d98ca1a68c01f8380c21db496d032f36c/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCompletions.java#L166-L167) is to include the current property key for completion if the current node's name is equal to the current property key. This allows you to get the completion item for the current property when the cursor is at the end like so:
Completion with cursor after key

### DIFF
--- a/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCompletions.java
+++ b/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCompletions.java
@@ -12,6 +12,7 @@ package com.redhat.quarkus.services;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
@@ -155,8 +156,19 @@ class QuarkusCompletions {
 			profile = key.getProfile();
 		}
 
+		Set<String> existingProperties = getExistingProperties(model);
+
 		// Completion on Quarkus properties
 		for (ExtendedConfigDescriptionBuildItem property : projectInfo.getProperties()) {
+			String propertyName = property.getPropertyName();
+
+			if (existingProperties.contains(propertyName) &&
+				node.getNodeType() == NodeType.PROPERTY_KEY &&
+				!((PropertyKey) node).getPropertyName().equals(propertyName)) {
+				// don't add completion items for properties that already exist
+				// unless current node has a key equal to current property name
+				continue;
+			}
 
 			CompletionItem item = new CompletionItem(property.getPropertyName());
 			item.setKind(CompletionItemKind.Property);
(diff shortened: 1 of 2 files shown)


quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCompletions.java
```diff
@@ -155,8 +156,19 @@ private static void collectPropertyKeySuggestions(int offset, Node node, Propert
...
 		}
 
+		Set<String> existingProperties = getExistingProperties(model);
+
 		// Completion on Quarkus properties
 		for (ExtendedConfigDescriptionBuildItem property : projectInfo.getProperties()) {
+			String propertyName = property.getPropertyName();
+
+			if (existingProperties.contains(propertyName) &&
+				node.getNodeType() == NodeType.PROPERTY_KEY &&
```
angelozerr: Is `if (existingProperties.contains(propertyName))` is not enough? Why writing this extra condition?

quarkus.ls/com.redhat.quarkus.ls/src/test/java/com/redhat/quarkus/services/ApplicationPropertiesCompletionTest.java
```diff
@@ -140,4 +140,18 @@ public void completionAfterProfil() throws BadLocationException {
 		testCompletionFor(value, true, c("quarkus.http.cors", "%dev.quarkus.http.cors=${1|false,true|}", r(0, 0, 5)));
 	}
 
+	@Test
+	public void noCompletionForExistingProperties() throws BadLocationException {
+		String value = "quarkus.http.cors=false\r\n" + //
+				"|";
+		testCompletionFor(value, false);
```
angelozerr: This test doesn't check the completion for existing completion. I suggest you that you use 

```
QuarkusAsset#testCompletionFor(String value, boolean snippetSupport, String fileURI, Integer expectedCount,
			QuarkusProjectInfo projectInfo, CompletionItem... expectedItems)
```

you create your own QuarkusProjectInfo  with 2 properties (ex: quarkus.application.name and quarkus.application.version)
```

In the same test you test 

 * with value="|" you should have 2 properties (by setting expectedCount=2)
 * with value="quarkus.application.name=\n|" you should have 2 properties (by setting expectedCount=1)

File: quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCompletions.java (hunk 1 above)
Comment (xorye) [reply quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCompletions.java]: It is there to allow the completion shown in the gif above. When the cursor is on or right after `quarkus.http.port`, you can see that the completion options include `quarkus.http.port`. This functionality was in this test case: https://github.com/redhat-developer/quarkus-ls/blob/0ececca93a54aa020876c33c5e4196cbfd335248/quarkus.ls/com.redhat.quarkus.ls/src/test/java/com/redhat/quarkus/services/ApplicationPropertiesCompletionTest.java#L47-L49

Comment (angelozerr) [reply quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCompletions.java]: Thanks for your explanation, forget my comment

File: quarkus.ls/com.redhat.quarkus.ls/src/test/java/com/redhat/quarkus/services/ApplicationPropertiesCompletionTest.java (hunk 2 above)
Comment (xorye) [reply quarkus.ls/com.redhat.quarkus.ls/src/test/java/com/redhat/quarkus/services/ApplicationPropertiesCompletionTest.java]: Oops, I should have been more attentive. Thanks for the correction.

File: quarkus.ls/com.redhat.quarkus.ls/src/test/java/com/redhat/quarkus/services/ApplicationPropertiesCompletionTest.java [hunk 3]
```diff
@@ -140,4 +145,28 @@ public void completionAfterProfil() throws BadLocationException {
...
+		ExtendedConfigDescriptionBuildItem p1 = new ExtendedConfigDescriptionBuildItem();
+		p1.setPropertyName("quarkus.http.cors");
+		properties.add(p1);
+		ExtendedConfigDescriptionBuildItem p2 = new ExtendedConfigDescriptionBuildItem();
+		p2.setPropertyName("quarkus.application.name");
+		properties.add(p2);
+
+		projectInfo.setProperties(properties);
+
+		testCompletionFor(value, true, null, 2, projectInfo);
```
angelozerr: could you add please quarkus properties c("quarkus.http.cors",, ..

xorye: I've made the change

angelozerr: Test are not very good, thank's!

PR #91, opened 2019-09-16
Set required for non-optional ExtendedConfigDescriptionBuildItems

xorye:
The `required` field for `ExtendedConfigDescriptionBuildItem` doesn't seem to ever be set to `true`. This PR sets the `required` field to `true` when appropriate.

What I'm not 100 percent sure about, is if `required` should be `default == null`. Maybe we should always set `required` to `true` instead? 
 
For optional properties, the `required` field will be overwritten anyway, here: https://github.com/xorye/quarkus-ls/blob/5352fde243a0f0670c37c0d910efcd35e8c24e7b/quarkus.jdt/com.redhat.quarkus.jdt.core/src/main/java/com/redhat/quarkus/jdt/core/JDTQuarkusManager.java#L445


angelozerr:
@xorye I think required is more complex. We need to play with Quarkus properties to check the rules. 

At first we can manage ConfigProperty according those rules https://quarkus.io/guides/application-configuration-guide#injecting-configuration-value and check if ConfigRoot follows the same rule.

After that we need to describe the behaviour of required in Quarkus LS. 

For instance with 

```java
@ConfigProperty(name = "greeting.message") 
String message;
```

`greeting.message` is required. It means that `greeting.message` must be declared in application.properties. We could have a validation for that and code action to insert required properties. But my fear is that you can set `greeting.message` with path environment. So in this case we should add an excluded settings (like we have for unknown properties)

The question is if we have application.properties like this:

`greeting.message=`

is it valid or not (we should play with Quarkus application to see of we will have `No config value of type [class java.lang.String] exists for: greeting.message.` exception in this case.

In other words, we must:

 * play with ConfigProperty to see if `greeting.message=` is valid. If it is not valid, we must add an error in this case and add an error in application.properties (first line and first character) which doesn't define `greeting.message`.
 * see if the Configproperty rules can be applied to ConfigRoot properties (properties which starts with `quarkus.`)

xorye:
I've looked at it in some detail, here's what I found out:

If we have this config property:
```
@ConfigProperty(name = "greeting.message")
String message;
````
and if we have `greeting.message=` in application.properties, we get an error if we try to access the endpoint:

image

For properties that start with a `quarkus.`, nothing seems to happen if you omit them. 
I've looked at `quarkus.log.syslog.use-counting-framing`, which expects a boolean: https://github.com/quarkusio/quarkus/blob/51e30bd01b288d558774c7aaaf92c8f439ab60d2/core/runtime/src/main/java/io/quarkus/runtime/logging/SyslogConfig.java#L63

**Comment by xorye** [comment]
I've made some updates to this PR in hopes of fixing #21

This PR assumes that @ConfigProperty and @ConfigItem both follow this rule for "required": a property is required if it does not have a default value, and its value type is not `java.util.Optional<java-object-here>`.

Here's the progress I've made so far:
First, I created a new validation type because I wasn't sure if it was a good idea to reuse value:
https://github.com/redhat-developer/quarkus-ls/blob/722053e160cf1f61058207233d3fc31c7ebd6209/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/ValidationType.java#L18-L21

Because @ConfigProperty and @ConfigItem follow the same rule, upon opening a newly created Quarkus project, there are errors because there are some required @ConfigItems that are not defined. 

image

If the property is missing, I wasn't sure where to put the error range, so I put it at the beginning of the document.

If I open quarkus quick starts app, I also get an error for required @ConfigProperty, greeting.mssage.


The code is not good right now, and there are no tests. I was just wondering if I could get feedback.

File: quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java [hunk 1]
```diff
@@ -242,6 +243,32 @@ private void addDiagnosticsForDuplicates() {
...
 	}
 
+	private void addDiagnosticsForMissingRequired(PropertiesModel document) {
+		for (ExtendedConfigDescriptionBuildItem property : projectInfo.getProperties()) {
+
+			String propertyName = property.getPropertyName();
+
+			DiagnosticSeverity severity = validationSettings.getDuplicate().getDiagnosticSeverity(propertyName);
+
+			if (property.getPropertyName().equals("greeting.name")) {
```
angelozerr: remove this code

angelozerr:
@xorye the main idea is to open a n application.properties from a Quarkus project sample from https://github.com/quarkusio/quarkus-quickstarts and have none errors.

It seems that you have a lot of errors with quarkus properties (from ConfigRoot) like `quarkus.application.name` which I think it's wrong. 

I suppose it's a bad idea to manage required with the same rule for ConfigProperty and ConfigRoot. I suggest you to speak about that in Zulip, but according your sample I think we should apply required rules only for ConfigProperty.

quarkus.ls/com.redhat.quarkus.ls/src/test/java/com/redhat/quarkus/settings/SettingsTest.java
```diff
@@ -75,6 +75,8 @@ public void initializationOptionsSettings() {
 		assertEquals("error", settings.getValidation().getUnknown().getSeverity());
 		assertEquals("error", settings.getValidation().getSyntax().getSeverity());
 		assertEquals("warning", settings.getValidation().getDuplicate().getSeverity());
+		assertEquals("error", settings.getValidation().getValue().getSeverity());
+		// assertEquals("error", settings.getValidation().getRequired().getSeverity());
```
angelozerr: why it is commented?

xorye:
Since this PR changed `JDTQuarkusManager.java` so that the `required` field is set here:https://github.com/redhat-developer/quarkus-ls/blob/1f951af62105d05072ed75b3983bd1a8a7a85039/quarkus.jdt/com.redhat.quarkus.jdt.core/src/main/java/com/redhat/quarkus/jdt/core/JDTQuarkusManager.java#L709

it would be great if all-quarkus-properties.json can be updated in the future. This PR is fine without that update, but I have to do an extra check here:https://github.com/redhat-developer/quarkus-ls/blob/1f951af62105d05072ed75b3983bd1a8a7a85039/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java#L254 to make sure `property.isRequired()` is not null.

angelozerr:
> o make sure property.isRequired() is not null.

what about changing Boolean to boolean in the property ExtendedConfigDescriptionBuildItem?

xorye:
> what about changing Boolean to boolean in the property ExtendedConfigDescriptionBuildItem?

This works. I've made a new commit for this change.

Changing Boolean to boolean caused me to change the `hashCode()` and `equals()` methods.  If it looks good, I can squash my commits so that this PR is ahead from master by 1 commit instead of 2.

quarkus.jdt/com.redhat.quarkus.jdt.core/src/main/java/com/redhat/quarkus/commons/ExtendedConfigDescriptionBuildItem.java
```diff
@@ -159,7 +159,7 @@ public boolean isBooleanType() {
 		"java.lang.Boolean".equals(getType()) ||
 		"java.util.Optional<java.lang.Boolean>".equals(getType());
 	}
-
+
```
angelozerr: remove this extra space

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
```diff
@@ -242,6 +244,29 @@ private void addDiagnosticsForDuplicates() {
...
+			DiagnosticSeverity severity = validationSettings.getRequired().getDiagnosticSeverity(propertyName);
+
+			if (severity != null && property.isRequired()) {
+				if (!existingProperties.containsKey(propertyName)) {
+					addDiagnostic("Missing required property '" + propertyName + "'", document, severity,
+					ValidationType.duplicate.name());
+
+				} else {
+					List<Property> propertyList = existingProperties.get(propertyName);
+					if (propertyList.size() == 1 && propertyList.get(0).getValue() == null) {
```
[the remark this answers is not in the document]
angelozerr: Please add tests with those 2 usecases.

xorye:
@angelozerr the two cases you mentioned are now taken care of:

image

image

(I've set the required severity to error only for the screenshots above. The required severity is none in this PR.)

I've also added a new test class [ApplicationPropertiesRequiredDiagnosticsTest.java
](https://github.com/redhat-developer/quarkus-ls/blob/47ae91b2f2ee244ff28a79748da6a804756be0b2/quarkus.ls/com.redhat.quarkus.ls/src/test/java/com/redhat/quarkus/services/ApplicationPropertiesRequiredDiagnosticsTest.java) for the purpose of testing the required diagnostics.

I added this new class because I wanted to create my own `QuarkusProjectInfo` with my own  `ExtendedConfigDescriptionBuildItem`s to use with multiple tests.

angelozerr:
It works like a charm. Well done @xorye !

PR #94, opened 2019-09-18
Validate application.properties value type

### LINKED ISSUE #33: Validate application.properties: type value
According the type value (ex : int, boolean, other?), validation should be done:

Ex :

 * with quarkus.datasource.enable-metrics which is a boolean

`quarkus.datasource.enable-metrics = XXXX
`

XXXX should be marked as error

 * with `quarkus.datasource.max-size` which is an int

`quarkus.datasource.max-size = XXXX
`

XXXX should be marked as error

xorye:
Fixes #33

This PR provides type validation for enums integer, boolean and float.

image

image

image

@angelozerr , does the ExtendedConfigDescriptionBuildItem.java file from quarkus.ls and quarkus.jdt.core have to be the same? 

### DIFF
--- a/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/commons/ExtendedConfigDescriptionBuildItem.java
+++ b/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/commons/ExtendedConfigDescriptionBuildItem.java
@@ -155,7 +155,42 @@ public class ExtendedConfigDescriptionBuildItem {
 	}
 
 	public boolean isBooleanType() {
-		return "boolean".equals(getType()) || "java.lang.Boolean".equals(getType());
+		return "boolean".equals(getType()) ||
+		"java.lang.Boolean".equals(getType()) ||
+		"java.util.Optional<java.lang.Boolean>".equals(getType());
+	}
+	
+	public boolean isIntegerType() {
+		return "int".equals(getType()) ||
+		"java.lang.Integer".equals(getType()) ||
+		"java.util.OptionalInt".equals(getType()) ||
+		"java.util.Optional<java.lang.Integer>".equals(getType());
+	}
+
+	public boolean isFloatType() {
+		return "float".equals(getType()) ||
+		"java.lang.Float".equals(getType()) ||
+		"java.util.Optional<java.lang.Float>".equals(getType());
+	}
+
+	public boolean isLongType() {
+		return "long".equals(getType()) ||
+		"java.lang.Long".equals(getType()) ||
+		"java.util.OptionalLong".equals(getType()) ||
+		"java.util.Optional<java.lang.Long>".equals(getType());
+	}
+
+	public boolean isDoubleType() {
+		return "double".equals(getType()) ||
+		"java.lang.Double".equals(getType()) ||
+		"java.util.OptionalDouble".equals(getType()) ||
+		"java.util.Optional<java.lang.Double>".equals(getType());
+	}
+
+	public boolean isShortType() {
+		return "short".equals(getType()) ||
+		"java.lang.Short".equals(getType()) ||
+		"java.util.Optional<java.lang.Short>".equals(getType());
 	}
 
 	@Override
--- a/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
+++ b/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
@@ -14,11 +14,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import org.eclipse.lsp4j.Diagnostic;
-import org.eclipse.lsp4j.DiagnosticSeverity;
-import org.eclipse.lsp4j.Range;
-import org.eclipse.lsp4j.jsonrpc.CancelChecker;
-
+import com.redhat.quarkus.commons.EnumItem;
 import com.redhat.quarkus.commons.ExtendedConfigDescriptionBuildItem;
 import com.redhat.quarkus.commons.QuarkusProjectInfo;
 import com.redhat.quarkus.model.Node;
@@ -28,6 +24,11 @@ import com.redhat.quarkus.model.Property;
 import com.redhat.quarkus.settings.QuarkusValidationSettings;
 import com.redhat.quarkus.utils.PositionUtils;
 import com.redhat.quarkus.utils.QuarkusPropertiesUtils;
+
+import org.eclipse.lsp4j.Diagnostic;
+import org.eclipse.lsp4j.DiagnosticSeverity;
+import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4j.jsonrpc.CancelChecker;
 
 /**
  * Quarkus validator to validate properties declared in application.properties.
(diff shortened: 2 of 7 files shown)


angelozerr:
> @angelozerr , does the ExtendedConfigDescriptionBuildItem.java file from quarkus.ls and quarkus.jdt.core have to be the same?

Yes please. The idea (in the future) is to create a common project with thoses classes which will be shared between JDT LS extension and Quarkus LS project and other Java LSP client like IJ could use it.

The PR is very promising and I'm sure @jeffmaury will like it.

But it will require some changes (I will review it in detail). But at first your PR should support the @fbricon  comment https://github.com/redhat-developer/quarkus-ls/issues/33#issuecomment-528374597

In other words:

```
quarkus.http.cors=true <-- OK
quarkus.http.cors=blablabla <-- Error
quarkus.http.cors=${blablabla} <-- OK
```

xorye:
> quarkus.http.cors=${blablabla} <-- OK

Ah yes, I forgot to write the test case for this case. I will write one now.

quarkus.ls/com.redhat.quarkus.ls/src/test/java/com/redhat/quarkus/services/ApplicationPropertiesDiagnosticsTest.java
```diff
@@ -32,17 +31,17 @@
 	@Test
 	public void validateUnknownProperties() throws BadLocationException {
 		String value = "# quarkus.datasource.url=jdbc:postgresql:quarkus_test\n" + //
-				"quarkus.datasource.driver=org.postgresql.Driver\n" + //
-				"quarkus.datasource.username=quarkus_test\n" + //
-				"quarkus.datasource.password=quarkus_test\n" + //
+				"quarkus.datasource.driver=\"org.postgresql.Driver\"\n" + //
```
xorye: My mistake, I must remove the quotations for these test cases.

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
```diff
@@ -79,7 +79,7 @@ private void validateProperty(Property property) {
 			if (metadata == null) {
 				// Validate Unknown property
 				validateUnknownProperty(propertyName, property);
-			} else {
+			} else if (property.getValue() != null) {
```
angelozerr: please remove this if condition and do the test inside validatePropertyValue which must manage validation type (as your PR) and required validation

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
```diff
@@ -130,13 +130,61 @@ private void validatePropertyValue(String propertyName, ExtendedConfigDescriptio
 			return;
 		}
 
-		// TODO validate boolean, int, enums types
-		String type = metadata.getType();
+		String value = property.getValue().getText().trim();
```
angelozerr: Replace with 

```java
String value = property.getPropertyValue();
if (value == null || value.isEmpty()) {
	// TODO: validate required.
	return;
}
```

and trim (if required should be done in property.getPropertyValue())

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
```diff
@@ -130,13 +130,61 @@ private void validatePropertyValue(String propertyName, ExtendedConfigDescriptio
...
-		// TODO validate boolean, int, enums types
-		String type = metadata.getType();
+		String value = property.getValue().getText().trim();
+		if (value.length() == 0) {
+			return;
+		}
+		
+		String errorMessage = null;
+		if (isInvalidEnumValue(metadata, value)) {
+			errorMessage = "Invalid enum value: '" + property.getValue().getText().trim() + "' is invalid for type " + metadata.getType();
```
angelozerr: replace with 

`errorMessage = "Invalid enum value: '" + value+ "' is invalid for type " + metadata.getType();`

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
```diff
@@ -130,13 +130,61 @@ private void validatePropertyValue(String propertyName, ExtendedConfigDescriptio
...
+		} else if (isValueTypeMismatch(metadata, value)) {
+			errorMessage = "Type mismatch: " + metadata.getType() + " expected";
+		}
+
+		if (errorMessage != null) {
+			addDiagnostic(errorMessage, property.getValue(), severity, ValidationType.value.name());
+		}
+	}
+
+	private boolean isInvalidEnumValue(ExtendedConfigDescriptionBuildItem metadata, String value) {
```
angelozerr: it's an utility method, please use static

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
```diff
@@ -130,13 +130,61 @@ private void validatePropertyValue(String propertyName, ExtendedConfigDescriptio
...
+		for (String e: enums) {
+			if (e.equals(value)) {
+				return false;
+			}
 		}
+
+		return true;
+	}
+
+	private boolean isValueTypeMismatch(ExtendedConfigDescriptionBuildItem metadata, String value) {
```
angelozerr: it's an utility method, please use static

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/commons/ExtendedConfigDescriptionBuildItem.java
```diff
@@ -125,7 +125,19 @@ public boolean isAvailableAtRun() {
...
+	}
+	public boolean isIntegerType() {
+		return "int".equals(getType()) ||
+		"java.lang.Integer".equals(getType()) ||
+		"java.util.OptionalInt".equals(getType()) ||
+		"java.util.Optional<java.lang.Integer>".equals(getType());
+	}
+
+	public boolean isFloatType() {
+		return "float".equals(getType());
```
angelozerr: It should be:

```java
return "float".equals(getType()) ||
		"java.lang.Float".equals(getType()) ||
		"java.util.Optional<java.lang.Float>".equals(getType());
```

angelozerr: It misses isLongType, isDoubleType

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
```diff
@@ -130,13 +130,61 @@ private void validatePropertyValue(String propertyName, ExtendedConfigDescriptio
...
+		return !isBuildtimePlaceholder(value) &&
+		((metadata.isIntegerType() && !isIntegerString(value)) ||
+		(metadata.isFloatType() && !isFloatString(value)) ||
+		(metadata.isBooleanType() && !isBooleanString(value)));
+	}
+
+	private boolean isBooleanString(String str) {
```
angelozerr: it's an utility method, please use static

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
```diff
@@ -130,13 +130,61 @@ private void validatePropertyValue(String propertyName, ExtendedConfigDescriptio
...
+		return "true".equals(str) || "false".equals(str);
+	}
+
+	private boolean isIntegerString(String str) {
```
angelozerr: it's an utility method, please use static

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
```diff
@@ -130,13 +130,61 @@ private void validatePropertyValue(String propertyName, ExtendedConfigDescriptio
...
+		return str.matches("[+-]?\\d+");
+	}
+
+	private boolean isFloatString(String str) {
```
angelozerr: it's an utility method, please use static

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
```diff
@@ -130,13 +130,61 @@ private void validatePropertyValue(String propertyName, ExtendedConfigDescriptio
...
+		return str.matches("[+-]?(\\d+\\.)?\\d+f?");
+	}
+
+	private boolean isBuildtimePlaceholder(String str) {
```
angelozerr: it's an utility method, please use static and please add javadoc

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
```diff
@@ -130,13 +130,61 @@ private void validatePropertyValue(String propertyName, ExtendedConfigDescriptio
...
+		return str.matches("^\\$\\{[A-Za-z\\d._-]*\\}$");
```
angelozerr: I think we could avoid using regexp (which can be slow) by writting str.startsWith("${");

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
angelozerr: what about isLongString and isDoubleString?

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
```diff
@@ -130,13 +130,61 @@ private void validatePropertyValue(String propertyName, ExtendedConfigDescriptio
...
+		(metadata.isFloatType() && !isFloatString(value)) ||
+		(metadata.isBooleanType() && !isBooleanString(value)));
+	}
+
+	private boolean isBooleanString(String str) {
+		return "true".equals(str) || "false".equals(str);
+	}
+
+	private boolean isIntegerString(String str) {
+		return str.matches("[+-]?\\d+");
```
angelozerr: To improve performance it's better to precompile the regexp. See the following code:

```
	private static final Pattern INTEGER_PATTER = Pattern.compile("[+-]?\\d+");
....	
	private boolean isIntegerString(String str) {
		return INTEGER_PATTER.matcher(str).matches();
	}

```

Please do like this for other matches.

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
xorye: just to confirm, should it be public static or private static?

angelozerr: private static as it is used only in this code.  If we see that this method will be used in the future in another code, we could create an utility class with static public.

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
angelozerr: BUT I think you should manage this method with Integer.parseInt loike https://stackoverflow.com/questions/237159/whats-the-best-way-to-check-if-a-string-represents-an-integer-in-java

quarkus.ls/com.redhat.quarkus.ls/src/test/java/com/redhat/quarkus/services/ApplicationPropertiesDiagnosticsTest.java
angelozerr: yes please

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusValidator.java
angelozerr: Once https://github.com/redhat-developer/quarkus-ls/pull/100 will be accepted, remove this method and use `metadata.getEnumItem(value) != null`

xorye: I think we still need `isInvalidEnumValue(metadata, value)`, because it also checks if the value is valid enum for the metadata or not

xorye:
@angelozerr I've rebased with the branch for this PR: https://github.com/redhat-developer/quarkus-ls/pull/100 and made the changes

angelozerr:
@xorye now #100 was merged in master could you rebase your PR from master please.

PR #103, opened 2019-09-24
Duplicate completion options for ConfigProperty

### LINKED ISSUE #101: Duplicate completion options for ConfigProperty
There are duplicate completion items for ConfigProperty when changing the default value.

To reproduce, continuously change the default value for a ConfigProperty, save the file, and check the list of completions in application.properties:

Gif showing the bug

This issue is a bit difficult to reproduce, I find that sometimes I get duplicates, sometimes I don't.

I find that duplicates are more likely to happen if you do not provide quotations (") for the default value, like I am doing in the gif.

angelozerr:
Duplicate completion options for ConfigProperty

Fixes #101

### DIFF
--- a/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/ls/QuarkusProjectInfoCache.java
+++ b/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/ls/QuarkusProjectInfoCache.java
@@ -15,6 +15,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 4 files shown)


angelozerr:
@xorye I have tried to fix the problem of completion duplicate. I set the PR as darft because I need to write tests. Could you play with it and tell me if it fixes your problem. Thanks!

@fbricon I'm using CopyOnWriteArrayList https://github.com/redhat-developer/quarkus-ls/pull/103/files#diff-2538c4f40f88343b007a35d0fff3cbecR47 to avoid having ConcurrentModificationException in the case of:

 * java sources changed and update Quarkus properties list
 * validation or completion is executed while the update of Quarkus properties list.

I think CopyOnWriteArrayList is good to use it in this usecase. @fbricon what do you think about that?

PR #110, opened 2019-10-09
Support for textDocument/formatting, textDocument/rangeFormatting

xorye:
Fixes #24 

### What this PR implements
* formatting and range formatting (spacing setting provided by client)
* format assertion methods for tests
* changed PropertiesParser so that in this case:
```
abcd       efghi  123
```
The `Property`'s start and end offset covers the whole line. The previous parser was implemented such that the `Property`'s start and end offset covered only `abcd` and did not parse the rest of the line.
This fix was helpful for the last property in the gif below.
Demo for formatting

For (range) formatting, the code iterates through the `Property` nodes that are children of the `PropertyModel`. Then, the `StringBuilder` accumulates the formatted properties.

### What this PR is missing
* javadocs
* tests

I will soon make a PR for vscode-quarkus for the formatting equals sign spacing setting.

PR #113, opened 2019-10-11
Fix profile tests on Windows OS.

angelozerr:
Fix profile tests on Windows OS.


xorye:
Works on my machine. I will change https://github.com/redhat-developer/quarkus-ls/pull/110 to use  System.lineSeparator()

PR #123, opened 2019-10-18
Add method to QuarkusLanguageClientAPI is now defaulted

### LINKED ISSUE #120: Provide default method when extending QuarkusLanguageClientAPI
It would be good for implementers that when a new method is added to the `QuarkusLanguageClientAPI`, it should be provided as a default method.
Ex: an new interface `QuarkusPropertyDefinitionProvider` has been added that defines a single `getPropertyDefinition` method, it could have been defaulted to `CompletableFuture.completedFuture(null)`

jeffmaury:
Fixes #120

Signed-off-by: Jeff MAURY <jmaury@redhat.com>


angelozerr:
Thanks @jeffmaury !

PR #128, opened 2019-10-24
CodeAction to fix unknown property by proposing similar name

### LINKED ISSUE #80: CodeAction to fix unknown property by proposing similar name
Given an unknown property:

`quarkus.application.nme=`

A code action could proposed to fix it with

`Did you mean with quarkus.application.name ?` 

To do that @fbricon suggest to use 

https://commons.apache.org/proper/commons-text/apidocs/org/apache/commons/text/similarity/LevenshteinDistance.html

Here a demo:

QuarkusCodeActionForUnknownDemo

angelozerr:
CodeAction to fix unknown property by proposing similar name

Fixes #80

### DIFF
--- a/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/ls/QuarkusTextDocumentService.java
+++ b/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/ls/QuarkusTextDocumentService.java
@@ -17,18 +17,10 @@ import java.util.concurrent.CompletableFuture;
 import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 
-import com.redhat.quarkus.commons.QuarkusProjectInfoParams;
-import com.redhat.quarkus.commons.QuarkusPropertiesChangeEvent;
-import com.redhat.quarkus.ls.commons.ModelTextDocument;
-import com.redhat.quarkus.ls.commons.ModelTextDocuments;
-import com.redhat.quarkus.model.PropertiesModel;
-import com.redhat.quarkus.services.QuarkusLanguageService;
-import com.redhat.quarkus.settings.QuarkusFormattingSettings;
-import com.redhat.quarkus.settings.QuarkusSymbolSettings;
-import com.redhat.quarkus.settings.QuarkusValidationSettings;
-import com.redhat.quarkus.settings.SharedSettings;
-
 import org.eclipse.lsp4j.ClientCapabilities;
+import org.eclipse.lsp4j.CodeAction;
+import org.eclipse.lsp4j.CodeActionParams;
+import org.eclipse.lsp4j.Command;
 import org.eclipse.lsp4j.CompletionItem;
 import org.eclipse.lsp4j.CompletionList;
 import org.eclipse.lsp4j.CompletionParams;
@@ -53,6 +45,17 @@ import org.eclipse.lsp4j.TextEdit;
 import org.eclipse.lsp4j.jsonrpc.CancelChecker;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
 import org.eclipse.lsp4j.services.TextDocumentService;
+
+import com.redhat.quarkus.commons.QuarkusProjectInfoParams;
+import com.redhat.quarkus.commons.QuarkusPropertiesChangeEvent;
+import com.redhat.quarkus.ls.commons.ModelTextDocument;
+import com.redhat.quarkus.ls.commons.ModelTextDocuments;
+import com.redhat.quarkus.model.PropertiesModel;
+import com.redhat.quarkus.services.QuarkusLanguageService;
+import com.redhat.quarkus.settings.QuarkusFormattingSettings;
+import com.redhat.quarkus.settings.QuarkusSymbolSettings;
+import com.redhat.quarkus.settings.QuarkusValidationSettings;
+import com.redhat.quarkus.settings.SharedSettings;
 
 /**
  * Quarkus text document service.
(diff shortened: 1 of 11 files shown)


angelozerr:
Here a demo:

QuarkusCodeActionForUnknownDemo

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/ls/QuarkusTextDocumentService.java
```diff
@@ -218,6 +226,37 @@ public void didSave(DidSaveTextDocumentParams params) {
...
+								sharedSettings.getFormattingSettings()) //
+						.stream() //
+						.map(ca -> {
+							if (codeActionLiteralSupport) {
+								Either<Command, CodeAction> e = Either.forRight(ca);
+								return e;
+							} else {
+								List<Object> arguments = Arrays.asList(uri, document.getDocument().getVersion(),
+										ca.getEdit().getDocumentChanges().get(0).getLeft().getEdits());
+								Command command = new Command(ca.getTitle(), "_quarkus.applyCodeAction", arguments);
```
fbricon: do we really still need client specific implementations to apply code actions?

quarkus.ls/com.redhat.quarkus.ls/pom.xml
```diff
@@ -111,6 +111,11 @@
 			<artifactId>org.eclipse.lsp4j.jsonrpc</artifactId>
 			<version>${lsp4j.version}</version>
 		</dependency>
+		<dependency>
+		    <groupId>org.apache.commons</groupId>
+		    <artifactId>commons-text</artifactId>
+		    <version>1.8</version>
+		</dependency>
```
fbricon: mmm we're adding 700KB (200KB of commons-text + 500KB of commons-lang) just for that feature. Seems overweight.
We should try to use proper shading to remove unnecessary cruft.

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
```diff
@@ -0,0 +1,98 @@
...
+					codeActions.add(replaceAction);
+				}
+			}
+		} catch (BadLocationException e) {
+			LOGGER.log(Level.SEVERE, "In QuarkusCodeActions, position error", e);
+		}
+	}
+
+	private static boolean isSimilar(String left, String right) {
+		int threshold = Math.round(MAX_DISTANCE_DIFF_RATIO * left.length());
```
fbricon: does it make sense to determine the threshold as 10% of the length of a word?

Alos it should be the distance from the target value, which is `right` here.

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/ls/QuarkusTextDocumentService.java
angelozerr: It's a copy paste from LSP4XML code action. If you wish I can remove it.

quarkus.ls/com.redhat.quarkus.ls/pom.xml
angelozerr: LevenshteinDistance from commons-text has none dependencies (just implement an interface but we can skip it), so if we can copy/paste this class, what do you think?

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
angelozerr: It should be fixed.

quarkus.ls/com.redhat.quarkus.ls/pom.xml
angelozerr: LevenshteinDistance class have none dependency (just one for implementing EditDistance interface) so I think we can copy/paste this class.

angelozerr: @fbricon I do a copy/paste of LevenshteinDistance  to remove the commons-text dependency.

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/ls/QuarkusTextDocumentService.java
angelozerr: @fbricon do you want i remove this code?

xorye:
Great PR

Should we support mapped properties too?
Gif showing codeaction with mapped properties

When the key is long, it is not within the threshold, and we get no code action even though I only removed "d" from "id".

When we do get a code action for the mapped property, the code action inserts the "{*}".

angelozerr:
> Should we support mapped properties too?

Indeed I have not supported it. Let me see if it's easy to support it otherwise we could fix that in an another PR if @fbricon is agree with that

angelozerr:
@xorye we decide to not support mapped property, please review my code again (and javadoc which speak about this LIMITATION), thanks!

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
```diff
@@ -0,0 +1,115 @@
...
+	 * @param projectInfo the Quarkus properties
+	 * @param codeActions code actions list to fill.
+	 */
+	private void doCodeActionsForUnknown(Diagnostic diagnostic, PropertiesModel document,
+			QuarkusProjectInfo projectInfo, List<CodeAction> codeActions) {
+		try {
+			// Get property name by using the diagnostic range
+			PropertyKey propertyKey = (PropertyKey) document.findNodeAt(diagnostic.getRange().getStart());
+			String propertyName = propertyKey.getPropertyName();
+			// Loop for each metata property
```
xorye: "metata property" should be "metadata property"

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
```diff
@@ -0,0 +1,115 @@
...
+			for (ExtendedConfigDescriptionBuildItem metaProperty : projectInfo.getProperties()) {
+				if (QuarkusPropertiesUtils.isMappedProperty(metaProperty.getPropertyName())) {
+					// FIXME: support mapped property
```
xorye: Do you think we might support mapped property in the future?

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
angelozerr: @xorye fixed

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
angelozerr: It should be fine but today I have not found a clean solution to do that.

angelozerr:
@fbricon @xorye is there something you wish I do?

xorye:
Looks good to me

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/ls/QuarkusTextDocumentService.java
fbricon: yup

PR #140, opened 2019-11-12
Provide codelenses for endpoints of the running app

### LINKED ISSUE #114: Provide codelenses for endpoints of the running app
connect to local running application
connect to a remote server???
The idea is to provide links from REST endpoints in the code that will be opened in the browser when clicking on them.

Look at how spring tools works with spring boot applications

angelozerr:
Provide codelenses for endpoints of the running app

Fixes #114

### DIFF
--- a/quarkus.jdt/com.redhat.quarkus.jdt.core/src/main/java/com/redhat/quarkus/jdt/core/JDTQuarkusManagerForJava.java
+++ b/quarkus.jdt/com.redhat.quarkus.jdt.core/src/main/java/com/redhat/quarkus/jdt/core/JDTQuarkusManagerForJava.java
@@ -1,0 +1,387 @@
+/*******************************************************************************
+* Copyright (c) 2019 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.quarkus.jdt.core;
+
+import static com.redhat.quarkus.jdt.internal.core.utils.AnnotationUtils.getAnnotation;
+import static com.redhat.quarkus.jdt.internal.core.utils.AnnotationUtils.getAnnotationMemberValue;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.InetSocketAddress;
+import java.net.Socket;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.jdt.core.Flags;
+import org.eclipse.jdt.core.IAnnotatable;
+import org.eclipse.jdt.core.IAnnotation;
+import org.eclipse.jdt.core.IClassFile;
+import org.eclipse.jdt.core.IClasspathEntry;
+import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IJavaElement;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.IMethod;
+import org.eclipse.jdt.core.ISourceRange;
+import org.eclipse.jdt.core.ISourceReference;
+import org.eclipse.jdt.core.IType;
+import org.eclipse.jdt.core.ITypeRoot;
+import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.jdt.internal.core.JavaProject;
+import org.eclipse.lsp4j.CodeLens;
+import org.eclipse.lsp4j.Command;
+import org.eclipse.lsp4j.Range;
+
+import com.redhat.quarkus.commons.QuarkusJavaCodeLensParams;
+
+/**
+ * JDT quarkus manager for Java files.
+ * 
+ * @author Angelo ZERR
+ *
+ */
+public class JDTQuarkusManagerForJava {
+
+	private static final Logger LOGGER = Logger.getLogger(JDTQuarkusManagerForJava.class.getName());
+
+	private static final String LOCALHOST = "localhost";
+
+	private static final int PING_TIMEOUT = 2000;
+
+	private static final String JAVAX_WS_RS_PATH_ANNOTATION = "javax.ws.rs.Path";
+
+	private static final String JAVAX_WS_RS_GET_ANNOTATION = "javax.ws.rs.GET";
+
+	private static final String PATH_VALUE = "value";
+
+	private static final JDTQuarkusManagerForJava INSTANCE = new JDTQuarkusManagerForJava();
+
+	private static class JDTQuarkusProjectInfo {
+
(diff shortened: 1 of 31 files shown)


angelozerr:
This draft PR requires the draft PR https://github.com/redhat-developer/vscode-quarkus/pull/146 (from vscode-quarkus): 

image

First please don't review the code, I must clean it, but you can start play with the codelens but there are several limitation:

- [ ] codelens should be displayed according a settings. Is we should check to the vscode-java codelens enabled settings?
- [ ] click on URL codelens open a web browser with a Command, but we should check that client capability can manage this open URL 
- [ ] when Java text changed, the ICompilationUnit get from JDTUtils is not synchronized with the text editor. How to fix that? Not sure with that but Spring LS seems creating their own ICompilationUnit, I would like to avoid doing that to avoird parsing twice the Java content.
- [ ] Codelens URL is displayed even if Quarkus application is not started, How to do that?
- [ ] the base URL is hard coded, we should compute it by using port of application.properties and gives the capability to customize this base URL (in an array string)) when base URL cannot be computed (ex : base URL coming from a pom.xml)

fbricon:
Before displaying the code lens, you might want to see if the url is accessible

angelozerr:
> Before displaying the code lens, you might want to see if the url is accessible (check for connection refused)

Indeed it's a limitation that I wrote: "Codelens URL is displayed even if Quarkus application is not started, How to do that?"

fbricon:
try https://stackoverflow.com/a/3584332/753170 before sending codelenses

fbricon:
Slashes should be de-duplicated.
Screen Shot 2019-11-12 at 11 23 12 AM

Using
```java
package foo.bar;

import javax.ws.rs.GET;
import javax.ws.rs.Path;
import javax.ws.rs.Produces;
import javax.ws.rs.core.MediaType;

@Path("/salut")
public class HelloResource {
    
    @GET
    @Produces(MediaType.TEXT_PLAIN)
    @Path("/tadaa")
    //should generate http://localhost:8080/salut/tadaa instead of http://localhost:8080/salut//tadaa
    public String hello() {
        return "Salut";
    }
}
```

angelozerr:
I must write tests but you can play with PR. Please note this PR requires https://github.com/redhat-developer/vscode-quarkus/pull/146

xorye:
works quite well on my end

quarkus.jdt/com.redhat.quarkus.jdt.core/src/main/java/com/redhat/quarkus/jdt/core/JDTQuarkusManagerForJava.java
```diff
@@ -0,0 +1,311 @@
...
+ * JDT quarkus manager for Java files.
+ * 
+ * @author Angelo ZERR
+ *
+ */
+public class JDTQuarkusManagerForJava {
+
+	private static final Logger LOGGER = Logger.getLogger(JDTQuarkusManagerForJava.class.getName());
+
+	private static final String LOCAHOST = "localhost";
```
fbricon: LOCALHOST

angelozerr: @fbricon updated

fbricon:
adding static methods, they get codelenses. I don't think this is allowed, certainly not for private methods:

Screen Shot 2019-11-26 at 7 40 30 PM

fbricon:
After adding `quarkus.http.port=9090` to application.properties and reopening a JAX-RS resource, the codelenses still point toward 8080

quarkus.jdt/com.redhat.quarkus.jdt.core/src/main/java/com/redhat/quarkus/jdt/core/JDTQuarkusManagerForJava.java
```diff
@@ -0,0 +1,351 @@
...
+	public List<? extends CodeLens> codeLens(QuarkusJavaCodeLensParams params, IJDTUtils utils,
+			IProgressMonitor monitor) throws JavaModelException {
+		String uri = params.getUri();
+		ITypeRoot typeRoot = resolveTypeRoot(uri, utils, monitor);
+		if (typeRoot == null) {
+			return Collections.emptyList();
+		}
+		IJavaElement[] elements = typeRoot.getChildren();
+		Collection<CodeLens> lenses = new LinkedHashSet<>(elements.length);
+		if (params.isUrlCodeLensEnabled()) {
```
fbricon: check for server available should happen here, only once

quarkus.jdt/com.redhat.quarkus.jdt.core/src/main/java/com/redhat/quarkus/jdt/core/JDTQuarkusManagerForJava.java
```diff
@@ -0,0 +1,351 @@
...
+				return;
+			}
+			if (element.getElementType() == IJavaElement.TYPE) {
+				IType type = (IType) element;
+				// Get value of JAX-RS @Path annotation from the class
+				String pathValue = getJaxRsPathValue(type);
+				if (pathValue != null) {
+					// Class is annotated with @Path
+					// Display code lens only if local server is available.
+					if (!params.isCheckServerAvailable()
```
xorye: `params.isCheckServerAvailable()` seems to be always false because `params.setCheckServerAvailable()` is never called.  Therefore causing https://github.com/redhat-developer/vscode-quarkus/pull/146#issuecomment-558761502

angelozerr: Indeed @xorye it should be fixed now.

quarkus.jdt/com.redhat.quarkus.jdt.core/src/main/java/com/redhat/quarkus/jdt/core/JDTQuarkusManagerForJava.java
angelozerr: If I do that I check the server for any Java file. In my case I check the server only once for Java class bound with JAX-RS Path (just for the class not for the method), that's to say, the server check is done once.

fbricon: there's no need to iterate on all members if the server is not even started

angelozerr:
> After adding quarkus.http.port=9090 to application.properties and reopening a JAX-RS resource, the codelenses still point toward 8080

It should work now, but please check that URL is available because sometimes when I change the value of port, the server is not restarted and keep the old port?

quarkus.jdt/com.redhat.quarkus.jdt.core/src/main/java/com/redhat/quarkus/jdt/core/JDTQuarkusManagerForJava.java
angelozerr: > there's no need to iterate on all members if the server is not even started

I would like to avoid checking the server state of Java class in not a JAX-RS (annotated with @Path).

Given this java file:
```
@Path
public class A {

  public void f()

  public void g()
}
```

One check is done (only on A java element IType not for the methods)


Given this java file

```
class B {

}
```

None check on server state is done since B is no bound to @Path.

angelozerr:
> adding static methods, they get codelenses. I don't think this is allowed, certainly not for private methods:

You are right, after playing with JAX-RS, the rule is:

 * method must be public (static seems working)
 * and method must be annotated with @GET, @PUT, @DELETE or @POST

angelozerr:
> After adding quarkus.http.port=9090 to application.properties and reopening a JAX-RS resource, the codelenses still point toward 8080

It's an issue for Quarkus I think https://github.com/quarkusio/quarkus/issues/5864

PR #141, opened 2019-11-12
Implement code action to add unknown property to exclude array

### LINKED ISSUE #81: CodeAction to add property to ignore for unknown validation
When there are an error with unkwown property, a CodeAction could be proposed to add the property in the excluded list of the settings.

This feature requires the check of client capability to update the proper settings:

 * for vscode -> settings.json
 * for Eclispe IDE -> preferences file

etc

xorye:
Fixes #81 

I have a few comments about this PR:

To test this PR, this PR: https://github.com/redhat-developer/vscode-quarkus/pull/152 from vscode-quarkus is needed.

In this PR, the extended capability that deals with whether or not the client is able to update their settings was named [`ExtendedClientSettingsUpdateCapabilities`](https://github.com/xorye/quarkus-ls/blob/81codeaction_unknown/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/ls/client/ExtendedClientSettingsUpdateCapabilities.java).

For the client to register the capability, the client must have their initialization options like so:
```
      initializationOptions: {
        extendedClientCapabilities: {
          clientSettingsUpdate: { supported: true }
        }
      }
```
The PR from vscode-quarkus mentioned above, already does this.

The command that quarkus-ls sends to vscode-quarkus was named `quarkus/clientSettingsUpdate`, which is a command that vscode-quarkus registers. Then, quarkus-ls would pass in 2 arguments to the  `quarkus/clientSettingsUpdate` command, the first argument being `validationUnknownExcluded` and the second argument being the property name. The first argument tells vscode-quarkus which setting to update.

Maybe it's better to only have a command called something like  `quarkus/addToUnknownValidationExcluded` that only deals with adding to the unknown validation excluded array? 

Also, is it a good idea to add a method in [`CodeActionFactory`](https://github.com/xorye/quarkus-ls/blob/81codeaction_unknown/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/ls/commons/CodeActionFactory.java) that creates code actions with commands? Currently, `CodeActionFactory` only has methods for insert, remove and replace, but no command.

### DIFF
--- a/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
+++ b/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
@@ -11,6 +11,7 @@ package com.redhat.quarkus.services;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -22,11 +23,15 @@ import com.redhat.quarkus.commons.QuarkusProjectInfo;
 import com.redhat.quarkus.ls.commons.BadLocationException;
 import com.redhat.quarkus.ls.commons.CodeActionFactory;
 import com.redhat.quarkus.ls.commons.TextDocument;
+import com.redhat.quarkus.ls.commons.client.CommandKind;
+import com.redhat.quarkus.ls.commons.client.ConfigurationItemEdit;
+import com.redhat.quarkus.ls.commons.client.ConfigurationItemEditType;
 import com.redhat.quarkus.model.PropertiesModel;
 import com.redhat.quarkus.model.Property;
 import com.redhat.quarkus.model.PropertyKey;
 import com.redhat.quarkus.model.PropertyValue;
 import com.redhat.quarkus.model.values.ValuesRulesManager;
+import com.redhat.quarkus.settings.QuarkusCommandCapabilities;
 import com.redhat.quarkus.settings.QuarkusFormattingSettings;
 import com.redhat.quarkus.utils.PositionUtils;
 import com.redhat.quarkus.utils.QuarkusPropertiesUtils;
@@ -34,6 +39,8 @@ import com.redhat.quarkus.utils.StringUtils;
 
 import org.eclipse.lsp4j.CodeAction;
 import org.eclipse.lsp4j.CodeActionContext;
+import org.eclipse.lsp4j.CodeActionKind;
+import org.eclipse.lsp4j.Command;
 import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
@@ -55,15 +62,17 @@ class QuarkusCodeActions {
 	 * <code>document</code> by using the given Quarkus properties metadata
 	 * <code>projectInfo</code>.
 	 * 
-	 * @param context            the code action context
-	 * @param range              the range
-	 * @param document           the properties model.
-	 * @param projectInfo        the Quarkus properties
-	 * @param formattingSettings the formatting settings.
+	 * @param context             the code action context
+	 * @param range               the range
+	 * @param document            the properties model.
+	 * @param projectInfo         the Quarkus properties
+	 * @param formattingSettings  the formatting settings.
+	 * @param commandCapabilities the command capabilities
 	 * @return the result of the code actions.
 	 */
 	public List<CodeAction> doCodeActions(CodeActionContext context, Range range, PropertiesModel document,
-			QuarkusProjectInfo projectInfo, ValuesRulesManager valuesRulesManager, QuarkusFormattingSettings formattingSettings) {
+			QuarkusProjectInfo projectInfo, ValuesRulesManager valuesRulesManager,
+			QuarkusFormattingSettings formattingSettings, QuarkusCommandCapabilities commandCapabilities) {
 		List<CodeAction> codeActions = new ArrayList<>();
 		if (context.getDiagnostics() != null) {
 			doCodeActionForAllRequired(context.getDiagnostics(), document, formattingSettings, codeActions);
@@ -71,7 +80,7 @@ class QuarkusCodeActions {
 			for (Diagnostic diagnostic : context.getDiagnostics()) {
 				if (ValidationType.unknown.name().equals(diagnostic.getCode())) {
 					// Manage code action for unknown
-					doCodeActionsForUnknown(diagnostic, document, projectInfo, codeActions);
+					doCodeActionsForUnknown(diagnostic, document, projectInfo, commandCapabilities, codeActions);
 				} else if (ValidationType.value.name().equals(diagnostic.getCode())) {
 					doCodeActionsForUnknownEnumValue(diagnostic, document, projectInfo, valuesRulesManager, codeActions);
 				}
@@ -88,13 +97,15 @@ class QuarkusCodeActions {
 	 * LIMITATION: mapped property are not supported.
 	 * </p>
 	 * 
-	 * @param diagnostic  the diagnostic
-	 * @param document    the properties model.
-	 * @param projectInfo the Quarkus properties
-	 * @param codeActions code actions list to fill.
+	 * @param diagnostic          the diagnostic
+	 * @param document            the properties model.
+	 * @param projectInfo         the Quarkus properties
+	 * @param commandCapabilities the command capabilities
+	 * @param codeActions         code actions list to fill.
 	 */
 	private void doCodeActionsForUnknown(Diagnostic diagnostic, PropertiesModel document,
-			QuarkusProjectInfo projectInfo, List<CodeAction> codeActions) {
+			QuarkusProjectInfo projectInfo, QuarkusCommandCapabilities commandCapabilities,
+			List<CodeAction> codeActions) {
 		try {
 			// Get property name by using the diagnostic range
 			PropertyKey propertyKey = (PropertyKey) document.findNodeAt(diagnostic.getRange().getStart());
@@ -113,6 +124,10 @@ class QuarkusCodeActions {
 						codeActions.add(replaceAction);
 					}
 				}
+			}
+
+			if (commandCapabilities.isCommandSupported(CommandKind.COMMAND_CONFIGURATION_UPDATE)) {
+				doCodeActionForIgnoreUnknownValidation(propertyName, diagnostic, codeActions);
 			}
 		} catch (BadLocationException e) {
 			LOGGER.log(Level.SEVERE, "In QuarkusCodeActions, position error", e);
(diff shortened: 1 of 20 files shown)


xorye:
I've made a new commit to this PR. (I will squash all commits to one, before this PR is ready to merge)

I created a new class: [ConfigurationItemEdit](https://github.com/redhat-developer/quarkus-ls/blob/820a2d94fa81f4cf8492bae9d10c42421d3cb486/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/ConfigurationItemEdit.java) which represents a change to perform on the client config.

I currently have `ConfigurationItemEdit` extending `lsp4j.ConfigurationItem`. I'm wondering, is this correct? At the time being, I don't think `ConfigurationItemEdit` and `lsp4j.ConfigurationItem` have an "is a" relationship.

Also, currently I am not setting the `lsp4j.ConfigurationItem.scopeUri` class variable.

For the code action's command, I am sending the `ConfigurationItemEdit` instance to the client, with the expectation that the client would read it and perform necessary changes to its config. I have updated the https://github.com/redhat-developer/vscode-quarkus/pull/152 PR to reflect that behaviour.

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/ls/client/ExtendedClientCapabilities.java
```diff
@@ -0,0 +1,28 @@
...
+ * Extended client capabilities not defined by the LSP.
+ * 
+ * @author Angelo ZERR
+ */
+public class ExtendedClientCapabilities {
+
+	private ExtendedClientSettingsUpdateCapabilities clientSettingsUpdate;
+
+	public ExtendedClientSettingsUpdateCapabilities getClientSettingsUpdate() {
+		return clientSettingsUpdate;
```
angelozerr: I think `configuration `should be used instead of `settings` (ex: ConfigurationItemEdit). A suggestion is to rename `clientSettingsUpdate` to `configurationUpdate`  or `configurationEdit`

@fbricon what do you think ?

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/ConfigurationItemEdit.java
```diff
@@ -0,0 +1,71 @@
...
+	private Object value;
+
+	/**
+	 * 
+	 * @param section   config section to change
+	 * @param operation type of change
+	 * @param value     the value for the change
+	 */
+	public ConfigurationItemEdit(String section, ConfigurationItemEditType editType, Object value) {
+		this.setSection(section);
```
angelozerr: super.setSection(section)

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/ConfigurationItemEdit.java
```diff
@@ -0,0 +1,71 @@
...
+		this.editType = editType;
+		this.value = value;
+	}
+
+	@Override
+	public boolean equals(final Object obj) {
```
angelozerr: Have you generated this equals and hashCode methods? It's better to use equals and hashCode of the extended ConfigurationItem class.

Here the code generated by Eclipse :

```java
@Override
public boolean equals(Object obj) {
	if (this == obj)
		return true;
	if (!super.equals(obj))
		return false;
	if (getClass() != obj.getClass())
		return false;
	ConfigurationItemEdit other = (ConfigurationItemEdit) obj;
	if (editType == null) {
		if (other.editType != null)
			return false;
	} else if (!editType.equals(other.editType))
		return false;
	if (value == null) {
		if (other.value != null)
			return false;
	} else if (!value.equals(other.value))
		return false;
	return true;
}

@Override
public int hashCode() {
	final int prime = 31;
	int result = super.hashCode();
	result = prime * result + ((editType == null) ? 0 : editType.hashCode());
	result = prime * result + ((value == null) ? 0 : value.hashCode());
	return result;
}
```

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/ConfigurationItemEditType.java
```diff
@@ -0,0 +1,16 @@
...
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.quarkus.services;
+
+public enum ConfigurationItemEditType {
```
angelozerr: Add javadoc

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/ls/client/ExtendedClientCapabilities.java
fbricon: +1 to use configuration

fbricon:
Given that we still don't have kubernetes properties support, the following would be annoying to exclude one by one:
```
kubernetes.group=myUser
kubernetes.registry=http://my.docker-registry.net
kubernetes.labels[0].key=foo
kubernetes.labels[0].value=bar
kubernetes.readiness-probe.initial-delay-seconds=20
kubernetes.readiness-probe.period-seconds=45
```

I'd say that if there are multiple unknown properties, we should be able to find their closest common parent key and then offer to exclude it globally. In this case we'd offer to also exclude kubernetes.*, in addition to each individual property

angelozerr:
@fbricon I had this idea but if you have a quarkus propertie we will ignore all quarkus properties with `quarkus.*` It's a little annoying, no? Perhaps we should do an exception for quarkus properties?

angelozerr:
@xorye after discussion with @fbricon we think it should be better to have a command capabilities instead of having clientSettingsUpdate:

```
"extendedClientCapabilities": {
    "command": {
        "commandKind": {
            "valueSet": [
                "quarkus.command.configuration.update"
            ]
        }
    }
} 
```

I start to create this extended client capabilities that you can copy/paste at 
https://github.com/redhat-developer/quarkus-ls/compare/master...angelozerr:client-capabilities

Using command is more generic because I could use this capability for codelens (click on CodeLens URL work only if client has the capability to manage the open URI command). 

In your case you are using configuration command update with CodeAction, but we could consume it too with CodeLens.

If you have any question don't hesitate to ping me.

Many thanks for the change!

xorye:
@angelozerr for `QuarkusCodeLens.java` and `QuarkusCodeActions.java` to access `CommandCapabilities`, do we need to create a `QuarkusCommandCapabilitiesSettings.java` in `com.redhat.quarkus/settings`?

angelozerr:
I think you ask me that because in lsp4xml we have that with codelens capabilties and codelens settings. I think its a bad idea to mix capabilty which is static with settings configuration which is dynamic. Try to avoid creating those new settings we will see if it is better

xorye:
@angelozerr I have updated his PR so that it uses CommandCapability.

I've added `QuarkusCommandCabilities.java` with a `public boolean isCommandSupported(String command)` method:
https://github.com/redhat-developer/quarkus-ls/blob/2429eea2b0c70ffad46589557754afdb4894ae1d/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/settings/QuarkusCommandCapabilities.java#L25-L34

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/settings/QuarkusCommandCapabilities.java
```diff
@@ -0,0 +1,39 @@
...
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.quarkus.settings;
+
+import com.redhat.quarkus.ls.client.CommandCapabilities;
+import com.redhat.quarkus.ls.client.CommandKind;
+
```
angelozerr: Add javadoc 

A wrapper around LSP {@link CommandCapabilities}.

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
```diff
@@ -184,37 +197,27 @@ private void doCodeActionsForUnknownEnumValue(Diagnostic diagnostic, PropertiesM
...
-	 * Returns the missing required property name from <code>diagnosticMessage</code>
-	 * @param diagnosticMessage the diagnostic message containing the property name in single quotes
-	 * @return the missing required property name from <code>diagnosticMessage</code>
-	 */
-	private String getPropertyNameFromRequiredMessage(String diagnosticMessage) {
-		int start = diagnosticMessage.indexOf('\'') + 1;
-		int end = diagnosticMessage.indexOf('\'', start);
-		return diagnosticMessage.substring(start, end);
+		Command command = new Command("Add " + propertyName + " to unknown excluded array",
+				"quarkus/configurationUpdate", Arrays.asList(configItemEdit));
```
angelozerr: "quarkus/configurationUpdate" should be replaced with CommandKind.COMMAND_CONFIGURATION_UPDATE, no?

angelozerr: replace `Arrays.asList(configItemEdit)` with `Collections.singletonList(configItemEdit)`

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/ls/client/CommandCapabilities.java
```diff
@@ -0,0 +1,101 @@
+/*******************************************************************************
+* Copyright (c) 2019 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.quarkus.ls.client;
```
angelozerr: Sorry it was my fault, but please move those classes in the `com.redhat.quarkus.ls.commons.client` package. It should be a commons LSP feature.

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/ls/client/ExtendedClientCapabilities.java
```diff
@@ -0,0 +1,31 @@
+/*******************************************************************************
+* Copyright (c) 2019 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.quarkus.ls.client;
```
angelozerr: please move to com.redhat.quarkus.ls.commons.client package.

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/ConfigurationItemEdit.java
```diff
@@ -0,0 +1,64 @@
+/*******************************************************************************
+* Copyright (c) 2019 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.quarkus.services;
```
angelozerr: please move to com.redhat.quarkus.ls.commons.client package

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/ConfigurationItemEditType.java
```diff
@@ -0,0 +1,20 @@
+/*******************************************************************************
+* Copyright (c) 2019 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.quarkus.services;
```
angelozerr: please move to com.redhat.quarkus.ls.commons.client

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
```diff
@@ -184,37 +200,27 @@ private void doCodeActionsForUnknownEnumValue(Diagnostic diagnostic, PropertiesM
...
-	private String getPropertyNameFromRequiredMessage(String diagnosticMessage) {
-		int start = diagnosticMessage.indexOf('\'') + 1;
-		int end = diagnosticMessage.indexOf('\'', start);
-		return diagnosticMessage.substring(start, end);
+		Command command = new Command("Add " + propertyName + " to unknown excluded array",
+				CommandKind.COMMAND_CONFIGURATION_UPDATE, Collections.singletonList(configItemEdit));
+		insertContentAction.setCommand(command);
+		insertContentAction.setKind(CodeActionKind.QuickFix);
+		insertContentAction.setDiagnostics(Arrays.asList(diagnostic));
+		codeActions.add(insertContentAction);
```
angelozerr: I think you should add a new CodeAction (only if propertyName doesn't starts with` quarkus.` or `%profile.quarkus.` ) to propose to exclude the property with pattern. See https://github.com/redhat-developer/quarkus-ls/pull/141#issuecomment-554935317 @fbricon comments.

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
```diff
@@ -184,37 +200,27 @@ private void doCodeActionsForUnknownEnumValue(Diagnostic diagnostic, PropertiesM
...
-	 */
-	private String getPropertyNameFromRequiredMessage(String diagnosticMessage) {
-		int start = diagnosticMessage.indexOf('\'') + 1;
-		int end = diagnosticMessage.indexOf('\'', start);
-		return diagnosticMessage.substring(start, end);
+		Command command = new Command("Add " + propertyName + " to unknown excluded array",
+				CommandKind.COMMAND_CONFIGURATION_UPDATE, Collections.singletonList(configItemEdit));
+		insertContentAction.setCommand(command);
+		insertContentAction.setKind(CodeActionKind.QuickFix);
+		insertContentAction.setDiagnostics(Arrays.asList(diagnostic));
```
angelozerr: replace with Collections.singletonList(diagnostic))

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
xorye: I will make a separate issue for this

angelozerr:
Great job @xorye !

xorye:
@angelozerr Thank you for your review & patience!

PR #145, opened 2019-11-14
Create code action for propose similar enum value

### LINKED ISSUE #130: CodeAction to fix value property by proposing similar value
Like https://github.com/redhat-developer/quarkus-ls/issues/80 for property name, it should be good to have the same feature for value when value is an enumeration.

xorye:
Fixes #130

Gif showing codeaction

### DIFF
--- a/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
+++ b/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
@@ -10,22 +10,28 @@
 package com.redhat.quarkus.services;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
+import com.redhat.quarkus.commons.EnumItem;
 import com.redhat.quarkus.commons.ExtendedConfigDescriptionBuildItem;
 import com.redhat.quarkus.commons.QuarkusProjectInfo;
 import com.redhat.quarkus.ls.commons.BadLocationException;
 import com.redhat.quarkus.ls.commons.CodeActionFactory;
 import com.redhat.quarkus.ls.commons.TextDocument;
 import com.redhat.quarkus.model.PropertiesModel;
+import com.redhat.quarkus.model.Property;
 import com.redhat.quarkus.model.PropertyKey;
+import com.redhat.quarkus.model.PropertyValue;
+import com.redhat.quarkus.model.values.ValuesRulesManager;
 import com.redhat.quarkus.settings.QuarkusFormattingSettings;
 import com.redhat.quarkus.utils.PositionUtils;
 import com.redhat.quarkus.utils.QuarkusPropertiesUtils;
 import com.redhat.quarkus.utils.StringUtils;
+
 import org.eclipse.lsp4j.CodeAction;
 import org.eclipse.lsp4j.CodeActionContext;
 import org.eclipse.lsp4j.Diagnostic;
@@ -57,7 +63,7 @@ class QuarkusCodeActions {
 	 * @return the result of the code actions.
 	 */
 	public List<CodeAction> doCodeActions(CodeActionContext context, Range range, PropertiesModel document,
-			QuarkusProjectInfo projectInfo, QuarkusFormattingSettings formattingSettings) {
+			QuarkusProjectInfo projectInfo, ValuesRulesManager valuesRulesManager, QuarkusFormattingSettings formattingSettings) {
 		List<CodeAction> codeActions = new ArrayList<>();
 		if (context.getDiagnostics() != null) {
 			doCodeActionForAllRequired(context.getDiagnostics(), document, formattingSettings, codeActions);
@@ -66,6 +72,8 @@ class QuarkusCodeActions {
 				if (ValidationType.unknown.name().equals(diagnostic.getCode())) {
 					// Manage code action for unknown
 					doCodeActionsForUnknown(diagnostic, document, projectInfo, codeActions);
+				} else if (ValidationType.value.name().equals(diagnostic.getCode())) {
+					doCodeActionsForUnknownEnumValue(diagnostic, document, projectInfo, valuesRulesManager, codeActions);
 				}
 			}
 		}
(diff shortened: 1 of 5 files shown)


quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/commons/QuarkusProjectInfo.java
```diff
@@ -64,6 +64,22 @@ public void setClasspathKind(ClasspathKind classpathKind) {
...
 	}
 
+	/**
+	 * Returns a <code>ExtendedConfigDescriptionBuildItem</code> representing the
+	 * <code>propertyName</code> property
+	 * @param propertyName the property name of interest
+	 * @return <code>ExtendedConfigDescriptionBuildItem</code> representing the
+	 * <code>propertyName</code> property
+	 */
+	public ExtendedConfigDescriptionBuildItem getPropertyFromName(String propertyName) {
```
angelozerr: Remove this method to use `QuarkusPropertiesUtils#getProperty(String propertyName, QuarkusProjectInfo info)` which manages search of mapped property.

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
```diff
@@ -111,6 +118,50 @@ private void doCodeActionsForUnknown(Diagnostic diagnostic, PropertiesModel docu
...
+	 */
+	private void doCodeActionsForUnknownEnumValue(Diagnostic diagnostic, PropertiesModel document,
+			QuarkusProjectInfo projectInfo, ValuesRulesManager valuesRulesManager, List<CodeAction> codeActions) {
+		try {
+			PropertyValue propertyValue = (PropertyValue) document.findNodeAt(diagnostic.getRange().getStart());
+			PropertyKey propertyKey = ((Property) propertyValue.getParent()).getKey();
+			String value = propertyValue.getValue();
+			String propertyName = propertyKey.getPropertyName();
+
+			ExtendedConfigDescriptionBuildItem metaProperty = projectInfo.getPropertyFromName(propertyName);
```
angelozerr: Replace with:

```java
ExtendedConfigDescriptionBuildItem metaProperty = QuarkusPropertiesUtils.getProperty(propertyName, projectInfo)
```

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
```diff
@@ -111,6 +118,50 @@ private void doCodeActionsForUnknown(Diagnostic diagnostic, PropertiesModel docu
...
+			if (metaProperty == null) {
+				return;
+			}
+
+			List<EnumItem> enums = metaProperty.getEnums();
```
angelozerr: In QuarkusCompletions, you have a private static method

```java
 Collection<EnumItem> getEnums(ExtendedConfigDescriptionBuildItem property, PropertiesModel model,
			ValuesRulesManager valuesRulesManager) 
```
which does the same thing (nut manages boolean type too).

I suggest you to move this private static method from QuarkusCompletion in the QuarkusPropertiesUtils and you consume it here. 
Replace this enums collection by creating a new method utility `Collection<EnumItem> getEnums(ExtendedConfigDescriptionBuildItem property, PropertiesModel model,	ValuesRulesManager valuesRulesManager)` You can find this method in

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
```diff
@@ -111,6 +118,50 @@ private void doCodeActionsForUnknown(Diagnostic diagnostic, PropertiesModel docu
...
+			if (enums == null || enums.isEmpty()) {
+				enums = valuesRulesManager.getValues(metaProperty, document);
+				if (enums == null || enums.isEmpty()) {
+					return;
+				}
+			}
+
+			for (EnumItem e : enums) {
+				if (isSimilar(e.getName(), value)) {
```
angelozerr: In LSP4XML we have a current PR which does the same thing (but with XML Schema). The idea is to provide a code actions for similar names (like you have done). And if you have none similar name, you provide a code action per enum item by saying 'Replace with '

See https://github.com/angelozerr/lsp4xml/blob/655e17c197011cf41fe1880c54a861f2c1ef00c2/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/participants/codeactions/cvc_complex_type_2_4_aCodeAction.java#L76

quarkus.ls/com.redhat.quarkus.ls/src/test/java/com/redhat/quarkus/services/ApplicationPropertiesCodeActionsTest.java
```diff
@@ -42,4 +42,29 @@ public void codeActionsForUnknownProperties() throws BadLocationException {
...
+		String value = "quarkus.log.syslog.async.overflow=BLACK";
+
+		Diagnostic d = d(0, 34, 39, "Invalid enum value: 'BLACK' is invalid for type org.jboss.logmanager.handlers.AsyncHandler$OverflowAction", DiagnosticSeverity.Error,
+				ValidationType.value);
+		testDiagnosticsFor(value, d);
+
+		testCodeActionsFor(value, d,
+				ca("Did you mean 'BLOCK'?", te(0, 34, 0, 39, "BLOCK"), d));
+	};
+
```
angelozerr: Add test with `Replace with` code action.

xorye:
Thank you for the great feedback, I've made the changes

quarkus.ls/com.redhat.quarkus.ls/src/test/java/com/redhat/quarkus/services/ApplicationPropertiesCodeActionsTest.java
```diff
@@ -37,9 +37,56 @@ public void codeActionsForUnknownProperties() throws BadLocationException {
...
+		Diagnostic d = d(0, 34, 47,
+				"Invalid enum value: 'unknown-value' is invalid for type org.jboss.logmanager.handlers.AsyncHandler$OverflowAction",
+				DiagnosticSeverity.Error,
+				ValidationType.value);
+
+		testDiagnosticsFor(value, d);
+		testCodeActionsFor(value, d,
+				ca("Replace with 'BLOCK'?", te(0, 34, 0, 47, "BLOCK"), d),
+				ca("Replace with 'DISCARD'?", te(0, 34, 0, 47, "DISCARD"), d));
+	};
```
angelozerr: Add a test with the mapped property `quarkus.log.category."org.acme".level=WARNIN`

quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java
```diff
@@ -111,6 +119,64 @@ private void doCodeActionsForUnknown(Diagnostic diagnostic, PropertiesModel docu
 		}
 	}
 
+	/**
+	 * Create code action fo
```
angelozerr: Fix javadoc

angelozerr:
The PR works very good, but similar name works if I write

`quarkus.log.level=WARNIN`

but if I write

`quarkus.log.level=WARNI`

It doesn't work (we have replace code action with all values instead of having similar code action with WARNING). 

In this case, 
`int threshold = Math.round(MAX_DISTANCE_DIFF_RATIO * reference.length());` is equals to 0. @fbricon have you an idea if we could manage the case with WARNI

WARNI

angelozerr:
I'm OK to merge this PR, but I wonder if `quarkus.log.level=WARNI` should work (I mean provide a code action which says `Did you mean 'WARNING' ?`) but it's a bug (or not?) from isSimilar method.

Perhaps we could merge this PR and fix this case (if it's possible) in an another issue. @fbricon what do you think?

fbricon:
Regardless of the levenshtein distance, there's only one enum that starts with the given value, it should be proposed

angelozerr:
> Regardless of the levenshtein distance, if there's only one enum that starts with the given value, it should be proposed

I agree with you, @xorye is there any chance that you work on this levenshtein issue in this PR? @fbricon if you have some idea how to fix it, you are welcome!

angelozerr:
> Regardless of the levenshtein distance

Ok I understand more your idea, we must check starts with at first in the isSimilar method. I think we should do that but only for value and not for property name.

xorye:
>  I think we should do that but only for value and not for property name.

Thanks for the feedback, I have it implemented like this: 

https://github.com/redhat-developer/quarkus-ls/blob/f5a3a219a4656c08e775582d32114c739ca1d61d/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/QuarkusCodeActions.java#L270-L272

I also added tests for this case

angelozerr:
@xorye it works like a charm, great job!

PR #148, opened 2019-11-18
Support for `@ConfigProperties`

### LINKED ISSUE #136: Support for `@ConfigProperties`
Quarkus provides the annotation `io.quarkus.arc.config.ConfigProperties ` that it is not supported. See 
https://quarkus.io/guides/application-configuration-guide#using-configproperties

We need to support it to collect the Quarkus properties in the JDT JDTQuarkusManager.

angelozerr:
Support for @ConfigProperties

Fixes #136


angelozerr:
I create a draft PR but there are some limitations:

 * code is ugly please don't review it:)
 * support only @Configproperties for class (not for interfaces)
 * doesn't support default value (when field is assigned with a value).
 * check some rules (ex: class must not extends class, etc) to generate the proper Quarkus properties.
 * write tests.

See https://quarkus.io/guides/config#using-configproperties for sample with ConfigProperties.

angelozerr:
@fbricon @xorye please review this PR. You can play with my test project  quarkus.jdt/com.redhat.quarkus.jdt.test/projects/maven/config-properties where I try to manage all cases with ConfigProperties annotation.

The only thing that I must implement is the definition for method used in interface (see IGreetingConfiguration)

angelozerr:
@fbricon @xorye find definition for method has been implemented, everything should work. Here a demo:

ConfigPropertiesDemo

Please give me your feedback.

xorye:
I found a corner case, but I don't think anyone would do this.

If this file:
https://github.com/angelozerr/quarkus-ls/blob/0829385ee5726afc88d0ab252336c806558af6c9/quarkus.jdt/com.redhat.quarkus.jdt.test/projects/maven/config-properties/src/main/java/org/acme/config/GreetingPublicFieldsConfiguration.java#L12-L14

Is changed to:
```
    public static class HiddenConfig {
        public List<String> recipients;
        public HiddenConfig config;
    }
```

then `quarkus/projectInfo` doesn't work

xorye:
> @fbricon @xorye find definition for method has been implemented, everything should work. Here a demo:
> 
> Please give me your feedback.

Its working well for me

angelozerr:
> I found a corner case, but I don't think anyone would do this.

Good catch, I l changed this case by ignoring the config property.

> then quarkus/projectInfo doesn't work

It should work now.

angelozerr:
> Its working well for me

Thanks for your feedback.

quarkus.jdt/com.redhat.quarkus.jdt.core/src/main/java/io/quarkus/arc/config/ConfigProperties.java
```diff
@@ -0,0 +1,22 @@
+package io.quarkus.arc.config;
```
fbricon: header

quarkus.jdt/com.redhat.quarkus.jdt.core/src/main/java/io/quarkus/deployment/bean/JavaBeanUtil.java
```diff
@@ -0,0 +1,61 @@
+package io.quarkus.deployment.bean;
```
fbricon: header

quarkus.jdt/com.redhat.quarkus.jdt.core/src/main/java/com/redhat/quarkus/jdt/core/JDTQuarkusManager.java
```diff
@@ -744,13 +982,18 @@ private static IAnnotation getAnnotation(IAnnotatable annotatable, String annota
...
-			if (annotationName.equals(annotation.getElementName())) {
+			if (isMatchAnnotation(annotation, annotationName)) {
 				return annotation;
 			}
 		}
 		return null;
 	}
 
+	private static boolean isMatchAnnotation(IAnnotation annotation, String annotationName) {
+		return annotationName.equals(annotation.getElementName())
```
fbricon: annotationName.equals(annotation.getElementName()) is already covered by annotationName.endsWith(annotation.getElementName())

quarkus.jdt/com.redhat.quarkus.jdt.test/src/main/java/com/redhat/quarkus/jdt/core/JDTQuarkusManagerConfigPropertiesTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+	private final int EXPECTED_PROPERTIES = EXPECTED_PROPERTIES_FROM_GREETING_INTERFACE
+			+ EXPECTED_PROPERTIES_FROM_PUBLIC_FIELDS + EXPECTED_PROPERTIES_FROM_GETTER
+			+ EXPECTED_PROPERTIES_FROM_NO_PREFIX + EXPECTED_PROPERTIES_FROM_STACK_OVERFLOW;
+
+	@Test
+	public void configProperties() throws Exception {
+
+		QuarkusProjectInfo infoFromJavaSources = getQuarkusProjectInfoFromMavenProject("config-properties",
+				QuarkusPropertiesScope.sources);
```
fbricon: any known examples of ConfigProperties to be read from binaries?

angelozerr: No it's like ConfigProperty, not found Quarkus extension which uses ConfigProperties annotations.

quarkus.jdt/com.redhat.quarkus.jdt.core/src/main/java/com/redhat/quarkus/jdt/core/JDTQuarkusManager.java
angelozerr: No it's not the same thing. annotation.getElementName() can have 2 values:

 * package + annotation name (it's the case when annotation comes from JAR)
 * just annotation name (it's the case when annotation comes from Java sources)

PR #166, opened 2019-12-11
Code action for ignore unknown property common parent

### LINKED ISSUE #150: Code action for ignoring unknown property keys using common parent key
As explained in this comment: https://github.com/redhat-developer/quarkus-ls/pull/141#issuecomment-554935317

xorye:
Fixes #150 

Demo:
Code action demo gif

This PR will never provide a code action for `quarkus.*`.

However in this case:
```
quarkus.unknown.test1=123
quarkus.unknown.test2=123
quarkus.unknown.test3=123
quarkus.unknown.test4=123
```

A code action for `quarkus.unknown.*` will be provided.

One question,
In this case:
```
a.b.c.d.e=123
a.b.c.d=123
```
If we hover over `a.b.c.d.e`, should we provide a code action per each common ancestor?
```
a.*
a.b.*
a.b.c.*
```

Or should we provide a code action only for the closest common parent?
```
a.b.c.*
```

Currently, this PR provides a code action per each common ancestor.

### DIFF
--- a/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/services/MicroProfileCodeActions.java
+++ b/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/services/MicroProfileCodeActions.java
@@ -17,14 +17,6 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
-import org.eclipse.lsp4j.CodeAction;
-import org.eclipse.lsp4j.CodeActionContext;
-import org.eclipse.lsp4j.CodeActionKind;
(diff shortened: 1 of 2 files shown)


fbricon:
Yeah code action for each common ancestor

fbricon:
toggling the pref from the command should add/remove the mirror cursor immediately

xorye:
Updated. The PR now uses a much more simpler solution. The PR does not look at all unknown properties in the application.properties file. Instead, the PR provides code actions without checking for common ancestors.

For example for this file with only one property

```
# Unknown property error
kubernetes.readiness-probe.initial-delay-seconds=20
```

The code actions are:
```
kubernetes.readiness-probe.initial-delay-seconds
kubernetes.readiness-probe.*
kubernetes.*
```

PR #178, opened 2020-01-04
Too many calls of microprofile/projectInfo

### LINKED ISSUE #174: Too many call of microprofile/projectInfo
When application.properties is opened, the collection of Quarkus properties (call of microprofile/projectInfo service) is done for each feature which requires the properties (completion, hover, validation).

It means if you open application.properties and you try to hover a property and pen completion, you will have 3 call of microprofile/projectInfo  service (one for validation, one for hover, one for completion). Once microprofile/projectInfo  result is done, there are no this problem since properties are cached.

To fix this issue, we should cache (for a given uri) the CompletableFuture which call microprofile/projectInfo and share this future between  all features (completion, validation, etc).

angelozerr:
Too many call of microprofile/projectInfo

Fixes #174

### DIFF
--- a/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/ls/MicroProfileProjectInfoCache.java
+++ b/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/ls/MicroProfileProjectInfoCache.java
@@ -38,7 +38,7 @@ import com.redhat.microprofile.ls.api.MicroProfileProjectInfoProvider;
  */
 class MicroProfileProjectInfoCache {
 
-	private final Map<String /* application.properties URI */, MicroProfileProjectInfoWrapper> cache;
+	private final Map<String /* application.properties URI */, CompletableFuture<MicroProfileProjectInfo>> cache;
 
 	private final MicroProfileProjectInfoProvider provider;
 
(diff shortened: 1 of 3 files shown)


angelozerr:
To test this PR, starts vscode-quarkus, open an application.properties and try to open completion  several times. You should see in Java log just one call of `microprofile/projectInfo`:

```
[Trace - 15:56:27] Sending request 'workspace/executeCommand - (2)'.
Params: {
    "command": "microprofile/projectInfo",
    "arguments": [
        {
```

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/ls/MicroProfileProjectInfoCache.java
```diff
@@ -187,32 +187,42 @@ public MicroProfileProjectInfoCache(MicroProfileProjectInfoProvider provider) {
...
-			return provider.getProjectInfo(params).thenApply(info ->
-			// information was loaded, update the cache
-			{
-				cache.put(params.getUri(), new MicroProfileProjectInfoWrapper(info));
-				return info;
-			});
+			CompletableFuture<MicroProfileProjectInfo> future = provider.getProjectInfo(params). //
+					thenApply(info -> new MicroProfileProjectInfoWrapper(info));
+			// cache the future.
+			cache.put(params.getUri(), future);
```
fbricon: where is the entry removed? looks like a memory leak waiting to happen

angelozerr: The remove is done when classpath changed. I think the remove should be done when project is closed. Do you think we should manage that?

fbricon: > The remove is done when classpath changed. I think the remove should be done when project is closed. Do you think we should manage that?

yes

fbricon: also, if document is deleted

angelozerr: To clarify your comment you wish i remove the cache when application.properties is removed and when project is closed. Is that? If it that i need to manage that on jdt ls extension side because diddelete doesnt exist in lsp

xorye:
This is really nice, with this PR, there is only one `microprofile/projectInfo` even if I start vscode-quarkus and ask for completion several times before it the completion results completely load.

But one question, when is `projectInfo.isCancelled()` true?
https://github.com/redhat-developer/quarkus-ls/blob/cecbe4c03d706cda8e0c935a07bd9008a7f07a6d/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/ls/MicroProfileProjectInfoCache.java#L192-L197

I naively thought that pressing Esc while the completion options were loading the first time, would cancel the completable future, but I realize that this is not the case.

angelozerr:
> But one question, when is projectInfo.isCancelled() true?

It's a very good question :) At first I did that to be sure that projectInfo future is valid (executed without errors and finished (not cancelled).

The idea is not to cancel the collection of Quarkus properties when completion, hover are canceled. The idea is to execute the process of collect of Quarkus properties when completion, hover, validation, etc required it. When completion is canceled, it should not cancel this process (you can see this case when you do Esc after opening completion).

Reproducing this use case is very hard (and not possible I think with vscode). This projectinfo future is canceled when JDT LS extension cancel the process of getProjectInfo. This is done in my another PR in https://github.com/redhat-developer/quarkus-ls/pull/180/files#diff-4845f95eb91f1df62337c6e24aafee10R127 (when monitor is canceled). 

How to cancel this progress monitor? The answer is that it's not possible in vscode, but it's possible with Eclipse IDE which provides progress view which show process and that you can stop it. Here a demo:

QuarkusProgress

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/ls/MicroProfileProjectInfoCache.java
angelozerr: @fbricon I create an issue to manage this cache with project events at https://github.com/redhat-developer/quarkus-ls/issues/184

xorye:
Wow! @angelozerr thank you for the explanation

angelozerr:
You are welcome @xorye !

PR #179, opened 2020-01-04
Download transitive dependencies of deployment JAR

angelozerr:
Download transitive dependencies of deployment JAR


angelozerr:
This PR fixes the bug with s2i and docker properties which should now be available when kubernetes dependencies is added.

image

This PR contains 2 tests for checking s2i an docker properties are generated..

xorye:
This PR works great on my machine

angelozerr:
Thanks @xorye for your feedback

PR #191, opened 2020-01-14
Add support for YAML configuration files

### LINKED ISSUE #112: Add support for YAML configuration files
Quarkus have the intention to support YAML configuration waith an `application.yaml` file. See [Add support for YAML configuration files](https://github.com/quarkusio/quarkus/issues/3904) issue.

We have 2 choices to support this feature:

 * implements our own YAML parser (like camel language server does that) but it's not perfect according @apupier .
 * consume yaml language server.

I create an issue about consuming [this yaml language server issue](https://github.com/redhat-developer/yaml-language-server/issues/190) and it seems that it's possible to contribute to the YAML Schema.

@JPinkney gived me a great explanation [how to contibute with YAML schema](https://github.com/redhat-developer/yaml-language-server/issues/190#issuecomment-540044607).

My idea is to generate a YAML schema from our Quarkus properties. If @fbricon and @xorye are OK I would like to explore this idea (for our next sprint).

angelozerr:
Add support for YAML configuration files

Fixes #112

### DIFF
--- a/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/commons/metadata/ItemMetadata.java
+++ b/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/commons/metadata/ItemMetadata.java
@@ -105,6 +105,11 @@ public class ItemMetadata extends ItemBase {
 		return phase == CONFIG_PHASE_BUILD_AND_RUN_TIME_FIXED || phase == CONFIG_PHASE_RUN_TIME;
 	}
 
+	public boolean isStringType() {
+		return "java.lang.String".equals(getType()) || //
+				"java.util.Optional<java.lang.String>".equals(getType());
+	}
+
 	public boolean isBooleanType() {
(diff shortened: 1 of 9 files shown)


angelozerr:
Here the current status of the YAML support:

YamlDemo

angelozerr:
By waiting for [answer in this issue,](https://github.com/quarkusio/quarkus/issues/6584) This PR doesn't support:

 * mp-rest (it supports it but not sure it works correctly)
 * mapped property.

angelozerr:
@fbricon @xorye I have experimented default value, and now when I take care of my JSON Schema, it generates all properties which have default values although property is not required:

YamlDefaultValueDemo

What do you think about that? Should take care of default value?

angelozerr:
@fbricon my last commit set the JSON Schema as lenient (I add a lenient parameter to configure the generation of the JSON Schema). Today the lenient parameter is set to true, but in the future we could compute this lenient value with a settings.

For the default value, I disable it to avoid the usecase https://github.com/redhat-developer/quarkus-ls/pull/191#issuecomment-575142239

PR #192, opened 2020-01-15
Fix NPE in CompletableFuture on LSP requests

xorye:
Fixes https://github.com/redhat-developer/vscode-quarkus/issues/185

I added this to the jdt.ls extension here : 
https://github.com/xorye/quarkus-ls/blob/277f609b0bbb07b4900aa354a6ba99a69331ad27/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/core/ls/MicroProfileDelegateCommandHandler.java#L132-L134

because the original exception here: https://github.com/redhat-developer/quarkus-ls/blob/master/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/core/PropertiesManager.java#L88

was being thrown while inside of a Java Job. Because of that, the `exceptionally` stage wasn't being reached on the client side here:
https://github.com/xorye/quarkus-ls/blob/277f609b0bbb07b4900aa354a6ba99a69331ad27/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/ls/MicroProfileProjectInfoCache.java#L209-L212

### DIFF
--- a/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/core/ls/MicroProfileDelegateCommandHandler.java
+++ b/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/core/ls/MicroProfileDelegateCommandHandler.java
@@ -127,6 +127,15 @@ public class MicroProfileDelegateCommandHandler implements IDelegateCommandHandl
 		} catch (InterruptedException e) {
 			LOGGER.log(Level.WARNING, "Error while joining MicroProfile properties collector job", e);
 		}
+		
+		Exception jobException = (Exception) job.getResult().getException();
+		if (jobException != null) {
+			if (jobException.getCause() != null) {
+				throw (Exception) jobException.getCause();
+			}
+			throw jobException;
+		};
+		
 		return projectInfo[0];
 	}
 
--- a/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/ls/ApplicationPropertiesTextDocumentService.java
+++ b/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/ls/ApplicationPropertiesTextDocumentService.java
@@ -172,7 +172,7 @@ public class ApplicationPropertiesTextDocumentService extends AbstractTextDocume
 		MicroProfileProjectInfoParams projectInfoParams = createProjectInfoParams(params.getTextDocument());
 		return getProjectInfoCache().getProjectInfo(projectInfoParams).thenComposeAsync(projectInfo -> {
 			if (projectInfo.getProperties().isEmpty()) {
-				return null;
+				return CompletableFuture.completedFuture(null);
 			}
 			// then get the Properties model document
 			return getPropertiesModel(params.getTextDocument(), (cancelChecker, document) -> {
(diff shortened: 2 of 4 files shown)


fbricon:
Can we get unit tests?

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/core/ls/MicroProfileDelegateCommandHandler.java
```diff
@@ -127,6 +127,10 @@ private static MicroProfileProjectInfo getMicroProfileProjectInfo(List<Object> a
 		} catch (InterruptedException e) {
 			LOGGER.log(Level.WARNING, "Error while joining MicroProfile properties collector job", e);
 		}
+		
+		if (!job.getResult().isOK()) {
+			throw new UnsupportedOperationException(String.format("Cannot find MicroProfileProjectInfo for '%s'", params.getUri()));
```
angelozerr: I would like to avoid throwing a new exception here since job could have anoter error. What about using `job.getResult().getException()` (or perhaps `job.getResult().getException().getCause()` to get the original error?

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/ls/ApplicationPropertiesTextDocumentService.java
```diff
@@ -243,7 +243,7 @@ public void didSave(DidSaveTextDocumentParams params) {
 		MicroProfileProjectInfoParams projectInfoParams = createProjectInfoParams(params.getTextDocument());
 		return getProjectInfoCache().getProjectInfo(projectInfoParams).thenComposeAsync(projectInfo -> {
 			if (projectInfo.getProperties().isEmpty()) {
-				return null;
+				return CompletableFuture.completedFuture(null);
```
angelozerr: You did that to avoid NPE in thenApply(info

angelozerr:
> Can we get unit tests?

You mean for MicroProfile LS side? If it that I think we could do that?

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/ls/ApplicationPropertiesTextDocumentService.java
xorye: I noticed that `return null` also caused an NPE so I made this change.

Where is the `thenApply(info)` you are referring to?

Is `thenApply(info)` being called by the caller for `public CompletableFuture<List<Either<Command, CodeAction>>> codeAction(CodeActionParams params)`?

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/core/ls/MicroProfileDelegateCommandHandler.java
xorye: Thank you, what do you think about the new code?

```
		Exception jobException = (Exception) job.getResult().getException();
		job.getResult().getException().getCause();

		if (jobException != null) {
			if (jobException.getCause() != null) {
				throw (Exception) jobException.getCause();
			}
			throw jobException;
		};
```

xorye:
I added a new unittest in `MicroProfileProjectInfoCacheTest.java` that simulates the case where the jdt.ls extension throws an exception when trying to find projectInfo

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/ls/ApplicationPropertiesTextDocumentService.java
angelozerr: Forget my comments.

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/core/ls/MicroProfileDelegateCommandHandler.java
angelozerr: yes it was my idea.

Issue #2, opened 2019-08-01
Completion support for value in application.properties


angelozerr:
Completion support for value in application.properties

`quarkus.log.console.async.overflow = |  ` <-- show BLOCK, DISCARD list
`quarkus.datasource.enable-metrics = |  ` <-- show true, false list

 * for enum
 * for boolean
 * for classes which implement an interface (checks if this case exists?)

angelozerr:
@xorye I think it's not an hard issue, I assign you, if you need assistance please tell me.

angelozerr:
> If these exist, would they be in ExtendedConfigDescriptionBuildItem.enums

I would like to avoid doing that since searching implementation classes could be long. More if you add in Java source a new implementation of class, it means that Quarkus properties should be computed again.

I think we should provide a new method in Quarkus JDT LS extension which returns list of Java classes which implement a given Java class and manage completion by calling it.

PR #197, opened 2020-01-22
Missing enum value completion/validation for optional enum

### LINKED ISSUE #175: Missing enum value completion/validation for optional enum
For the following properties: 
```
quarkus.hibernate-search.elasticsearch.automatic-indexing.synchronization-strategy
quarkus.hibernate-search.elasticsearch.backends.{*}.automatic-indexing.synchronization-strategy
quarkus.hibernate-search.elasticsearch.backends.{*}.index-defaults.lifecycle.strategy
quarkus.hibernate-search.elasticsearch.backends.{*}.index-defaults.lifecycle.required-status
quarkus.hibernate-search.elasticsearch.backends.{*}.indexes.{*}.lifecycle.strategy
quarkus.hibernate-search.elasticsearch.backends.{*}.indexes.{*}.lifecycle.required-status
```
each of these properties expect a `java.util.Optional<ENUM VALUE>` as its value. However, when invoking completion for the value, there are no completion options for the enum value.

Also, there is no value validation, meaning that there is no error diagnostic if the value is not a valid enum value.

angelozerr:
Missing enum value completion/validation for optional enum

Fixes #175

### DIFF
--- a/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/commons/metadata/ConfigurationMetadata.java
+++ b/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/commons/metadata/ConfigurationMetadata.java
@@ -49,7 +49,7 @@ public class ConfigurationMetadata {
 	 *         otherwise.
 	 */
 	public ItemHint getHint(ItemMetadata property) {
-		return getHint(property.getName(), property.getType());
+		return getHint(property.getName(), property.getHintType());
 	}
 
 	/**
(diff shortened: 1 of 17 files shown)


angelozerr:
Here a demo with `quarkus.datasource.transaction-isolation-level` which have optional value type 
`java.util.Optional<io.agroal.api.configuration.AgroalConnectionFactoryConfiguration.TransactionIsolation>` OptionalEnumDemo

xorye:
If I go to definition for the enum value for `quarkus.datasource.transaction-isolation-level`, I see the following:
image

Therefore, I think we should have `repeatable-read` as a completion option:
image

image

fbricon:
we no longer have value completion for quarkus.log.console.async.overflow= (see https://github.com/redhat-developer/quarkus-ls/issues/2)
Also, enum values don't seem proposed in application.yaml

angelozerr:
Sorry guys my PR was so bad. It worked because values for isolation level was hard coded. (the problem described in https://github.com/redhat-developer/quarkus-ls/pull/197#issuecomment-577311311)

All problem should be fixed:

 * the REPEATABLE_READ should be shown in the completion
 * quarkus.log.console.async.overflow property (No optional enum) should work
 * quarkus.datasource.transaction-isolation-level (optional enum) should work on YAML too

BUT we don't manage correctly value for enum. According the documentation we should have `repeatable-read` instead of having `REPEATABLE_READ `

PR #201, opened 2020-01-27
Value from Java enum are not correct

### LINKED ISSUE #198: Value from Java enum are not correct
Today value from Java enum are used in the application.properties file 

For instance we have:

`quarkus.log.console.async.overflow = BLOCK
`
where `BLOCK` is the Java enum name.

But we should have (see discussion at https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/What.20is.20the.20case.20sensitive.20rule.20for.20value.20property.20with.20enum)

`quarkus.log.console.async.overflow = block 
`
We must study the rule of convert of Java enum value (see https://quarkus.io/guides/writing-extensions#enhanced-conversion)

It seems that `BLOCK` and `block` are correct for `quarkus.log.console.async.overflow`

We have the same problem with `quarkus.datasource.transaction-isolation-level`. Our allowed value are for instance `REPEATABLE_READ` although documentation 
See https://quarkus.io/guides/all-config#quarkus-agroal_quarkus.datasource.transaction-isolation-level says `repeatable-read`

angelozerr:
Value from Java enum are not correct

Fixes #198

### DIFF
--- a/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/commons/metadata/ConverterKind.java
+++ b/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/commons/metadata/ConverterKind.java
@@ -1,0 +1,78 @@
+/*******************************************************************************
+* Copyright (c) 2019 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.microprofile.commons.metadata;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import io.quarkus.runtime.util.StringUtil;
+
+/**
+ * Converter kind.
+ * 
+ * @author Angelo ZERR
+ *
+ */
+public enum ConverterKind {
+
+	KEBAB_CASE(1), VERBATIM(2);
+
+	private final int value;
+
+	ConverterKind(int value) {
+		this.value = value;
+	}
+
+	public int getValue() {
+		return value;
+	}
+
+	public static ConverterKind forValue(int value) {
+		ConverterKind[] allValues = ConverterKind.values();
+		if (value < 1 || value > allValues.length)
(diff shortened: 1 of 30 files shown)


angelozerr:
Here a demo with kebab_case and verbatim converter with a Quarkus property and a MicroProfile property which are bound to the same Java enumeration:

EnumKebaCaseDemo

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/commons/metadata/ConverterKind.java
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * Convert the given value to kebab case.
+	 * 
+	 * @param value
+	 * @return
+	 * @see https://github.com/quarkusio/quarkus/blob/1457e12766d46507674f8e8d8391fc5a5e8f0103/core/runtime/src/main/java/io/quarkus/runtime/configuration/HyphenateEnumConverter.java#L54
+	 */
+	private static String hyphenate(String value) {
+		StringBuffer target = new StringBuffer();
```
fbricon: StringBuilder

angelozerr: At fist it's a copy/paste code from Quarkus. And I cannot used StringBuilder since `Matcher.appendReplacement` supports only StringBuffer as parameter

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/io/quarkus/runtime/util/StringUtil.java
```diff
@@ -0,0 +1,220 @@
+package io.quarkus.runtime.util;
```
fbricon: please add link to the original source, and document the original license

angelozerr: I add links to the original Quarkus StringUtil class.

angelozerr: License Apache 2 has been added.

Issue #189, opened 2020-01-10
Missing mp-rest properties


fbricon:
According to the [microprofile rest client spec](https://download.eclipse.org/microprofile/microprofile-rest-client-1.3.3/microprofile-rest-client-1.3.3.pdf), the following properties should be supported: 

> The values of the following properties will be provided via MicroProfile Config:
• com.mycompany.remoteServices.MyServiceClient/mp-rest/url: The base URL to use for this service,
the equivalent of the baseUrl method. This property (or */mp-rest/uri) is considered required,
however implementations may have other ways to define these URLs/URIs.
• com.mycompany.remoteServices.MyServiceClient/mp-rest/uri: The base URI to use for this service,
the equivalent of the baseUri method. This property (or */mp-rest/url) is considered required,
however implementations may have other ways to define these URLs/URIs. This property will
override any baseUri value specified in the @RegisterRestClient annotation.
• com.mycompany.remoteServices.MyServiceClient/mp-rest/scope: The fully qualified classname to a
CDI scope to use for injection, defaults to javax.enterprise.context.Dependent as mentioned
above.
• com.mycompany.remoteServices.MyServiceClient/mp-rest/providers: A comma separated list of
fully-qualified provider classnames to include in the client, the equivalent of the register
method or the @RegisterProvider annotation.
• com.mycompany.remoteServices.MyServiceClient/mprest/providers/com.mycompany.MyProvider/priority will override the priority of the provider for
this interface.
• com.mycompany.remoteServices.MyServiceClient/mp-rest/connectTimeout: Timeout specified in
milliseconds to wait to connect to the remote endpoint.
• com.mycompany.remoteServices.MyServiceClient/mp-rest/readTimeout: Timeout specified in
milliseconds to wait for a response from the remote endpoint.
Implementations may support other custom properties registered in similar fashions or other ways.
The url property must resolve to a value that can be parsed by the URL converter required by the
MicroProfile Config spec. Likewise, the uri property must resolve to a value that can be parsed by
the URI converter. If both the url and uri properties are declared, then the uri property will take
precedence.
The providers property is not aggregated, the value will be read from the highest property
17
ConfigSource.

right now we only have 3 of those.

angelozerr:
@jclingan I would lile to manage the last MP Rest property

```
com.mycompany.remoteServices.MyServiceClient/mprest/providers/com.mycompany.MyProvider/priority will override the priority of the provider for
this interface.
```
I mean 

[...]

Issue #203, opened 2020-01-28
Support for complex properties


angelozerr:
There are some property which are complex composed by java classes (which implement some interfaces), package names, etc

Here some usecase:

# log category

See https://quarkus.io/guides/logging#category-example

```
quarkus.log.category."io.quarkus.smallrye.jwt".level=TRACE
quarkus.log.category."io.undertow.request.security".level=TRACE
``` 

In this sample,we should have completion for package names (after category) and validation.

# mp rest priority

`com.mycompany.remoteServices.MyServiceClient/mprest/providers/com.mycompany.MyProvider/priority = 1`

In this sample we should have completion for mp rest provider (after /providers). mp rest provider is a class which implements some interfaces. We should have too definition on this provider class, hover and validation.

PR #204, opened 2020-01-28
Support for <mp-rest-client-class-name>/mp-rest/providers/*/priority

angelozerr:
Support for /mp-rest/providers/*/priority

See https://github.com/redhat-developer/quarkus-ls/issues/189

### DIFF
--- a/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/core/providers/MicroProfileRegisterRestClientProvider.java
+++ b/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/core/providers/MicroProfileRegisterRestClientProvider.java
@@ -95,7 +95,7 @@ public class MicroProfileRegisterRestClientProvider extends AbstractAnnotationTy
 
 	private static final String[] ANNOTATION_NAMES = { REGISTER_REST_CLIENT_ANNOTATION };
 
-	private static final String MP_REST_CLASS_REFERENCE_TYPE = "${mp.register.rest.client.class}";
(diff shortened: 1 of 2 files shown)


angelozerr:
This PR supports last mp-rest property `<mp-rest-client-class-name>/mp-rest/providers/<mp-provider-class-name>/priority`. You will have completion about this property. But as it's a complex property you need to add in your settings this validation rule:

```json
"quarkus.tools.validation.unknown.excluded": [
        "*/mp-rest/providers/*/priority"
    ]
```

to avoid having an error. When https://github.com/redhat-developer/quarkus-ls/issues/203 will be supported, it will fix that without this settings.

I will do a PR in vscode-quarkus to explain this settings.

angelozerr:
See my pr https://github.com/redhat-developer/vscode-quarkus/pull/203 about settings

xorye:
This PR works/looks good for me

I know this isn't part of this PR but I don't understand what's being done with `itemHint` here? 
https://github.com/redhat-developer/quarkus-ls/blob/db1039a9f8b5eb20c95dcaabf140c4b9ab11287f/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/core/providers/MicroProfileRegisterRestClientProvider.java#L162-L177
It's being created and a value is being added to it but it looks like the code doesn't do anything with it after that?

angelozerr:
> Oops, nevermind about this, I noticed getItemHint() adds the hint to the hintsCache

Yes that's it :)

If you think we can merge the PR, please do it!

xorye:
@angelozerr one small thing, is it possible to escape the `<` and `>` characters in your commit message? 

So that we could have `Support for <mp-rest-client-class-name>/mp-rest/providers/*/priority` instead of `Support for /mp-rest/providers/*/priority` ?

angelozerr:
@xorye commit has been changed

xorye:
Could we try the Travis test again?

PR #209, opened 2020-02-04
Add support for MicroProfile Fault Tolerance

angelozerr:
Add support for MicroProfile Fault Tolerance

See https://github.com/redhat-developer/vscode-quarkus/issues/174


angelozerr:
Please don't review the code for the moment. To test this PR, I suggest you play with the microprofile-fault_tolerance project which is stored in the test project from this PR (I must write test) 

You can see the demo:

MPFTDemo

angelozerr:
There are some tests now, the PR can be reviewed.

microprofile.jdt/com.redhat.microprofile.jdt.test/src/main/java/com/redhat/microprofile/jdt/core/MicroProfileFaultToleranceTest.java
```diff
@@ -0,0 +1,63 @@
...
+import static com.redhat.microprofile.jdt.internal.core.MicroProfileAssert.assertPropertiesDuplicate;
+import static com.redhat.microprofile.jdt.internal.core.MicroProfileAssert.p;
+
+import org.junit.Test;
+
+import com.redhat.microprofile.commons.MicroProfileProjectInfo;
+import com.redhat.microprofile.commons.MicroProfilePropertiesScope;
+
+/**
+ * Test collect MicroProfile properties for MicroProfile Fault Tolerance
```
fbricon: collection of

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/core/providers/MicroProfileFaultToleranceProvider.java
```diff
@@ -0,0 +1,328 @@
...
+					// type
+					String methodResultTypeName = getResolvedResultTypeName(method);
+					IType returnType = findType(method.getJavaProject(), methodResultTypeName);
+					String type = getPropertyType(returnType, methodResultTypeName);
+
+					try {
+						utils.discoverSource(method.getClassFile(), new NullProgressMonitor());
+					} catch (CoreException e) {
+						// TODO Auto-generated catch block
+						e.printStackTrace();
```
fbricon: use logger / send to client

microprofile.jdt/com.redhat.microprofile.jdt.test/src/main/java/com/redhat/microprofile/jdt/core/MicroProfileFaultToleranceTest.java
angelozerr: @fbricon fixed

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/core/providers/MicroProfileFaultToleranceProvider.java
angelozerr: @fbricon code has been changed, and now it's fixed.

fbricon:
please fix the conflict

angelozerr:
> please fix the conflict

done

xorye:
From https://github.com/eclipse/microprofile-fault-tolerance/blob/master/spec/src/main/asciidoc/configuration.asciidoc:

>If the parameters for a particular annotation need to be configured with the same value for a particular class, use the config property \<classname\>/\<annotation\>/\<parameter\> for configuration. For an instance, use the following config property to override all maxRetries for Retry specified on the class MyClient to 100.

Should I be able to see, for example, `org.acme.MyClient/Retry/maxDuration` as a completion option?

fbricon:
CI Build failed: 
  MicroProfilePropertiesListenerManagerTest.classpathChanged:92 expected:<1> but was:<2>

fbricon:
> Should I be able to see, for example, `org.acme.MyClient/Retry/maxDuration` as a completion option?

correct. Means fault tolerance support is incomplete

angelozerr:
> correct. Means fault tolerance support is incomplete

Please retry it. I changed my strategy to collect MP Fault Tolerance properties and now I provide only MP Fault Tolerance annotation (Asynchrounous, Retry, etc) only if one method or one class of the project uses it.

For instance 

If we have one class in the project like this:

```java
@Asynchrounous
public class MyClient {

}
```

`Asynchrounous/enabled` and `MyClient/Asynchrounous/enabled` should be availabled but `Retry/enabled` will not be availabled .

xorye:
Awesome, I am now getting `<classname>/<annotation>/<parameter>` completion options for the annotations that are present in the project.

A few things I noticed: 
I get an unknown property error/warning with the `MP_Fault_Tolerance_NonFallback_Enabled` property
image

After playing around with the PR (I was adding new methods in `MyClient.java` with different annotations), I noticed duplicate completion options:
image

I don't know how much this would help, but here is the server trace:
https://gist.github.com/xorye/5c02361f66f853922ae3431db1bad1ba

#### Enhancement ideas
I have some enhancement ideas, maybe we could have support for the `java.time.temporal.ChronoUnit` enum?

For example, `org.acme.MyClient/serviceA/Retry/delayUnit` expects a `java.time.temporal.ChronoUnit` enum value.

Also, go to definiton for  `<classname>/<annotation>/<parameter>` properties would be nice.

Perhaps I should create new issues for these ideas?

angelozerr:
> I get an unknown property error/warning with the `MP_Fault_Tolerance_NonFallback_Enabled `property

Good catch, this property must be hard coded, it's fixed.

> After playing around with the PR (I was adding new methods in MyClient.java with different annotations), I noticed duplicate completion options:

It should fixed for <annotation>/<parameter> (even with duplicate method), but if you duplicate methods and save your java files, you will have duplicate properties for <classname>/methodname>/<annotation>/<parameter>. I could fix that by caching it, but this case should not occurs if you have a java valid file.

> I have some enhancement ideas, maybe we could have support for the java.time.temporal.ChronoUnit enum?

Good catch! I forget to manage type of parameter. As ChronoUnit it uses the standard mechanism with hints. It should be fixed now.

> Also, go to definiton for <classname>/<annotation>/<parameter> properties would be nice.
Good catch again! It should be fixed now.

Please review it again.

xorye:
The new fixes are working great on my machine.

One more thing I found: 

In MyClient.java, we have the following:
```
@Asynchronous
@Bulkhead
public class MyClient {
```

And as expected, I am getting completion options for `Bulkhead/<parameter>` and `Asynchronous/enabled`. However, if I add the `@CircuitBreaker` annotation like so:

```
@Asynchronous
@Bulkhead
@CircuitBreaker
public class MyClient {
```

and save the file, I do not get `CircuitBreaker/<parameter>` completion options. But, I do get the new completion options if I restart VS Code.

angelozerr:
> and save the file, I do not get CircuitBreaker/<parameter> completion options. 

Indeed! It should be fixed now.

PR #210, opened 2020-02-04
Support ConfigProperties for 1.2

### LINKED ISSUE #200: Support ConfigProperties for 1.2
We should change the strategy for generate properties with ConfigProperties on JDT LS for 1.2 https://github.com/quarkusio/quarkus/wiki/Migration-Guide-1.2#configproperties

@fbricon do you think we should support all version of Quarkus?

angelozerr:
Support ConfigProperties for 1.2

Fixes #200

Signed-off-by: azerr azerr@redhat.com


angelozerr:
This PR support ConfigProperties 1.2 with naming strategy.

You can play with the `com.redhat.microprofile.jdt.test\projects\maven\config-properties` project with the the 3 properties:

```
greetingPublicFields.hidden-config.recipients = 
greeting-verbatim-naming-strategy.hiddenConfig.recipients = 
quarkus.arc.config-properties-default-naming-strategy = kebab-case
```

Please note the changes of `quarkus.arc.config-properties-default-naming-strategy` doesn't recompute the properties. You must save a Java file (to recompute the properties from sources or better close and reopen the vscode).

If you save a Java file and properties are not computed, I think it's a problem with time of compilation. Indeed teh JDT LS extension search the value of `quarkus.arc.config-properties-default-naming-strategy` inside `target/classes/application.properties` (for maven) and not from `src/main/resources/application.properties` (file where you edit the properties).

If compilation takes time, `target/classes/application.properties`  will not be updated correctly (it will not have the same value than `src/main/resources/application.properties` ) and collection properties will use a bad value for `quarkus.arc.config-properties-default-naming-strategy`

xorye:
I created this new file (in the example from https://github.com/quarkusio/quarkus/wiki/Migration-Guide-1.2#configproperties):
```
package org.acme.config;

import io.quarkus.arc.config.ConfigProperties;

@ConfigProperties
public class SomeConfig {
    public String fooBar;
}
```
and I have been experiencing duplicate values in completion.
It is difficult to reproduce this issue (I'm not sure what might have caused it?)
image

Also, I'm having trouble getting verbatim properties even though I have set `quarkus.arc.config-properties-default-naming-strategy=verbatim` and restarted VS Code. I also ensured that the config was set in `target/classes/application.properties` as well.
image

image

angelozerr:
> Also, I'm having trouble getting verbatim properties even though I have set quarkus.arc.config-properties-default-naming-strategy=verbatim and restarted VS Code. I also ensured that the config was set in target/classes/application.properties as well.

It should be fixed now (I write test for that), please retry it.

angelozerr:
> and I have been experiencing duplicate values in completion.
It is difficult to reproduce this issue (I'm not sure what might have caused it?)

I cannot reproduce it -(

If you find a usecase which causes the problem, it should be really cool.

angelozerr:
@xorye if you can reproduce the duplicate problem, could you attach please the traces from Java & Quarkus Language Support

Issue #196, opened 2020-01-21
ConfigProperty name hover should take application.yaml into consideration


xorye:
This feature https://github.com/redhat-developer/quarkus-ls/pull/151 only checks for config values in `application.properties`.

It would be nice if it could also check `application.yaml`.

angelozerr:
Indeed,and we need too get the http.server.port too from yaml file

angelozerr:
Fixed with https://github.com/redhat-developer/quarkus-ls/pull/220

PR #220, opened 2020-02-10
Parse application.yaml for REST URL codelens

### LINKED ISSUE #195: REST endpoint CodeLenses should take application.yaml into consideration
If quarkus.http.port is set in application.yaml, it should be used. Proper priorities should be used when both yaml and properties files are found in the classpath (not sure which one wins)

xorye:
Fixes #195 

If `application.yaml` and `application.properties` are both present, priority was given to `application.yaml` (Source: https://quarkus.io/guides/config#add-yaml-config-support)

image


xorye:
I think I need to setup Maven to copy the snakeyaml dependency to the lib folder.

What I did to make this PR was to manually download the jar here: https://mvnrepository.com/artifact/org.yaml/snakeyaml/1.25

and place it in `microprofile.jdt/com.redhat.microprofile.jdt.core/lib/`

I will take care of this tomorrow.

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/core/project/YAMLConfigSource.java
```diff
@@ -0,0 +1,70 @@
...
+import org.eclipse.jdt.core.IJavaProject;
+import org.yaml.snakeyaml.Yaml;
+
+/**
+ * YAML config source implementation
+ * 
+ * @author dakwon
+ *
+ */
+public class YAMLConfigSource extends AbstractConfigSource<LinkedHashMap<String, Object>> {
```
angelozerr: Use Map (which is an interface) instead of LinkedHashMap (which is an implementation)

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/core/project/YAMLConfigSource.java
```diff
@@ -0,0 +1,70 @@
...
+
+	public YAMLConfigSource(String configFileName, IJavaProject javaProject) {
+		super(configFileName, javaProject);
+	}
+
+	@Override
+	protected LinkedHashMap<String, Object> loadConfig(InputStream input) throws IOException {
```
angelozerr: Use Map

xorye:
In my latest commit, I am using the maven dependency plugin to download the snakeyaml jar into `com.redhat.microprofile.jdt.core/lib/`. 

But, the way I do it is slightly different than how jdt.ls downloads the remark library.
jdt.ls has the plugin in the parent `pom.xml`, while I have the plugin in the `pom.xml` located under `com.redhat.microprofile.jdt.core/`.

This is because I realized that if I had placed it in the parent pom, the snakeyaml jar will  not only be placed into `lib/` directories under `com.redhat.microprofile.jdt.core`, but also `com.redhat.microprofile.jdt.site`, `com.redhat.microprofile.jdt.test` and `com.redhat.microprofile.jdt.quarkus`.

Since we only need it in `com.redhat.microprofile.jdt.core`, I opted to put the plugin information in 
`com.redhat.microprofile.jdt.core/pom.xml`.

xorye:
I noticed that this PR also fixes #196 as well.

I will soon write tests for #196 in this PR

xorye:
My most recent commit adds a test for https://github.com/redhat-developer/quarkus-ls/issues/196. If this PR is okay to merge, I will squash the two commits together.

xorye:
The test has been updated once more

angelozerr:
Works like a charm, good job @xorye !

PR #227, opened 2020-02-18
Empty completion after an error from microprofile/projectInfo

### LINKED ISSUE #228: Empty completion after an error from microprofile/projectInfo
When completion (or other feature like validation, hover, etc) is called it consume the JDT LS microprofile/projectInfo 
to collect properties. When `microprofile/projectInfo` throws an error and completion is redone, the properties are empty, even if 
classpath changed.

This poblem occurs because of [this fix](https://github.com/redhat-developer/quarkus-ls/pull/192) (when application.properties doesn't belong to a Java project). The fix cache the completable future after 
CompletableFuture.exceptions is called (when error is catch). The completable future which must be cached must be done before the exception is cacth
(before CompletableFuture.exceptions is called).

By doing that, we need to fix the case of application.properties doesn't belong to a Java project which must retun an empty project info
instead of throwing an error, otherwise, it will call every time the JDT LS extension microprofile/projectInfo  each time completion is triggered 
(should not be done since application.properties doesn't belong to a Java project).

An another case is when application.properties belongs to an Eclipse project but not in Java project. The JDT LS extension must return a project info with 
the proper project location uri. By doing that, when user add a pom.xml, a classpath changed will be fired and a microprofile/propertiesChanged will be fired
and MicroProfile LS will reload the properties.

angelozerr:
Empty completion after an error from microprofile/projectInfo

Fixes #228

### DIFF
--- a/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/commons/MicroProfileProjectInfo.java
+++ b/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/commons/MicroProfileProjectInfo.java
@@ -9,6 +9,8 @@
 *******************************************************************************/
 package com.redhat.microprofile.commons;
 
+import java.util.Collections;
+
 import com.redhat.microprofile.commons.metadata.ConfigurationMetadata;
 
 /**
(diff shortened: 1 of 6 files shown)


xorye:
I'm experiencing some trouble testing this PR:

I created an Eclipse project (not a Java project) like this:
image

I then opened the project in VS Code, created `src/main/resources/application.properties` and invoked completion in the `application.properties` file.

However, the microprofile jdt.ls extensions believes the `application.properties` file is not an eclipse project:
image

Did I do something wrong?

xorye:
This PR seems good.

To simulate an error in the microprofile/projectInfo call, I just did the following:
```
private boolean once = false;
.
.
.

if (!once) {
    once = true;
    throw new UnsupportedOperationException(String.format("Cannot find IFile for '%s'", params.getUri()));
}			
```

angelozerr:
> To simulate an error in the microprofile/projectInfo call, I just did the following:

Indeed it's hard to have some errors on JDT LS side (I had some errors when I edited pom.xml and saved it sereral times quickly (but it was hard to reproduce it). Your "once" code is exactly a good test.

Issue #229, opened 2020-02-20
Split Quarkus / MicroProfile support


angelozerr:
Today we have MicroProfile / Quarkus support for 

 * `application.properties` (completion, hover, validation, code action) for MicroProfile / Quarkus properties.
 * java files with codelens (show URL) and hover (show value of ConfigProperty defined in application.properties)

# application.properties

The `application.properties` support is proper managed 

 * on JDT LS extension side a propertiesProvider extension point is used to collect any properties (MicroProfile ConfigProperty, MP Rest, Quarkus ConfigRoot, Quarkus kubernetes, etc). We have 2 plugins which manages that (one for MicroProfile and one for Quarkus). For this part, there are nothing to do.

 * on MicroProfile LS side, the main code is not linked to Quarkus (just profile and some JSON file which manage Level for value which is specific for Quarkus). For the moment I think we can live with that.

So I think for  `application.properties` support, there are nothing to do.

# java codelens, hover

For java codelens hover support, it's not very clean:

* on JDT LS extension side 
  * codelens for JAX-RS uses `quarkus.http.server.port` property to retrieve server port although this property is unknown for MicroProfile. We need to provide extensibility for codelens.
  * for hover, it supports only MicroProfile ConfigProperty but in the future we would like to support Quarkus ConfigRoot, so we need to provide extensibility for hover too.

 * on MicroProfile LS it's generic (not linked to Quarkus), but MicroProfile LS manages .properties and .java files both. It causes the problem with java formatter. See https://github.com/redhat-developer/vscode-quarkus/issues/166 because MicroProfile LS returns the format capability to true (because it can format application.properties) but it is not able to format java files.

## Solutions

There are 3 solutions to fix the extensibility and problem with MicroProfile LS (java formatter):

### Solution 1: Keep MicroProfile LS to manage java

* on MicroProfile LS side, we keep the problem https://github.com/redhat-developer/vscode-quarkus/issues/166.
  * Pros: 
     * nothing to do 
     * one process 
     * share data (like projectInfo but today we don't need that)
     * will continue to work with IJ
  * Cons : problem with java formatter
* on JDT LS side, we must provide MicroProfile extensions points for codelens, hover, and diagnostics too to support the issue https://github.com/redhat-developer/quarkus-ls/issues/217

### Solution 2: Create new MicroProfile LS for java

* on MicroProfile LS side, we create a new MicroProfile LS for java
  * Pros : 
    * fix java formatter and future support
[...]

PR #232, opened 2020-02-28
Provide codeLens participant

angelozerr:
Provide codeLens participant

See https://github.com/redhat-developer/quarkus-ls/issues/229

### DIFF
--- a/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/core/java/IJavaCodeLensParticipant.java
+++ b/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/core/java/IJavaCodeLensParticipant.java
@@ -15,18 +15,18 @@ import java.util.List;
 
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
-import org.eclipse.lsp4j.Diagnostic;
+import org.eclipse.lsp4j.CodeLens;
 
 /**
- * Java diagnostics participants API.
+ * Java codeLens participants API.
  * 
  * @author Angelo ZERR
  *
  */
-public interface IJavaDiagnosticsParticipant {
+public interface IJavaCodeLensParticipant {
 
 	/**
-	 * Returns true if diagnostics must be collected for the given context and false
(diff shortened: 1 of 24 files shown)


microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/core/java/JavaFeatureDefinition.java
```diff
@@ -27,62 +29,65 @@
...
 
-	private final IJavaHoverParticipant hoverParticipant;
+	private final IJavaCodeLensParticipant codeLensParticipant;
 	private final IJavaDiagnosticsParticipant diagnosticsParticipant;
+	private final IJavaHoverParticipant hoverParticipant;
 
-	public JavaFeatureDefinition(IJavaHoverParticipant hoverParticipant,
-			IJavaDiagnosticsParticipant diagnosticsParticipant) {
-		this.hoverParticipant = hoverParticipant;
+	public JavaFeatureDefinition(IJavaCodeLensParticipant codeLensParticipant,
```
xorye: What do you think about using constructor overloading instead? So that we can create instances  `new JavaFeatureDefinition(codeLensParticipant)` instead of `new JavaFeatureDefinition(codeLensParticipant, null, null)`?

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/core/java/IJavaCodeLensParticipant.java
```diff
@@ -0,0 +1,77 @@
...
+
+	/**
+	 * Collect codeLens according to the context.
+	 * 
+	 * @param context the java codeLens context
+	 * @param monitor the progress monitor
+	 * 
+	 * @throws CoreException
+	 */
+	void collectCodeLens(JavaCodeLensContext context, IProgressMonitor monitor) throws CoreException;
```
xorye: I've noticed that `collectCodeLens` and `collectDiagnostics` has a `void` return type, whereas `collectHover` has a `Hover` return type.

Is there a reason why `collectHover` returns `Hover` instead of `void` so that they are all consistent?

PR #233, opened 2020-03-02
When pasting, the comments gets duplicated when

angelozerr:
When pasting, the comments gets duplicated when "editor.formatOnPaste":true

Fixes https://github.com/redhat-developer/vscode-quarkus/issues/220


xorye:
Works/looks great. Thank you for fixing this

PR #234, opened 2020-03-03
Provide URL CodeLens for MicroProfile RegisterRestClient

### LINKED ISSUE #167: Provide URL CodeLens for MicroProfile RegisterRestClient
Given this Java file:

```java
@Path("/v2")
@RegisterRestClient
public interface CountriesService {

    @GET
    @Path("/name/{name}")
    @Produces("application/json")
    Set<Country> getByName(@PathParam String name);
```

And this configuration in application.properties:

`org.acme.restclient.CountriesService/mp-rest/url=https://restcountries.eu/rest`

we should display an URL CodeLens for the JAX-RS method (client):

```java
[https://restcountries.eu/rest/v2/name/{name}]
Set<Country> getByName(@PathParam String name);
```

In this case we should avoid perhaps testing the availability of the server ?

angelozerr:
Provide URL CodeLens for MicroProfile RegisterRestClient

Fixes #167

### DIFF
--- a/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/restclient/java/MicroProfileRestClientCodeLensParticipant.java
+++ b/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/restclient/java/MicroProfileRestClientCodeLensParticipant.java
@@ -9,24 +9,24 @@
 * Contributors:
 *     Red Hat Inc. - initial API and implementation
 *******************************************************************************/
-package com.redhat.microprofile.jdt.internal.jaxrs.java;
+package com.redhat.microprofile.jdt.internal.restclient.java;
 
 import static com.redhat.microprofile.jdt.core.jaxrs.JaxRsUtils.createURLCodeLens;
 import static com.redhat.microprofile.jdt.core.jaxrs.JaxRsUtils.getJaxRsPathValue;
 import static com.redhat.microprofile.jdt.core.jaxrs.JaxRsUtils.isJaxRsRequestMethod;
+import static com.redhat.microprofile.jdt.core.utils.AnnotationUtils.getAnnotation;
+import static com.redhat.microprofile.jdt.core.utils.AnnotationUtils.getAnnotationMemberValue;
 import static com.redhat.microprofile.jdt.core.utils.JDTTypeUtils.overlaps;
-import static com.redhat.microprofile.jdt.internal.jaxrs.JaxRsConstants.JAVAX_WS_RS_PATH_ANNOTATION;
+import static com.redhat.microprofile.jdt.internal.restclient.MicroProfileRestClientConstants.REGISTER_REST_CLIENT_ANNOTATION;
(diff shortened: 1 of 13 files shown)


angelozerr:
To test this PR, open the `rest-client-quickstart` project (used for the tests) and follow the demo:

MPRestClientCodeLensDemo

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/restclient/java/MicroProfileRestClientCodeLensParticipant.java
```diff
@@ -0,0 +1,167 @@
...
+					String pathValue = getJaxRsPathValue(type);
+					collectURLCodeLenses(type.getChildren(), url, pathValue, mpProject, lenses, params, utils, monitor);
+				}
+				continue;
+			} else if (element.getElementType() == IJavaElement.METHOD) {
+				if (utils.isHiddenGeneratedElement(element)) {
+					continue;
+				}
+				// ignore element if method range overlaps the type range, happens for generated
+				// bytcode, i.e. with lombok
```
xorye: // bytecode,

xorye:
I am getting duplicate `/` if the url/uri or path ends in a `/`, although I'm not sure if this is a common occurrence?

For example, if I have
```
@Path("/v2/")
@RegisterRestClient(baseUri = "hello.com")
public interface CountriesService {

    @GET
    @Path("/name/{name}")
    @Produces("application/json") 
    Set<Country> getByName(@PathParam String name);

```
since I have a `/v2/` for the `@Path` annotation, the codelens I get is `hello.com/v2//name/{name}`

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/restclient/java/MicroProfileRestClientCodeLensParticipant.java
angelozerr: fixed

angelozerr:
> I am getting duplicate / if the url/uri or path ends in a /, although I'm not sure if this is a common occurrence?

I fixed the problem and write tests for that. There were the same problem for JAX-RS since now it's a JAX RS utilities.

xorye:
Great. I have one last comment, the codeLens are not meant to be clicked, right? In that case I noticed that if the codelens command is set to an empty string, the codelens is not clickable in VS Code. I wonder if that's the case for other clients?

angelozerr:
> Great. I have one last comment, the codeLens are not meant to be clicked, right? 

No it should be clickable (like URL CodeLens for JAX-RS), it doesn't work for you?

> In that case I noticed that if the codelens command is set to an empty string, the codelens is not clickable in VS Code. I wonder if that's the case for other clients?

The LSP specification is not very clear about that, but the only way to set a text for Codelens is to create a command and update the title of the command. When command has no id (empty string), the command is not executed on vscode. On Eclipse IDE I don't know since it is implementing (or perhaps finished now) but I think it's the right way.

angelozerr:
@xorye do you think we can merge this PR?

Issue #236, opened 2020-03-09
Support for `MicroProfile Health` CodeAction


angelozerr:
See https://github.com/MicroShed/microprofile-language-server/blob/517056559d82244355b51365d6617caf5b36be1f/src/main/java/com/microprofile/lsp/internal/helper/QuickFixHelper.java

angelozerr:
Fixed with https://github.com/redhat-developer/quarkus-ls/pull/253

PR #238, opened 2020-03-10
Consume LSP4J 0.9.0

### LINKED ISSUE #237: Consume LSP4J 0.9.0
Consume last version of LSP4J 0.9.0. It requires some code changes because LSP4J 0.9.0 breaks some API.

angelozerr:
Consume LSP4J 0.9.0

Fixes #237

### DIFF
--- a/microprofile.ls/com.redhat.microprofile.ls/pom.xml
+++ b/microprofile.ls/com.redhat.microprofile.ls/pom.xml
@@ -15,7 +15,7 @@
 		<maven.compiler.source>1.8</maven.compiler.source>
 		<maven.build.timestamp.format>yyyyMMdd-HHmm</maven.build.timestamp.format>
 		<dev.build.timestamp>${maven.build.timestamp}</dev.build.timestamp>
-		<lsp4j.version>0.8.1</lsp4j.version>
+		<lsp4j.version>0.9.0</lsp4j.version>
 		<jboss.releases.repo.id>jboss-releases-repository</jboss.releases.repo.id>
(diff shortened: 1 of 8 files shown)


jeffmaury:
Just wondering on the effect for the next release of JBoss Tools which will be based on 2020-03 ?

angelozerr:
> Just wondering on the effect for the next release of JBoss Tools which will be based on 2020-03 ?

Ok I remove the 0.9.0 constraints on JDT LS side. So now there is only MicroProfile LS which depends on  0.9.0, so I think it should work.

But for https://github.com/redhat-developer/quarkus-ls/issues/236 I need 0.9.0 on JDT LS side because JDT LS depends on 0.9.0 and I have a compilation problem with Diagnostic#getCode():

 * in 0.8.1 Diagnostic#getCode() returns just a String
 * in 0.9.0 Diagnostic#getCode() returns String or Number (with Either).

Perhaps I could use jav areflection to support 0.8.1 and 0.9.0 to get the code to support the both lsp4j version. @fbricon what do you think about that?

angelozerr:
@jeffmaury could you test my PR and check it works with Eclipse IDE / LSP4E which consumes lsp4j 0.8.1. Thanks for your feedback.

jeffmaury:
Just installed latest 2019-03 and they are using LSP4J 0.9.0 so we should be good for next release of JBoss Tools

angelozerr:
Great thanks @jeffmaury for your feedback. So I add this contstraint to the MANIFEST.MF of JDT plugins.

PR #244, opened 2020-03-11
MicroProfile metrics property support

### LINKED ISSUE #241: Support for `MicroProfile Metrics` properties
According to the MP Metrics specs https://github.com/eclipse/microprofile-metrics/blob/dc94b84ec90dd4a0bc983336e4273247a4e415cc/spec/src/main/asciidoc/architecture.adoc, we should provide the following properties :

 * `mp.metrics.tags` 
 * `mp.metrics.appName` 

We could check https://github.com/eclipse/microprofile-metrics/blob/734cc5cfffad05fa288cf4c35ec0a2421487c3e4/api/src/main/java/org/eclipse/microprofile/metrics/MetricID.java#L72 exists in the classpath to provide those properties.

xorye:
Fixes #241 

The two properties (`mp.metrics.tags` and `mp.metrics.appName`) were hardcoded.

I was unable to find official documentation about these two properties, so this is what I have for the documentation for now:

Documentation for `mp.metrics.tags`:
"List of tag values. Tag values set through `mp.metrics.tags` MUST escape equal symbols `=` and commas  `,` with a backslash `\`."

Documentation for `mp.metrics.appName`:
"The app name."

This is how the new properties look in VS Code:

image

image

There are two things that are missing: the phase and the extension name. @angelozerr would it be a good idea to hardcode these as well?


microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/metrics/properties/MicroProfileMetricsProvider.java
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Properties provider that provides MicroProfile Metrics properties from the
+ * 
+ * @author David Kwon
+ * 
+ * @see https://github.com/eclipse/microprofile-metrics/blob/dc94b84ec90dd4a0bc983336e4273247a4e415cc/spec/src/main/asciidoc/architecture.adoc
+ *
+ */
+public class MicroProfileMetricsProvider extends AbstractAnnotationTypeReferencePropertiesProvider {
```
angelozerr: I think we should simplify provider class which wishes to implement static properties. Here my suggestion:

 * create abstract class AbstractStaticPropertiesProvider like this (please change name of class, method if you don't like it and please comment each methods):

```java
package com.redhat.microprofile.jdt.core;

import org.eclipse.core.runtime.IProgressMonitor;
import org.eclipse.jdt.core.search.SearchMatch;
import org.eclipse.jdt.core.search.SearchPattern;

public abstract class AbstractStaticPropertiesProvider extends AbstractPropertiesProvider {

	@Override
	public final void beginSearch(SearchContext context, IProgressMonitor monitor) {
		if (isAdaptedFor(context, monitor)) {
			collectStaticProperties(context, monitor);
		}
	}

	protected abstract boolean isAdaptedFor(SearchContext context, IProgressMonitor monitor);

	protected abstract void collectStaticProperties(SearchContext context, IProgressMonitor monitor);

	@Override
	public void collectProperties(SearchMatch match, SearchContext context, IProgressMonitor monitor) {
		// Do nothing
	}

	@Override
	protected String[] getPatterns() {
		return null;
	}

	@Override
	protected SearchPattern createSearchPattern(String pattern) {
		return null;
	}

}

```

 * implements this class like this:

```java
package com.redhat.microprofile.jdt.core;

import org.eclipse.core.runtime.IProgressMonitor;
import org.eclipse.jdt.core.IJavaProject;

import com.redhat.microprofile.jdt.core.utils.JDTTypeUtils;

public class MicroProfileMetricsProvider extends AbstractStaticPropertiesProvider{

	@Override
	protected boolean isAdaptedFor(SearchContext context, IProgressMonitor monitor) {
		IJavaProject javaProject = context.getJavaProject();
		return (JDTTypeUtils.findType(javaProject, METRIC_ID) == null);
	}

	@Override
	protected void collectStaticProperties(SearchContext context, IProgressMonitor monitor) {
		IPropertiesCollector collector = context.getCollector();
		String docs = "List of tag values.\r\n"
				+ "Tag values set through `mp.metrics.tags` MUST escape equal symbols `=` and commas `,` with a backslash `\\`.";
		super.addItemMetadata(collector, GLOBAL_TAGS_VARIABLE,
				"java.util.Optional<java.lang.String>", docs, null, null, null, null, null, false);

		docs = "The app name.";
		super.addItemMetadata(collector, APPLICATION_NAME_VARIABLE,
				"java.util.Optional<java.lang.String>", docs, null, null, null, null, null, false);
	}

}

```

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/metrics/properties/MicroProfileMetricsProvider.java
```diff
@@ -0,0 +1,75 @@
...
+
+	private static final String MP_METRICS_ADDED = MicroProfileMetricsProvider.class
+			.getName() + "#MicroProfileMetrics";
+	
+	public static final String METRIC_ID = "org.eclipse.microprofile.metrics.MetricID";
```
angelozerr: Please move Metrics class and properties name in a MicroProfileMetricsConstants (like other MicroProfile*Constants)

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/metrics/properties/MicroProfileMetricsProvider.java
```diff
@@ -0,0 +1,75 @@
...
+	
+	@Override
+	public void collectProperties(SearchMatch match, SearchContext context, IProgressMonitor monitor) {
+		IJavaProject javaProject = context.getJavaProject();
+		if (JDTTypeUtils.findType(javaProject, METRIC_ID) == null) {
+			return;
+		}
+		if (context.get(MP_METRICS_ADDED) == null) {
+			IPropertiesCollector collector = context.getCollector();
+			String docs = "List of tag values.\r\n"
```
angelozerr: Use context.getDocumentFormat() to generate markdown (like `) or not..

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/metrics/properties/MicroProfileMetricsProvider.java
xorye: This looks great!

In my latest  push, for the `AbstractStaticPropertiesProvider#getPatterns` method, I decided to return an empty array rather than null, since returning null created an NPE in `AbstractPropertiesProvider#createSearchPattern` here:

```
...
String[] patterns = getPatterns();
    for (String pattern : patterns) {
    ...
```

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/metrics/properties/MicroProfileMetricsProvider.java
xorye: @angelozerr is this still needed? If yes, maybe something like
```
if (no markdown support) {
    docs = docs.replaceAll("`", "");
}
```
to remove the ` from the documentation?

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/metrics/properties/MicroProfileMetricsProvider.java
angelozerr: > This looks great!

Glad you like it.

>  I decided to return an empty array rather than null, since returning null created an NPE

I think it should be better to return null and fix the NPE in AbstractPropertiesProvider#createSearchPattern

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/metrics/MicroProfileMetricsConstants.java
```diff
@@ -0,0 +1,34 @@
...
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.jdt.core.search.SearchMatch;
+
+import com.redhat.microprofile.jdt.core.AbstractAnnotationTypeReferencePropertiesProvider;
+import com.redhat.microprofile.jdt.core.AbstractStaticPropertiesProvider;
+import com.redhat.microprofile.jdt.core.IPropertiesCollector;
+import com.redhat.microprofile.jdt.core.SearchContext;
+import com.redhat.microprofile.jdt.core.utils.JDTTypeUtils;
+
```
angelozerr: Add javadoc

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/metrics/properties/MicroProfileMetricsProvider.java
angelozerr: For the moment we consider that we return every time markdown, please forget my comment. We will improve that in an another issue.

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/metrics/properties/MicroProfileMetricsProvider.java
xorye: Thanks, fixed

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/core/AbstractStaticPropertiesProvider.java
```diff
@@ -0,0 +1,59 @@
...
+	protected abstract void collectStaticProperties(SearchContext context, IProgressMonitor monitor);
+
+	@Override
+	public void collectProperties(SearchMatch match, SearchContext context, IProgressMonitor monitor) {
+		// Do nothing
+	}
+
+	@Override
+	protected String[] getPatterns() {
+		return new String[0];
```
angelozerr: return null

xorye: sorry, fixed

PR #245, opened 2020-03-12
MicroProfile OpenTracing property support

### LINKED ISSUE #240: Support for `MicroProfile OpenTracing` properties
See properties at
https://github.com/eclipse/microprofile-opentracing/blob/master/spec/src/main/asciidoc/configuration.asciidoc

I think we should manage those properties by hard coding it like https://github.com/redhat-developer/quarkus-ls/blob/dddd2248474d18db1f72c048499adf32332da265/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/faulttolerance/properties/MicroProfileFaultToleranceProvider.java#L286 but by checking if there are MicroProfile OpenTracing in the classpath.

To do that we could check that `org.eclipse.microprofile.opentracing.Traced` annotation exists in the classpath. See example at https://github.com/redhat-developer/quarkus-ls/blob/dddd2248474d18db1f72c048499adf32332da265/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/jaxrs/java/JaxRsCodeLensParticipant.java#L67

xorye:
Fixes #240 

This PR adds support for two new properties: `mp.opentracing.server.skip-pattern` and `mp.opentracing.server.operation-name-provider`.

Here is what they look like in VS Code:

image

image


microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/opentracing/properties/MicroProfileOpenTracingProvider.java
```diff
@@ -0,0 +1,54 @@
...
+
+	@Override
+	protected void collectStaticProperties(SearchContext context, IProgressMonitor monitor) {
+		IPropertiesCollector collector = context.getCollector();
+		String docs = "Specifies a skip pattern to avoid tracing of selected REST endpoints.";
+		super.addItemMetadata(collector, SKIP_PATTERN, "java.util.Optional<java.lang.String>", docs, null, null,
+				null, null, null, false);
+
+		docs = "Specifies operation name provider for server spans. Possible values are `http-path` and `class-method`.";
+		super.addItemMetadata(collector, OPERATION_NAME_PROVIDER, "java.util.Optional<java.lang.String>", docs, null,
```
angelozerr: It seems `mp.opentracing.server.operation-name-provider` has 2 values:

 * class-method
 * http-path

See  https://github.com/eclipse/microprofile-opentracing/blob/master/spec/src/main/asciidoc/microprofile-opentracing.asciidoc#server-span-name

To support that you need to fill hints. See a sample of JSON hints https://github.com/redhat-developer/quarkus-ls/blob/f3cb154d719b6c846ce5d9126de611f6d52a3943/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/restclient/properties/MicroProfileRegisterRestClientProvider.java#L73

So the idea is to create a hint like https://github.com/redhat-developer/quarkus-ls/blob/f3cb154d719b6c846ce5d9126de611f6d52a3943/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/restclient/properties/MicroProfileRegisterRestClientProvider.java#L164 with the name `mp.opentracing.server.operation-name-provider` and fill it with 2 ValueHint

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/opentracing/properties/MicroProfileOpenTracingProvider.java
```diff
@@ -0,0 +1,54 @@
...
+	protected boolean isAdaptedFor(SearchContext context, IProgressMonitor monitor) {
+		IJavaProject javaProject = context.getJavaProject();
+		return (JDTTypeUtils.findType(javaProject, TRACED_ANNOTATION) != null);
+	}
+
+	@Override
+	protected void collectStaticProperties(SearchContext context, IProgressMonitor monitor) {
+		IPropertiesCollector collector = context.getCollector();
+		String docs = "Specifies a skip pattern to avoid tracing of selected REST endpoints.";
+		super.addItemMetadata(collector, SKIP_PATTERN, "java.util.Optional<java.lang.String>", docs, null, null,
```
angelozerr: It seems type is `java.util.regex.Pattern` see https://github.com/eclipse/microprofile-opentracing/blob/master/spec/src/main/asciidoc/microprofile-opentracing.asciidoc#disabling-server-side-tracing. 

An idea (if it's easy) is to manage validation on MicroProfile LS side to check pattern is valid. (a naive idea is try to create a java.util.regex.Pattern with the property value by catching it and if exception is thrown we could display the error.

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/opentracing/MicroProfileOpenTracingConstants.java
```diff
@@ -0,0 +1,24 @@
...
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.microprofile.jdt.internal.opentracing;
+
+/**
+ * MicroProfile Open Tracing constants
+ * @author David Kwon
+ *
```
angelozerr: Add

```java
@see https://github.com/eclipse/microprofile-opentracing
```

xorye:
I have made the changes.

For `mp.opentracing.server.skip-pattern`, for the time being, I have added validation using `catch`.
Here it is in action:
image
In this case, the value (`[`) is not a valid pattern because it is missing a closing bracket.

For `mp.opentracing.server.skip-pattern`, the new hints were added:
image

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/opentracing/properties/MicroProfileOpenTracingProvider.java
```diff
@@ -0,0 +1,66 @@
...
+		String docs = "Specifies a skip pattern to avoid tracing of selected REST endpoints.";
+		super.addItemMetadata(collector, SKIP_PATTERN, "java.util.regex.Pattern", docs, null, null,
+				null, null, null, false);
+
+		docs = "Specifies operation name provider for server spans. Possible values are `http-path` and `class-method`.";
+		super.addItemMetadata(collector, OPERATION_NAME_PROVIDER, "java.util.Optional<java.lang.String>", docs, null,
+				null, null, null, null, false);
+		
+		ItemHint itemHint = collector.getItemHint(OPERATION_NAME_PROVIDER);
+		itemHint.setSourceType(OPERATION_NAME_PROVIDER);;
```
xorye: Is it correct to set the source type to `OPERATION_NAME_PROVIDER` as well?

xorye:
My latest push adds the hint descriptions:
image

and also displays the exception error message when the regex pattern is invalid:
image

Also, for `mp.opentracing.server.operation-name-provider`, I have set `class-method` as the default value since the documentation says `class-method` will be used if operation name provider is not specified:
https://github.com/eclipse/microprofile-opentracing/blob/master/spec/src/main/asciidoc/microprofile-opentracing.asciidoc#server-span-name
image

At the moment, it seems strange to make `mp.opentracing.server.operation-name-provider` have a value type of `java.util.Optional<java.lang.String>`, since the values that it expects are `http-path` and `class-method`. Additionally, this is weird because if I put any other value, I get this message:
image

I've scoured https://github.com/eclipse/microprofile-opentracing to see if they are some sort of enum, but no luck. What about just setting the value type to `"\"http-path\" or \"class-method\""`?
image

angelozerr:
For me it looks very good. Could you add a test on MicroProfile LS side with pattern?

microprofile.jdt/com.redhat.microprofile.jdt.test/src/main/java/com/redhat/microprofile/jdt/core/opentracing/MicroProfileOpenTracingTest.java
```diff
@@ -0,0 +1,62 @@
...
+				p(null, "mp.opentracing.server.operation-name-provider", "java.util.Optional<java.lang.String>",
+						"Specifies operation name provider for server spans. Possible values are `http-path` and `class-method`.",
+						false, null, null, null, 0, null)
+
+		);
+
+		assertPropertiesDuplicate(infoFromClasspath);
+		
+		assertHints(infoFromClasspath, h("mp.opentracing.server.operation-name-provider", null, true, "mp.opentracing.server.operation-name-provider", //
+				vh("class-method", null, null), //
```
angelozerr: I think it misses description, no?

xorye: fixed

xorye:
In my latest push, I have added a new MicroProfileLS pattern test.

xorye:
fixing travis errors

angelozerr:
Great job @xorye !

PR #252, opened 2020-03-16
MicroProfile OpenAPI property support

### LINKED ISSUE #216: Support for `MicroProfile Open API` properties
MicroProfile Open API provides some properties that we should supported. See 

https://github.com/eclipse/microprofile-open-api/blob/master/spec/src/main/asciidoc/microprofile-openapi-spec.adoc#311-core-configurations

xorye:
Fixes #216 

The properties model parser: 
https://github.com/xorye/quarkus-ls/blob/a31744ae4aaf18eeeca537db87294e397c7384c3/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/parser/PropertiesParser.java

has been changed to support multi-line values:
image

Currently, `mp.openapi.schema{ }` must be updated so that there can be support for key values (the content that goes inside of `{ }`) containing `.` characters.


xorye:
Update:

The properties parser changes have been moved to another PR: https://github.com/redhat-developer/quarkus-ls/pull/254 

Additionally for the time being, `mp.openapi.schema.*` would need to be ignored from unknown properties validation. To do this in VS Code, the `mp.openapi.schema.*` entry can be added to the `quarkus.tools.validation.unknown.excluded` config array.

angelozerr:
> The properties parser changes have been moved to another PR: #254

Thanks!

> Additionally for the time being, mp.openapi.schema.* would need to be ignored from unknown properties validation. To do this in VS Code, the mp.openapi.schema.* entry can be added to the quarkus.tools.validation.unknown.excluded config array.

Please create a PR for that in vscode-quarkus to add it by default in https://github.com/redhat-developer/vscode-quarkus/blob/9228fa03c965870d4e9c296b293e71157952008a/package.json#L268

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/openapi/MicroProfileOpenAPIConstants.java
```diff
@@ -0,0 +1,28 @@
+package com.redhat.microprofile.jdt.internal.openapi;
```
angelozerr: add header

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/openapi/MicroProfileOpenAPIConstants.java
```diff
@@ -0,0 +1,28 @@
...
+
+public class MicroProfileOpenAPIConstants {
```
angelozerr: add javadoc

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/openapi/properties/MicroProfileOpenAPIProvider.java
```diff
@@ -0,0 +1,112 @@
...
+	protected boolean isAdaptedFor(SearchContext context, IProgressMonitor monitor) {
+		IJavaProject javaProject = context.getJavaProject();
+		return (JDTTypeUtils.findType(javaProject, OPEN_API_CONFIG) != null);
+	}
+
+	@Override
+	protected void collectStaticProperties(SearchContext context, IProgressMonitor monitor) {
+		IPropertiesCollector collector = context.getCollector();
+		String docs = "Configuration property to specify the fully qualified name of the OASModelReader implementation.";
+		super.addItemMetadata(collector, MODEL_READER, "java.util.Optional<java.lang.String>", docs, null, null,
```
angelozerr: To know the type of each openapi properties, I think we should see in smallrye implementation. For instance https://github.com/smallrye/smallrye-open-api/blob/51e103fd5c2dcdc83136f45798330903aecbb9de/implementation/src/main/java/io/smallrye/openapi/api/OpenApiConfigImpl.java#L74 you can see that model reader is a String and not an java.util.Optional<java.lang.String>. So I think the correct type is `java.lang.String`. It's the same thing for the other properties.

xorye: Sounds good. Just double checking, properties like `mp.openapi.scan.packages`: https://github.com/smallrye/smallrye-open-api/blob/51e103fd5c2dcdc83136f45798330903aecbb9de/implementation/src/main/java/io/smallrye/openapi/api/OpenApiConfigImpl.java#L107

have a `Set<String>` type. Should I go ahead with `java.lang.String` since we do not have validation support for `Set<String>`? Or maybe the type should be `java.util.Set<java.lang.String>` in case we do support that type in the future?

xorye: fixed

PR #254, opened 2020-03-17
Parse PropertyValue when spanning multiple lines

xorye:
This PR updates the properties model parser to support multi-line values.

Before this PR:
image
Before this PR, since multi-line values were not supported, the second and third lines were being parsed as property keys, causing the 'unknown property' diagnostics.

After this PR:
image

Example content used in the images:
```
# Configuration file
# key = value
quarkus.application.name= quarkus \
    application \
    name
```

### DIFF
--- a/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/Node.java
+++ b/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/Node.java
@@ -81,7 +81,7 @@ public abstract class Node {
 		if (start == -1 || end == -1) {
 			return null;
 		}
-		return getOwnerModel().getText(start, end);
+		return getOwnerModel().getText(start, end).replace("\\\n", "");
 	}
 
 	/**
--- a/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/Node.java
+++ b/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/Node.java
@@ -81,7 +81,21 @@ public abstract class Node {
 		if (start == -1 || end == -1) {
 			return null;
 		}
-		return getOwnerModel().getText(start, end).replace("\\\n", "");
+		return getOwnerModel().getText(start, end, false);
+	}
+
+	/**
+	 * Returns the text of the node
+	 * 
+	 * @param skipMultiLine determines whether or not new lines characters and backslashes
+	 * should be preserved for multi line text values
+	 * @return the text of the node
+	 */
+	public String getText(boolean skipMultiLine) {
+		if (start == -1 || end == -1) {
+			return null;
+		}
+		return getOwnerModel().getText(start, end, skipMultiLine);
 	}
 
 	/**
--- a/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/PropertiesModel.java
+++ b/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/PropertiesModel.java
@@ -164,8 +164,48 @@ public class PropertiesModel extends Node {
 		return model;
 	}
 
-	public String getText(int start, int end) {
-		return document.getText().substring(start, end);
+	/**
+	 * Returns the text from the <code>start</code> offset (inclusive) to the <code>end</code>
+	 * offset (exclusive).
+	 * 
+	 * @param start         the start offset
+	 * @param end           the end offset
+	 * @param skipMultiLine determines whether or not new lines characters and backslashes
+	 * should be preserved for multi line text values
+	 * @return the text from the <code>start</code> offset (inclusive) to the <code>end</code>
+	 * offset (exclusive).
+	 */
+	public String getText(int start, int end, boolean skipMultiLine) {
+		String text = document.getText();
+		if (!skipMultiLine) {
+			return text.substring(start, end);
+		}
+
+		StringBuilder sb = new StringBuilder();
+		int i = start;
+		boolean trimLeading = false;
+		while (i < end) {
+			char curr = text.charAt(i);
+			if (curr == '\\') {
+				if (i < end - 1 && text.charAt(i + 1) == '\n') {
+					i += 2;
+					trimLeading = true;
+					continue;
+				} else if (i < end - 2 && text.charAt(i + 1) == '\r' && text.charAt(i + 2) == '\n') {
+					i += 3;
+					trimLeading = true;
+					continue;
+				}
+			}
+
+			if (!trimLeading || !Character.isWhitespace(curr)) {
+				trimLeading = false;
+				sb.append(curr);
+			} 
+			
+			i++;
+		}
+		return sb.toString();
 	}
 
 	public int offsetAt(Position position) throws BadLocationException {
--- a/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/Property.java
+++ b/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/Property.java
@@ -98,10 +98,18 @@ public class Property extends Node {
 	 * Returns the property name without the profile of the property key and null
 	 * otherwise.
 	 * 
+	 * For multiline property names, this method returns the property name
+	 * with the backslashes and newlines removed.
+	 * 
 	 * <ul>
 	 * <li>'%dev.' will return null.</li>
 	 * <li>'%dev.key' will return 'key'.</li>
 	 * <li>'key' will return 'key'.</li>
+	 * <li>
+	 *   'key1.\
+	 *    key2.\
+	 *    key3' will return 'key1.key2.key3'
+	 * </li>
 	 * </ul>
 	 * 
 	 * @return the property name without the profile of the property key and null
@@ -119,10 +127,18 @@ public class Property extends Node {
 	 * Returns the property name with the profile of the property key and null
 	 * otherwise.
 	 * 
+	 * For multiline property names, this method returns the property name
+	 * with the profile, with backslashes and newlines removed.
+	 * 
 	 * <ul>
 	 * <li>'%dev.' will return '%dev.'.</li>
 	 * <li>'%dev.key' will return '%dev.key'.</li>
 	 * <li>'key' will return 'key'.</li>
+	 * <li>
+	 *   '%dev.\'
+	 *   'key1.\
+	 *    key2' will return '%dev.key1.key2'
+	 * </li>
 	 * </ul>
 	 * 
 	 * @return the property name with the profile of the property key and null
@@ -138,6 +154,9 @@ public class Property extends Node {
 
 	/**
 	 * Returns the property value and null otherwise.
+	 * 
+	 * For multiline property values, this method returns the property value
+	 * with backslashes and newlines removed.
 	 * 
 	 * @return the property value and null otherwise.
 	 */
--- a/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/parser/PropertiesHandler.java
+++ b/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/parser/PropertiesHandler.java
@@ -121,5 +121,4 @@ public interface PropertiesHandler {
 	void blankLine(ParseContext context);
 
 	void delimiterAssign(ParseContext context);
-
 }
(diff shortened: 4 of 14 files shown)


xorye:
My latest push brings support for multi-line property keys:
image

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/parser/PropertiesHandler.java
```diff
@@ -122,4 +122,22 @@
...
+
+	/**
+	 * Returns true if property value exists and end offset
+	 * is missing. Returns false otherwise.
+	 * 
+	 * @return true if property value exists and end offset
+	 * is missing. Returns false otherwise.
+	 */
+	boolean propertyValueEndMissing();
+
```
xorye: I decided to add these methods to `PropertiesHandler` in order to determine if the parser was done parsing the name/value.

If `propertyNameEndMissing()` or `propertyValueEndMissing()` returns true, this would mean that the name/value was not finished being parsed, and must continue parsing the name/value on the next line.

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/Node.java
```diff
@@ -81,7 +81,7 @@ public String getText() {
 		if (start == -1 || end == -1) {
 			return null;
 		}
-		return getOwnerModel().getText(start, end);
+		return getOwnerModel().getText(start, end).replace("\\\n", "");
```
angelozerr: Could you explain which usecase you wish to cover? Is it working with Windows OS line separator `\r\n` ?

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/parser/PropertiesHandler.java
angelozerr: Why have you done this choice? The idea of the handler is the same idea than SAX handler, just collect some events. 

It's strange to delegate the propertyNameEndMissing behavior to the handler implementation. 

IMHO, It should be cleaner to manage this case inside the PropertiesParser. But perhaps it's hard to do that?

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/Node.java
xorye: The use case for this was for multiline property keys.

For example, for this application.properties file
```
quarkus.\
application.name = name
```
the property key text would be: `quarkus.\\napplication.name`, which would create an unknown property error because of the `\\n`. This line removes the `\\n` so that the property can be recognized.

> Is it working with Windows OS line separator \r\n ?
Thank you for reminding me. I did not test with Windows.

angelozerr: Ok thank you for the clarification. getText is used for comments too with formatting, is it annoying to remove every time \n? It's just a question if you need to do that every time and just for some cases (like getPropertyName).

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/parser/PropertiesHandler.java
angelozerr: @xorye I think we should have 2 methods :

 * getText(boolean skipMultiLine)
 * getText() which call getText(false)

because getText() is used too for formatting. I suspect that you should have a trouble with https://github.com/redhat-developer/quarkus-ls/blob/master/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/services/MicroProfileFormatter.java#L155 when you will try to format property key in multi line (you will loose multi line, I think)

Separate into to methods will give the capability to use getText in different context 

 * for getPropertyName, \n must be escaped
 * for formatting it should not.

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/Node.java
angelozerr: @xorye I think we should have 2 methods :

 * getText(boolean skipMultiLine)
 * getText() which call getText(false)

because getText() is used too for formatting. I suspect that you should have a trouble with https://github.com/redhat-developer/quarkus-ls/blob/master/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/services/MicroProfileFormatter.java#L155 when you will try to format property key in multi line (you will loose multi line, I think)

Separate into to methods will give the capability to use getText in different context 

 * for getPropertyName, \n must be escaped
 * for formatting it should not.

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/parser/PropertiesHandler.java
xorye: My latest push fixes this by managing the cases inside PropertiesParser. I now use this enum to keep track of what needs to be parsed on the next line:
https://github.com/redhat-developer/quarkus-ls/blob/3e4c5b42ed8db0c97aa16211ef6ecca61ad9c691/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/parser/PropertiesParser.java#L45-L52

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/Node.java
xorye: @angelozerr I tried to do this for my latest commit. To skip multiline, I was decided to look for \\n and \\r\n.

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/PropertiesModel.java
```diff
@@ -165,7 +165,31 @@ public void error(ParseContext context, ErrorEvent errorEvent) throws ParseExcep
...
+	public String getText(int start, int end, boolean skipMultiLine) {
+		String text = document.getText();
+		if (!skipMultiLine) {
+			return text.substring(start, end);
+		}
+
+		StringBuilder sb = new StringBuilder();
+		int i = start;
+		while (i < end) {
+			if (text.charAt(i) == '\\') {
```
angelozerr: Store char for i to avoid calling twice charat of i

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/Property.java
```diff
@@ -129,11 +134,15 @@ public String getPropertyName() {
 	 *         otherwise.
 	 */
 	public String getPropertyNameWithProfile() {
+		return getPropertyNameWithProfile(false);
+	}
+
+	public String getPropertyNameWithProfile(boolean skipMultiLine) {
```
angelozerr: Do we need this method signature ?

xorye: Having this method signature helps because without it, we get this case:
image
Even though `%prod.quarkus.application.name` exists (as a multiline property key), the completion options provide it in this case when it shouldn't

If these new method signatures are OK, I will go ahead and change the javadoc if you wish

angelozerr: @xorye I understand that getPropertyNameWithProfile must return the property name by skipping multi line to retrieve it in hover, validation, definition, etc and I think we should have just one method getPropertyNameWithProfile () which skip multi line and that's it (you can keep tour code with getPropertyNameWithProfile(boolean skipMultiLine) like now but please set this method to private. The idea is to provide a clean API for user which wishes consumes the PropertiesModel.

angelozerr: > If these new method signatures are OK, I will go ahead and change the javadoc if you wish

If you can remove it (to avoid having new methods), it should be better and getPropertyNameWithProfile , getPropertyName etc must skip multi line. It should be very good if you could clarify that in the javadoc of each methods.

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/Node.java
```diff
@@ -81,7 +81,14 @@ public String getText() {
 		if (start == -1 || end == -1) {
 			return null;
 		}
-		return getOwnerModel().getText(start, end);
+		return getOwnerModel().getText(start, end, false);
+	}
+
+	public String getText(boolean skipMultiLine) {
```
angelozerr: add javadoc

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/PropertiesModel.java
```diff
@@ -164,8 +164,29 @@ public void error(ParseContext context, ErrorEvent errorEvent) throws ParseExcep
 		return model;
 	}
 
-	public String getText(int start, int end) {
-		return document.getText().substring(start, end);
+	public String getText(int start, int end, boolean skipMultiLine) {
```
angelozerr: add javadoc

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/Node.java
xorye: fixed

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/PropertiesModel.java
xorye: fixed

angelozerr:
@xorye your PR looks very good, there is some little limitation.

Given this usecase:

```
quarkus.\
application.name = quarkus \
    application \
    name
```

you will see that outline show a carriage return:

image

To fix that, you should replace 

https://github.com/redhat-developer/quarkus-ls/blob/13b9e0793b6cb9696187044815ffbff9151c0a81/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/services/MicroProfileSymbolsProvider.java#L119

with 

`return key.getPropertyNameWithProfile();`

Please write a test with that.

angelozerr:
@xorye another point. It seems that space after `\` change the behavior (if you have a space it's not a multi line).

You can check with that with this little code:

```
public static void main(String[] args) throws IOException {
		Properties properties = new Properties();
		properties.load(new StringReader("greeting.\\\r\n" + 
				"             message = hello\r\n" + 
				"greeting.name = quarkus"));
		System.err.println(properties.keySet());
		System.err.println(properties.getProperty("greeting.message"));
		System.err.println(properties.getProperty("greeting.name"));
	}
```

The result of this main is

```
[greeting.name, greeting.message]
hello
quarkus
```

I mean we have `greeting.message` because there is no space after `greeting.\`

But if you add a space after `greeting.\ ` 

```
public static void main(String[] args) throws IOException {
		Properties properties = new Properties();
		properties.load(new StringReader("greeting.\\ \r\n" + 
				"             message = hello\r\n" + 
				"greeting.name = quarkus"));
		System.err.println(properties.keySet());
		System.err.println(properties.getProperty("greeting.message"));
		System.err.println(properties.getProperty("greeting.name"));
	}
```

the result is:

```
[greeting.name, message, greeting. ]
null
quarkus
```

xorye:
>  It seems that space after \ change the behavior (if you have a space it's not a multi line).

@angelozerr thank you for bringing this up. I'm not sure if this is a feature of `java.util.Properties` or Quarkus, but in my experience, the multiline keys and values only work if there is no space after the `\` so it seems the above behavior is desired?

Secondly, I noticed that for this use case you mentioned:
```
quarkus.\
application.name = quarkus \
    application \
    name
```

image

the property value is "quarkus&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;application&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;name". When trying this out in a Quarkus project, the value is actually just be "quarkus application name", which means that the start of each line is having its whitespace trimmed. Looks like this is another thing I must fix for this PR.

I will make the changes + test for the document symbol carriage return problem very soon.

angelozerr:
Could you write tests with Space After \ please

xorye:
Yes, I have written some here:
https://github.com/redhat-developer/quarkus-ls/blob/d10885f65b5c321cbe99aa816b2d34f930d94522/microprofile.ls/com.redhat.microprofile.ls/src/test/java/com/redhat/microprofile/parser/PropertiesModelTest.java#L160-L198

xorye:
I also wrote a test for the whitespace trimming:
https://github.com/redhat-developer/quarkus-ls/blob/d10885f65b5c321cbe99aa816b2d34f930d94522/microprofile.ls/com.redhat.microprofile.ls/src/test/java/com/redhat/microprofile/parser/PropertiesModelTest.java#L112-L131

PR #267, opened 2020-03-23
Read static properties from json

xorye:
This PR is missing javadoc and it might not be very clean. I would just like some feedback if this is on the right track or not. Thank you.


angelozerr:
> I would just like some feedback if this is on the right track or not.

Yes it is in the right way, but we can simplify more the static provider implementation. The final goal is to have this implementation for managing properties and hints both:

```java
public class MicroProfileOpenTracingProvider extends AbstractStaticPropertiesProvider {

	@Override
	protected boolean isAdaptedFor(SearchContext context, IProgressMonitor monitor) {
		IJavaProject javaProject = context.getJavaProject();
		return (JDTTypeUtils.findType(javaProject, TRACED_ANNOTATION) != null);
	}
	
	@Override
	protected InputStream getInputStream() {
		return MicroProfileOpenTracingProvider.class.getResourceAsStream("metadata.json")
	}
}
```

I will try to explain how to do that in my review.

microprofile.jdt/com.redhat.microprofile.jdt.core/META-INF/MANIFEST.MF
```diff
@@ -38,3 +38,5 @@ Export-Package: com.redhat.microprofile.commons,
 Bundle-ClassPath: lib/snakeyaml-1.25.jar,
  .
 Eclipse-ExtensibleAPI: true
+Import-Package: com.google.gson;version="2.8.2",
```
angelozerr: please remove version

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/core/ItemMetadataReader.java
```diff
@@ -0,0 +1,22 @@
...
+import com.google.gson.stream.JsonReader;
+import com.redhat.microprofile.commons.metadata.ItemMetadata;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.nio.charset.StandardCharsets;
+
+public class ItemMetadataReader {
```
angelozerr: please remove this class, the load of JSON must be done in AbstractStaticPropertiesProvider. If we see that the JSON load must be done in an another code, we will move this load into an utilitilies class. The idea is to try to limit utilities class. More you will have utilities class, more it will be hard for contributor to understand which utility it can use.

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/metrics/properties/MicroProfileMetricsProvider.java
```diff
@@ -33,6 +37,8 @@
  */
 public class MicroProfileMetricsProvider extends AbstractStaticPropertiesProvider {
 
+	private static final Logger LOGGER = Logger.getLogger(MicroProfileMetricsProvider.class.getName());
```
angelozerr: remove this logger

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/metrics/properties/MicroProfileMetricsProvider.java
```diff
@@ -42,14 +48,17 @@ protected boolean isAdaptedFor(SearchContext context, IProgressMonitor monitor)
 	@Override
 	protected void collectStaticProperties(SearchContext context, IProgressMonitor monitor) {
```
angelozerr: replace this `collectStaticProperties` method with 

```java
@Override
protected InputStream getInputStream() {
	return MicroProfileMetricsProvider.class.getResourceAsStream("metadata.json")
}
```

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/openapi/properties/MicroProfileOpenAPIProvider.java
```diff
@@ -43,6 +38,8 @@
  */
 public class MicroProfileOpenAPIProvider extends AbstractStaticPropertiesProvider {
 
+	private static final Logger LOGGER = Logger.getLogger(MicroProfileOpenAPIProvider.class.getName());
```
angelozerr: remove this logger

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/openapi/properties/MicroProfileOpenAPIProvider.java
```diff
@@ -52,61 +49,17 @@ protected boolean isAdaptedFor(SearchContext context, IProgressMonitor monitor)
 	@Override
 	protected void collectStaticProperties(SearchContext context, IProgressMonitor monitor) {
```
angelozerr: replace this `collectStaticProperties` method with 

```java
@Override
protected InputStream getInputStream() {
	return MicroProfileOpenAPIProvider.class.getResourceAsStream("metadata.json")
}
```

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/opentracing/properties/MicroProfileOpenTracingProvider.java
```diff
@@ -35,6 +41,8 @@
  */
 public class MicroProfileOpenTracingProvider extends AbstractStaticPropertiesProvider {
 
+	private static final Logger LOGGER = Logger.getLogger(MicroProfileOpenTracingProvider.class.getName());
```
angelozerr: remove this logger

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/opentracing/properties/MicroProfileOpenTracingProvider.java
```diff
@@ -44,13 +52,17 @@ protected boolean isAdaptedFor(SearchContext context, IProgressMonitor monitor)
 	@Override
 	protected void collectStaticProperties(SearchContext context, IProgressMonitor monitor) {
```
angelozerr: replace this `collectStaticProperties` method with 

```java
@Override
protected InputStream getInputStream() {
	return MicroProfileOpenTracingProvider.class.getResourceAsStream("metadata.json")
}
```

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/resources/com/redhat/microprofile/jdt/internal/metrics/properties/metadata.json
```diff
@@ -0,0 +1,18 @@
+[
+	{
+		"type": "java.util.Optional<java.lang.String>",
+		"required": true,
```
angelozerr: are you sure ?

xorye:
@angelozerr thank you so much for the detailed feedback.

>extension with microprofile-opentracing. Doing that will give the capability to write snippets for application.properties for microprofile-opentracing and filter it if project have no dependency to microprofile-opentracing

For this point, do you mean `"extensionName": "microprofile-opentracing"`?

If yes, would it be "microprofile-opentracing-api" instead because of the jar name?
image

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/resources/com/redhat/microprofile/jdt/internal/metrics/properties/metadata.json
xorye: > prefer using java.lang.String instead of using java.util.Optionnal<java.lang.String> (if it is the case)

For `mp.metrics.tags` and `mp.metrics.appName`, I think it should stay as java.util.Optional<java.lang.String> because of this code:
https://github.com/eclipse/microprofile-metrics/blob/da7045a380be7175e87da8585f053fb4593723e3/api/src/main/java/org/eclipse/microprofile/metrics/MetricID.java#L123

and this code:
https://github.com/eclipse/microprofile-metrics/blob/da7045a380be7175e87da8585f053fb4593723e3/api/src/main/java/org/eclipse/microprofile/metrics/MetricID.java#L127

But for the required, it should be false, right? Since the JSON was from the server trace, I will investigate why required is being set to true.

angelozerr: The code is `config.getOptionalValue(GLOBAL_TAGS_VARIABLE, String.class);` the second parameter is a String, not the Optional, I suppose it's the call of config.getOptionalValue which transform the String to Optional<String>. So I think it's just a String.

> But for the required, it should be false, right? Since the JSON was from the server trace, I will investigate why required is being set to true.

You are right there is a bug. In PropertiesCollector we have
`property.setRequired(defaultValue == null);`

It should be:

`property.setRequired(defaultValue != null);`

Perhaps it should be fixed in an another PR, because I fear tests will not good. We need to regenereate all-quarkus-properties.json too I think. Please create an issue for that.

angelozerr:
> @angelozerr thank you so much for the detailed feedback.

You are welcome!

> For this point, do you mean "extensionName": "microprofile-opentracing"?

Yes!

> If yes, would it be "microprofile-opentracing-api" instead because of the jar name?

Extension name is a notion of Quarkus, it can be different from the JAR name. See https://github.com/redhat-developer/quarkus-ls/issues/133

If we use JAR name,my fear is that if  "microprofile-opentracing-api.jar" change like  "mp-opentracing-api", the snippet context will not work more on MicroProfile LS.

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/resources/com/redhat/microprofile/jdt/internal/metrics/properties/metadata.json
xorye: To confirm, the type should be `java.lang.String` and required should stay `true`?

angelozerr: > To confirm, the type should be java.lang.String 

yes

> and required should stay true

no, please change that for this usecase. For the automtaic update of required of defaultValue we will fix that in an another PR.

xorye: Fixed

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/core/AbstractStaticPropertiesProvider.java
```diff
@@ -39,7 +66,40 @@ public final void beginSearch(SearchContext context, IProgressMonitor monitor) {
 	 * @param context the building scope context
 	 * @param monitor the progress monitor
 	 */
-	protected abstract void collectStaticProperties(SearchContext context, IProgressMonitor monitor);
+	protected void collectStaticProperties(SearchContext context, IProgressMonitor monitor) {
+		if (metadata == null) {
+			try {
+				metadata = getMetadata();
+			} catch (IOException e) {
+				// TODO Auto-generated catch block
```
angelozerr: Add a LOGGER.log.....

microprofile.jdt/com.redhat.microprofile.jdt.core/static-properties/mp-opentracing-metadata.json
```diff
@@ -0,0 +1,34 @@
+{
+	"properties": [
+		{
+			"type": "java.util.Optional<java.util.regex.Pattern>",
```
angelozerr: are you sure it's java.util.Optional ?

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/core/AbstractStaticPropertiesProvider.java
```diff
@@ -39,7 +66,40 @@ public final void beginSearch(SearchContext context, IProgressMonitor monitor) {
...
+				e.printStackTrace();
+			}
+		}
+		if (metadata != null) {
+			context.getCollector().merge(metadata);
+		}
+	}
+
+	protected ConfigurationMetadata getMetadata() throws IOException {
```
angelozerr: Add javadoc

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/core/AbstractStaticPropertiesProvider.java
```diff
@@ -39,7 +66,40 @@ public final void beginSearch(SearchContext context, IProgressMonitor monitor) {
...
+		InputStream in = getInputStream();
+		Reader reader = new InputStreamReader(in, StandardCharsets.UTF_8.name());
+		return createGson().fromJson(reader, ConfigurationMetadata.class);
+	}
+
+	protected InputStream getInputStream() throws IOException {
```
angelozerr: Add javadoc

angelozerr:
Great job @xorye !

PR #272, opened 2020-03-26
Add context to snippets

rzgry:
Part of redhat-developer#265

Adds context `type` to be used when deciding if an annotation should be displayed

### DIFF
--- a/microprofile.ls/com.redhat.microprofile.ls/src/main/resources/com/redhat/microprofile/snippets/mp-metrics.json
+++ b/microprofile.ls/com.redhat.microprofile.ls/src/main/resources/com/redhat/microprofile/snippets/mp-metrics.json
@@ -10,7 +10,10 @@
       "\tdescription = \"${2:description}\"",
       ")"
     ],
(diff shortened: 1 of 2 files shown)


microprofile.ls/com.redhat.microprofile.ls/src/main/resources/com/redhat/microprofile/snippets/mp-metrics.json
```diff
@@ -10,7 +10,11 @@
       "\tdescription = \"${2:description}\"",
       ")"
     ],
-    "description": "An annotation that contains the metadata information when requesting a metric to be injected or produced. This annotation can be used on fields of type Meter, Timer, Counter, and Histogram. For Gauge, the @Metric annotation can only be used on producer methods/fields."
+    "description": "An annotation that contains the metadata information when requesting a metric to be injected or produced. This annotation can be used on fields of type Meter, Timer, Counter, and Histogram. For Gauge, the @Metric annotation can only be used on producer methods/fields.",
+    "context": {
+      "kind": "annotation",
```
angelozerr, line 15: Could you remove kind information. For the moment I will not support it, because it will require to call JDT LS extension each time to know if completion was triggered before a class, a method, etc. I will support type for the moment which I think it is more important.

angelozerr:
Thanks @rzgry !

PR #269, opened 2020-03-24
Support for `MicroProfile RestClient` CodeAction

angelozerr:
Support for `MicroProfile RestClient` CodeAction

Fixes #255

Signed-off-by: azerr <azerr@redhat.com>

PR #282, opened 2020-03-30
Hide openapi code action if it is not applicable

### LINKED ISSUE #280: Hide OpenAPI source action if it is not applicable
Currently the source action to generate OpenAPI annotations will always appear even if it is not applicable.  However, someone choosing this will see nothing happen, which is confusing.  Instead, we should not display this option if it is not applicable.

bensonnw:
Fixes #280 

Signed-off-by: bensonn@ca.ibm.com


angelozerr:
@bensonnw could you update your description commit (in the issue and in the commit) like https://github.com/redhat-developer/quarkus-ls/pull/269

I mean : 

 * Fixes #280
 * Signed-off-by: ....

angelozerr:
Thanks @bensonnw

PR #284, opened 2020-03-31
Allow exclude unknown property pattern * to match forward slashes

### LINKED ISSUE #237: Consume LSP4J 0.9.0

xorye:
Fixes https://github.com/redhat-developer/vscode-quarkus/issues/237 by allowing a pattern terminating with `*` to match everything, including forward slashes

### DIFF
--- a/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/utils/AntPathMatcher.java
+++ b/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/utils/AntPathMatcher.java
@@ -216,7 +216,7 @@ public class AntPathMatcher implements PathMatcher {
 	 * as far as the given base path goes is sufficient)
 	 * @return {@code true} if the supplied {@code path} matched, {@code false} if it didn't
 	 */
-	protected boolean doMatch(String pattern,  String path, boolean fullMatch,
+	protected boolean doMatch(String pattern, String path, boolean fullMatch,
 			 Map<String, String> uriTemplateVariables) {
 
 		if (path == null || path.startsWith(this.pathSeparator) != pattern.startsWith(this.pathSeparator)) {
@@ -228,7 +228,17 @@ public class AntPathMatcher implements PathMatcher {
 			return false;
 		}
 
-		String[] pathDirs = tokenizePath(path);
+		String[] pathDirs = null;
+		if (pattern.indexOf("/") > -1) {
+			// if pattern includes '/', this means that the
+			// path string should be tokenized with this.pathSeparator = "/"
+			pathDirs = tokenizePath(path);
+		} else {
+			// else, assume that '/' in the path string is not being
+			// treated differently than any other character.
+			// As a result, the path string is considered one single entity.
+			pathDirs = new String[] {path};
+		}
 		int pattIdxStart = 0;
 		int pattIdxEnd = pattDirs.length - 1;
 		int pathIdxStart = 0;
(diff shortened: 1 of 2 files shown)


xorye:
A different solution to this issue that does not change `AntPathMatcher.java` would be to provide:
`org.acme.restclient.CountriesService/mp-rest/url`
`org.acme.restclient.CountriesService/mp-rest/**`
`org.acme.restclient.CountriesService/**`

in the quick fix, instead of providing:
`org.acme.restclient.CountriesService/mp-rest/url`
`org.acme.restclient.*`
`org.acme.*`
`org.*`

angelozerr:
The original antpathmatcher is used to match path. In pour case WE want to match a given property which can contrains slash si your solution is fine. But I Wonder if WE could jeep the original antpathmatcher and do some escape before calling it. M'y fear is that we will introduce some bugs By modifing antpathmatcher. In other words is original antpathmatherbcan support match with escaped slash?

xorye:
It seems that the original ant path matcher cannot support escaped slash.

The `tokenizePath` method: https://github.com/redhat-developer/quarkus-ls/blob/e085c21c31bbc4711b665f510cd2478281599278/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/utils/AntPathMatcher.java#L429-L436

will still separate the path based on `/`.

For example, if we call `tokenizePath("a\\/b\\/c")`, the result is `["a\", "b\", "c"]` instead of `["a/b/c"]`

xorye:
A different solution that still involves changing ant path matcher that I came up with, is this:

Replace https://github.com/redhat-developer/quarkus-ls/blob/3597f0697e35f36aad8af8cdab023438860e7d8b/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/utils/AntPathMatcher.java#L231

with:
```
String[] pathDirs = null;
if (Arrays.asList(pattDirs).contains("**")) {
	pathDirs = tokenizePath(path);
} else {
	pathDirs = new String[]{path};
}
```

angelozerr:
@xorye the default AntPathMatcher constructor uses '/' as path separator. Are you testing to create the AntPathMatcher with `new AntPathMatcher("")` ?

xorye:
In my latest push, I have added more specific tests.

I have also added one more change to `AntPathMatcher.java`, in order for this test case to pass.

The new change:
https://github.com/redhat-developer/quarkus-ls/blob/a423e8f39bb73d87816194396715e5de596257ad/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/utils/AntPathMatcher.java#L231-L241

The test case for that new change
https://github.com/redhat-developer/quarkus-ls/blob/d369348a1d51393ddf889cf54ebfffdcd2f8c77e/microprofile.ls/com.redhat.microprofile.ls/src/test/java/com/redhat/microprofile/services/ApplicationPropertiesDiagnosticsTest.java#L221-L227

The application properties file for the test case:
```
com.mycompany.remoteServices.MyServiceClient/mp-rest/url/foo=url
com.mycompany.remoteServices.MyServiceClient/mp-rest/uri/bar=uri
com.mycompany.remoteServices.MyOtherClient/mp-rest/url/foo=url
com.mycompany.remoteServices.MyOtherClient/mp-rest/uri/bar=uri
com.mycompany.foo=bar
```
Please let me know if my comments/fix in `AntPathMatcher.java` are not good, I am more than happy to improve the comments/fix.

fbricon:
works as advertised. Thanks @xorye!

PR #292, opened 2020-04-06
Add support for java.math.BigDecimal and java.math.BigInteger types

### LINKED ISSUE #261: Support for java.math.BigDecimal values
It would be great if there were application.properties and application.yaml value support for the `java.math.BigDecimal` type and `java.util.Optional<java.math.BigDecimal>` type.

Currently, in a project with the SmallRye OpenTracing Quarkus extension installed:
```
<dependency>
    <groupId>io.quarkus</groupId>
    <artifactId>quarkus-smallrye-opentracing</artifactId>
</dependency>
```

the `quarkus.jaeger.sampler-param` property expects a `java.util.Optional<java.math.BigDecimal>` type.

Since the language server does not recognize this value, if I set:
```quarkus.jaeger.sampler-param=false```,
there is no diagnostic informing the user that `false` is invalid for the property.

### PR DESCRIPTION
**Pull request by xorye**
Fixes #261

To test this PR, please open the `microprofile-open-tracing` project and experiment with the `quarkus.jaeger.sampler-param` property in `application.properties`. This property expects the `java.math.BigDecimal` type.

`java.math.BigDecimal` can be represented in different ways, as explained here: https://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html#BigDecimal(java.lang.String)

image

image

image

image

This also works for `application.yaml` files.

As for the `java.math.BigInteger` type, I'm not aware of any property with this type, but I added support for it anyways. For the meantime, this can be tested by adding:
```
    @ConfigProperty(name = "greeting.BigInteger")
    BigInteger bigInteger;

```
in a resource class.

### DIFF
--- a/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/services/MicroProfileValidator.java
+++ b/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/services/MicroProfileValidator.java
@@ -9,6 +9,8 @@
 *******************************************************************************/
 package com.redhat.microprofile.services;
 
+import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -32,6 +34,7 @@ import com.redhat.microprofile.model.values.ValuesRulesManager;
 import com.redhat.microprofile.settings.MicroProfileValidationSettings;
 import com.redhat.microprofile.utils.MicroProfilePropertiesUtils;
 import com.redhat.microprofile.utils.PositionUtils;
+import com.redhat.microprofile.utils.StringUtils;
 
 /**
  * Quarkus validator to validate properties declared in application.properties.
@@ -207,7 +210,9 @@ class MicroProfileValidator {
 				|| (metadata.isBooleanType() && !isBooleanString(value))
 				|| (metadata.isDoubleType() && !isDoubleString(value))
 				|| (metadata.isLongType() && !isLongString(value))
-				|| (metadata.isShortType() && !isShortString(value))) {
+				|| (metadata.isShortType() && !isShortString(value))
+				|| (metadata.isBigDecimalType() && !isBigDecimalString(value))
+				|| (metadata.isBigIntegerType() && !isBigIntegerString(value))) {
 			return "Type mismatch: " + metadata.getType() + " expected";
 		}
 		return null;
@@ -218,6 +223,9 @@ class MicroProfileValidator {
 	}
 
 	private static boolean isIntegerString(String str) {
+		if (!StringUtils.hasText(str)) {
+			return false;
+		}
 		try {
 			Integer.parseInt(str);
 			return true;
@@ -227,6 +235,9 @@ class MicroProfileValidator {
 	}
 
 	private static boolean isFloatString(String str) {
+		if (!StringUtils.hasText(str)) {
+			return false;
+		}
 		try {
 			Float.parseFloat(str);
 			return true;
@@ -236,6 +247,9 @@ class MicroProfileValidator {
 	}
 
 	private static boolean isLongString(String str) {
+		if (!StringUtils.hasText(str)) {
+			return false;
+		}
 		try {
 			Long.parseLong(str);
 			return true;
@@ -245,6 +259,9 @@ class MicroProfileValidator {
 	}
 
 	private static boolean isDoubleString(String str) {
+		if (!StringUtils.hasText(str)) {
+			return false;
+		}
 		try {
 			Double.parseDouble(str);
 			return true;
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/services/MicroProfileValidator.java [hunk 1]
```diff
@@ -262,6 +266,24 @@ private static boolean isShortString(String str) {
 		}
 	}
 
+	private static boolean isBigDecimalString(String str) {
+		try {
```
fbricon, line 270: check if str is null or empty before. No need to throw an exception if it can be avoided (exception creation is expensive)

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/services/MicroProfileValidator.java
```diff
@@ -262,6 +266,24 @@ private static boolean isShortString(String str) {
...
+			new BigDecimal(str);
+			return true;
+		} catch (NumberFormatException e) {
+			return false;
+		}
+	}
+
+	private static boolean isBigIntegerString(String str) {
+		try {
```
fbricon, line 279: check if str is null or empty before. No need to throw an exception if it can be avoided (exception creation is expensive)

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/services/MicroProfileValidator.java
xorye: fixed

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/services/MicroProfileValidator.java
xorye: fixed

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/services/MicroProfileValidator.java
```diff
@@ -227,6 +232,7 @@ private static boolean isIntegerString(String str) {
 	}
 
 	private static boolean isFloatString(String str) {
+		if (str == null || str.isEmpty()) return false;
```
fbricon, line 235: please always use { } for if statements. See https://blog.codecentric.de/en/2014/02/curly-braces/

angelozerr: We should provide in `com.redhat.microprofile.utils.StringUtils.isEmpty(String s)` like https://github.com/apache/commons-lang/blob/e389ce1edcc909444944dcb4064050f4eab7b872/src/main/java/org/apache/commons/lang3/StringUtils.java#L3582

because there are a lot of code which does that in MicroProfileValidator (and perhaps in an another classes?).

xorye: @fbricon cool article, fixed 
@angelozerr maybe we use this method? https://github.com/redhat-developer/quarkus-ls/blob/bdf6c5fc10f3fe03ad3d708086551739abb1348d/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/utils/StringUtils.java#L49

angelozerr: This method return true if there are some spaces in the string. I'm not sure that it's the expected behavior.

fbricon: No it returns true if there are non-whitespace characters

fbricon: So yes, StringUtils.hasText(String) does the job

angelozerr: If you want to use hasText, let's go, but it will change the original behavior because this method returns true if there are only spaces (it's a kind of trim).

fbricon: no it doesn't

xorye: Updated to use `StringUtils.hasText(String) `

PR #296, opened 2020-04-07
MircoProfile LRA property support

### LINKED ISSUE #243: Support for `MicroProfile LRA` properties
Support the `mp.lra.propagation.active` property. See https://github.com/eclipse/microprofile-lra/blob/2d7b24b4bcb755eadb19c74dadd504cc41b0c094/spec/src/main/asciidoc/microprofile-lra-spec.adoc#322-configuration-parameters

This property must be hard coded an dprovided only if MP LRA is in the classpath (check for instance if https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java exists)

xorye:
Fixes #243 

For testing, this PR adds a new test project: `microprofile-lra` and adds support for a new property: `mp.lra.propagation.active`:
image
image

For the property type, I've decided to use `java.lang.String` instead of `boolean` because of this explanation from the spec: https://github.com/eclipse/microprofile-lra/blob/2d7b24b4bcb755eadb19c74dadd504cc41b0c094/spec/src/main/asciidoc/microprofile-lra-spec.adoc#322-configuration-parameters
image


microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/lra/MicroProfileLRAConstants.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * MicroProfile LRA constants
+ * 
+ * @author David Kwon
+ *
+ */
+public class MicroProfileLRAConstants{
+
+	public static final String LRA = "org.eclipse.microprofile.lra.annotation.ws.rs.LRA";
```
angelozerr, line 22: Rename to `LRA_ANNOTATION`

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/internal/lra/properties/MicroProfileLRAProvider.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.jdt.core.IJavaProject;
+
+import com.redhat.microprofile.jdt.core.AbstractStaticPropertiesProvider;
+import com.redhat.microprofile.jdt.core.MicroProfileCorePlugin;
+import com.redhat.microprofile.jdt.core.SearchContext;
+import com.redhat.microprofile.jdt.core.utils.JDTTypeUtils;
+
+public class MicroProfileLRAProvider extends AbstractStaticPropertiesProvider {
```
angelozerr, line 22: Add Javadoc

microprofile.jdt/com.redhat.microprofile.jdt.core/static-properties/mp-lra-metadata.json
```diff
@@ -0,0 +1,11 @@
+{
+	"properties": [
+		{
+			"type": "java.lang.String",
+			"extensionName": "microprofile-lra",
```
angelozerr, line 5: I think it should be `microprofile-lra-api` (I mean the artifact id, please use the proper artifact id). I think we have the same problem with other startic json files (we should end extension name with -api, but please check that).

I'm sorry it's me that I suggest you to use that, but I was wrong.

angelozerr:
@xorye travis fails because you need to push the `.project` file too.

angelozerr:
> For the property type, I've decided to use java.lang.String instead of boolean because of this explanation from the spec: https://github.com/eclipse/microprofile-lra/blob/2d7b24b4bcb755eadb19c74dadd504cc41b0c094/spec/src/main/asciidoc/microprofile-lra-spec.adoc#322-configuration-parameters

I think it should be better to use boolean, because the expected value is a boolean.

I was not aware about those MP converters https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/converters.asciidoc and If I understand correctly, we could write 

`quarkus.jni.enable = YES` 

which is correct according to the doc, but vscode-quarkus displays an error.

I suggets you that you create an issue with this Boolean converter to improve that (I think a simple fix is to accept Y, YES? ON for boolean  value on MicroProfile LS side. but we need to study more the real behavior of boolean Quarkus properties)

microprofile.jdt/com.redhat.microprofile.jdt.core/static-properties/mp-lra-metadata.json
xorye: Sounds good. Since `-api` is in the artifact name for lra, metrics, openapi and opentracing, my latest push adds the `-api` at the end of the extension names.

xorye:
My latest push changes the type to java.lang.String

PR #297, opened 2020-04-08
Filter for Java (server) snippet

### LINKED ISSUE #265: Filter for Java (server) snippet
Once https://github.com/redhat-developer/quarkus-ls/pull/251 will be merged, we need to improve Java snippets completion:

* `filter Java snippets` according to some conditions like:
  * if `project has some dependency ` (ex : Open API @Operation snippet should be shown only if the project have the openapi dependency)
  * `where completion has been triggered` (ex : Open API @Operation snippet should be shown only if completion is triggered before a class, etc)

To do that, we must delegate those 2 features on JDT LS side. We could have a `microprofile/java/completion/resolveCompletionContext` JDT LS command which:

 * resolve condition (ex have this dependency? is completion is triggered before class, etc)

The idea is when completion is opened and snippets requires condition and compute of context  , it should call one time the resolveCompletionContext which should resolve condition and compute of context.

angelozerr:
Filter for Java (server) snippet

Fixes #265


angelozerr:
To test this PR, open the `microprofile-metrics` project (used for test), open a Java file, type @ to open completion, you should see only metrics snippets:

image

Remove metrics dependency and replace with openapi dependency in the pom.xml 

In other words, replace:

```xml
 <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-smallrye-metrics</artifactId>
    </dependency>

```
with 

```xml
 <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-smallrye-openapi</artifactId>
    </dependency>
```

After refreshing, you should see only openapi snippets:

image

@rzgry I had to fix your openapi snippets because some type package was wrong. Please check my fix are OK.

angelozerr:
I need to clean my code and write tests before switching to a real PR.

microprofile.ls/com.redhat.microprofile.ls/src/main/resources/com/redhat/microprofile/snippets/mp-openapi.json
```diff
@@ -93,7 +93,7 @@
     ],
     "description": "Encapsulates input parameters.",
     "context": {
-      "type": "org.eclipse.microprofile.openapi.annotations.Parameters"
+      "type": "org.eclipse.microprofile.openapi.annotations.parameters.Parameters.Parameters"
```
rzgry, line 96: I think this should just be `org.eclipse.microprofile.openapi.annotations.parameters.Parameters`

xorye:
Works great on my machine.

It wasn't a part of this PR but just one small comment so far:
https://github.com/redhat-developer/quarkus-ls/blob/389c18411e849a6edf6b3727d60243f2bf8f030e/microprofile.ls/com.redhat.microprofile.ls/src/main/resources/com/redhat/microprofile/snippets/mp-openapi.json#L106

Please remove the extra space before the last `"` so that it becomes ` "\trequired = ${2:true},",`

microprofile.ls/com.redhat.microprofile.ls/src/main/resources/com/redhat/microprofile/snippets/mp-openapi.json
angelozerr: fixed

angelozerr:
> Works great on my machine.

Thanks for the feedback.

> Please remove the extra space before the last " so that it becomes "\trequired = ${2:true},",

fixed.

@xorye  could you please review my code (javadoc and the README please).

microprofile.ls/README.md
```diff
@@ -62,23 +62,39 @@ Java and properties completion snippets are managed by the MicroProfile LS (snip
...
+    "body": [
+      ...
+    ],
+    "context": {
+    	"type": "org.eclipse.microprofile.openapi.annotations.Operation"
+    }
+  }
+```
+
+means that the snippet is shown only if the project has the `org.eclipse.microprofile.openapi.annotations.Operation` Java Annotation in the classpath. In otherwords, only when the Java project have a dependency to the MicroProfile  Open API.
```
xorye, line 97: In other words, only when the Java project has a dependency on MicroProfile Open API.

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/ls/JavaTextDocumentService.java
```diff
@@ -80,7 +80,8 @@ public JavaTextDocumentService(MicroProfileLanguageServer microprofileLanguageSe
 		this.documents = new JavaTextDocuments(microprofileLanguageServer);
 	}
 
-	// ------------------------------ did* for Java file ------------------------------
+	// ------------------------------ did* for Java file
+	// ------------------------------
```
xorye, line 84: Please fix formatting

xorye:
Also, it looks like `SnippetContextProperties.java` and `SnippetContextPropertiesDeserializer` can be deleted?

microprofile.ls/README.md
angelozerr: fixed.

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/ls/JavaTextDocumentService.java
angelozerr: fixed

angelozerr:
> Also, it looks like SnippetContextProperties.java and SnippetContextPropertiesDeserializer.java can be deleted?

Good catch! Fixed.

xorye:
Awesome, I'm good with this being merged

angelozerr:
Thanks @xorye for your great feedback.

PR #308, opened 2020-04-21
Completion in non-quarkus/non-mp project causes errors

angelozerr:
Completion in non-quarkus/non-mp project causes errors

Fixes https://github.com/redhat-developer/vscode-quarkus/issues/247


fbricon:
Thanks @angelozerr

Issue #312, opened 2020-04-24
Filter properties snippet with properties instead of dependency


angelozerr:
Today properties snippets uses `context/dependency` to show the snippet according to the dependency classpath:

```
"Add datasource properties": {
		"prefix": "qds",
		"body": [
			"quarkus.datasource.url=${1|jdbc:mariadb://localhost:3306/mydb,jdbc:mysql://localhost:3306/test,jdbc:h2:mem:mydb,jdbc:postgresql://localhost/mydb,jdbc:sqlserver://localhost:1433;databaseName=mydb|}",
			"quarkus.datasource.driver=${2|org.mariadb.jdbc.Driver,org.mysql.jdbc.Driver,org.h2.Driver,org.postgresql.Driver,com.microsoft.sqlserver.jdbc.SQLServerDriver|}",
			"quarkus.datasource.username=${3:developer}",
			"quarkus.datasource.password=${4:developer}",
			"quarkus.hibernate-orm.database.generation=${5|update,drop-and-create,create,drop,none|}"
		],
		"description": "Configure Quarkus datasource",
		"context": {
			"dependency": "quarkus-agroal"
		}
	}
```

There are some limitation, because those properties could be linked to the dependency version.

To fix this limitation, a simple fix is to filter according to the properties to insert

```
...
		"context": {
			"properties": ["quarkus.datasource.url",
                                              "quarkus.hibernate-orm.database.generation"]
		}
	}
```

PR #317, opened 2020-05-21
Fixed datasource snippet

### LINKED ISSUE #316: Quarkus datasource snippet is invalid
Hi,

The current datasource snippet is invalid: https://github.com/redhat-developer/quarkus-ls/blob/master/microprofile.ls/com.redhat.microprofile.ls/src/main/resources/com/redhat/quarkus/snippets/quarkus-properties.json#L5 .

`quarkus.datasource.url` should be `quarkus.datasource.jdbc.url`. Note the `jdbc` part. Right now you are using a mix of the new config and the old deprecated config.

If I were you, I would move to something like the following for the snippet:
```
quarkus.datasource.db-kind=...
quarkus.datasource.username=...
quarkus.datasource.password=...
quarkus.datasource.jdbc.url=...
quarkus.datasource.jdbc.min-size=...
quarkus.datasource.jdbc.max-size=...
```
The first 3 are general and common to JDBC and reactive datasources. The last 3 are specific to JDBC. Having min size and max size is a good idea IMHO. Default values of 5 and 15 would probably be OK.

Also I think I would remove the Hibernate ORM property from the snippet as it has nothing to do here and will lead to an unknown property warning if the user doesn't have the Hibernate ORM extension. Better add a specific Hibernate ORM snippet.

datho7561:
Added the suggested fields to the datasource snippet. Moved the Hibernate ORM portion of the snippet to a separate snippet that only shows up if Hibernate ORM is used by the Quarkus project. Fixes #316.

### DIFF
--- a/microprofile.ls/com.redhat.microprofile.ls/src/main/resources/com/redhat/quarkus/snippets/quarkus-properties.json
+++ b/microprofile.ls/com.redhat.microprofile.ls/src/main/resources/com/redhat/quarkus/snippets/quarkus-properties.json
@@ -26,5 +26,15 @@
 		"context": {
 			"dependency": "quarkus-jaeger"
 		}
+	},
+	"Add Hibernate ORM datasource properties": {
+		"prefix": "qhormd",
+		"body": [
+			"quarkus.hibernate-orm.database.generation=${5|none,drop-and-create,create,drop,update|}"
+		],
+		"description": "Configure Quarkus Hibernate ORM datasource",
+		"context": {
+			"dependency": "quarkus-hibernate-orm"
+		}
 	}
 }
--- a/microprofile.ls/com.redhat.microprofile.ls/src/main/resources/com/redhat/quarkus/snippets/quarkus-properties.json
+++ b/microprofile.ls/com.redhat.microprofile.ls/src/main/resources/com/redhat/quarkus/snippets/quarkus-properties.json
@@ -1,4 +1,20 @@
 {
+	"Add datasource properties for Hibernate ORM": {
+		"prefix": "qds",
+		"body": [
+			"quarkus.datasource.db-kind=${1|mariadb,mysql,h2,postgresql,derby,mssql|}",
+			"quarkus.datasource.username=${2:developer}",
+			"quarkus.datasource.password=${3:developer}",
+			"quarkus.datasource.jdbc.url=${4|jdbc:mariadb://localhost:3306/mydb,jdbc:mysql://localhost:3306/test,jdbc:h2:mem:mydb,jdbc:postgresql://localhost:5432/mydb,jdbc:derby://localhost:1527/mydb,jdbc:sqlserver://localhost:1433;databaseName=mydb|}",
+			"quarkus.datasource.jdbc.min-size=${5:5}",
+			"quarkus.datasource.jdbc.max-size=${6:15}",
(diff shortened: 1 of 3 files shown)


microprofile.ls/com.redhat.microprofile.ls/src/main/resources/com/redhat/quarkus/snippets/quarkus-properties.json
```diff
@@ -25,5 +26,15 @@
 		"context": {
 			"dependency": "quarkus-jaeger"
 		}
+	},
+	"Add Hibernate ORM datasource properties": {
+		"prefix": "qhormd",
+		"body": [
+			"quarkus.hibernate-orm.database.generation=${5|none,drop-and-create,create,drop,update|}"
```
fbricon, line 33: I don't think having a 1 line snippet is worth it, when simply autocompleting quarkus.hibernate-orm.database.generation provides the same choices (I haven't checked but I assume it's the current behavior)

datho7561: I checked, and completing `quarkus.hibernate-orm.database.generation` doesn't give the list of options to cycle through like the snippet does. The available options are listed in the completion box though:
MicroprofileIssue316Demo1
I agree that a one line snippet doesn't make much sense, so I'll remove it.

xorye: Indeed, `quarkus.hiberbate-orm.database.generation` has no value completion options for ow because it is defined as a String, and not an enum. But yes, I agree with the fact that having a one line snippet might not be worth it.

Maybe we can have a hibernate-specific snippet which also contains the datasource properties: 
```
# datasource configuration
quarkus.datasource.db-kind = postgresql
quarkus.datasource.username = hibernate
quarkus.datasource.password = hibernate
quarkus.datasource.jdbc.url = jdbc:postgresql://localhost:5432/hibernate_db

# drop and create the database at startup (use `update` to only update the schema)
quarkus.hibernate-orm.database.generation=drop-and-create
```
which is from the Quarkus Hibernate ORM guide: https://quarkus.io/guides/hibernate-orm

This would be fine because it seems that Hibernate ORM has quarkus-agroal as a dependency: https://github.com/quarkusio/quarkus/blob/2d8867d9f1481b02a94a334da86a2068cd4733a7/extensions/hibernate-orm/runtime/pom.xml#L20-L23, meaning that the datasource properties won't be unknown properties.

Or alternatively, we can add the hibernate-specific snippet after https://github.com/redhat-developer/quarkus-ls/issues/312 is fixed

datho7561: Eventually using the method from #312 sounds like a good idea. For the mean time, I could add another snippet with the same prefix that appears higher on the list when quarkus-hibernate-orm is loaded.

datho7561: I removed the hibernate orm snippet entirely

fbricon:
@datho7561 thanks!

PR #318, opened 2020-05-21
Fix expected items length passed as null

xorye:


xorye:
Nevermind about this one. There may be cases where we want to check for the existence of *some* completion items, but don't want to specify all of them. In that case, setting `expectedCount` to null is correct, because setting it to null causes the length assertion to never happen:
https://github.com/redhat-developer/quarkus-ls/blob/27d554f92d8d545145410fc34cf3815f86a117bd/microprofile.ls/com.redhat.microprofile.ls/src/test/java/com/redhat/microprofile/services/MicroProfileAssert.java#L175-L177

PR #321, opened 2020-05-29
Fix build failure and other improvements

fbricon:
- moved quarkus snippet tests to quarkus.ls
- generated microprofile.ls test jar so test classes can be reused in
quarkus tests
- fixed failing tests when running on Java > 12
- added buildAll scripts for convenience

Signed-off-by: Fred Bricon <fbricon@gmail.com>


fbricon:
I slipped in some small changes to snippets assertions, since it's related to the original problem.

microprofile.ls/com.redhat.microprofile.ls/pom.xml
```diff
@@ -1,4 +1,4 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0"
+	<project xmlns="http://maven.apache.org/POM/4.0.0"
```
angelozerr, line 1: remove spaces

PR #327, opened 2020-06-08
Build uber JAR for Quarkus specific LS

### LINKED ISSUE #326: quarkus.ls.ext is not producing an uber JAR
The doc https://github.com/redhat-developer/quarkus-ls/blob/master/quarkus.ls.ext/README.md claims that an uber JAR is produced. I did not find one.
This is a blocker for JBoss Tools as it will prevent us to use the latest version since the split.

jeffmaury:
Fixes #326

Signed-off-by: Jeff MAURY <jmaury@redhat.com>

### DIFF
--- a/quarkus.ls.ext/com.redhat.quarkus.ls/pom.xml
+++ b/quarkus.ls.ext/com.redhat.quarkus.ls/pom.xml
@@ -121,37 +121,37 @@
 					<serverId>${jboss.releases.repo.id}</serverId>
 				</configuration>
 			</plugin>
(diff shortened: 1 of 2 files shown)


jeffmaury:
@angelozerr @fbricon I really need a feedback on this as the code freeze for JBoss Tools is close: either reject it or merge so that we know if we need to update our builds / integration

jeffmaury:
Closing after discussion with @angelozerr : JBoss Tools will not use 2 different JARs like VSCode

PR #329, opened 2020-06-22
Rename quarkus settings/commands

### LINKED ISSUE #325: Rename `quarkus` prefix settings (used by language server) with `microprofile`
Today settings uses `quarkus` as prefix for settings used by language server. It should be renamed with `microprofile`. See 
https://github.com/rzgry/prototype-vscode-microprofile/blob/73fc02276981257b936888b670a0ad8529c2bd93/package.json#L77

The question is about vscode-quarkus, should we remove the settings used by language server?

rzgry:
Fixes #325 

Marking this PR are as a draft for now since these settings will need to be either be renamed in vscode-quarkus or removed once there is a published vscode-microprofile that starts the LS.


fbricon:
@rzgry please re-open that PR against https://github.com/eclipse/lsp4mp. Should be pretty straightforward

PR #330, opened 2020-06-29
Send project name to the client along with labels.

datho7561:
This is useful when presenting the different projects in a menu,
such as in https://github.com/redhat-developer/vscode-quarkus/issues/215.

### DIFF
--- a/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/core/ProjectLabelManager.java
+++ b/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/core/ProjectLabelManager.java
@@ -73,7 +73,7 @@ public class ProjectLabelManager {
 	private ProjectLabelInfoEntry getProjectLabelInfo(IProject project, List<String> types) {
 		String uri = JDTMicroProfileUtils.getProjectURI(project);
 		if (uri != null) {
-			return new ProjectLabelInfoEntry(uri, getProjectLabels(project, types));
+			return new ProjectLabelInfoEntry(uri, project.getName(), getProjectLabels(project, types));
 		}
 		return null;
 	}
--- a/microprofile.jdt/com.redhat.microprofile.jdt.test/src/main/java/com/redhat/microprofile/jdt/core/ProjectLabelTest.java
+++ b/microprofile.jdt/com.redhat.microprofile.jdt.test/src/main/java/com/redhat/microprofile/jdt/core/ProjectLabelTest.java
@@ -78,7 +78,27 @@ public class ProjectLabelTest {
 		assertLabels(projectLabelEntries, gradle, "gradle");
 	}
 
-	private void assertProjectLabelInfoContainsProject(List<ProjectLabelInfoEntry> projectLabelEntries,
+	@Test
+	public void projectNameMaven() throws Exception {
+		IJavaProject quarkusMaven = BasePropertiesManagerTest.loadMavenProject(MavenProjectName.using_vertx);
+		IJavaProject maven = BasePropertiesManagerTest.loadMavenProject(MavenProjectName.empty_maven_project);
+		List<ProjectLabelInfoEntry> projectLabelEntries = ProjectLabelManager.getInstance().getProjectLabelInfo();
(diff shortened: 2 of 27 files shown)


microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/core/ProjectLabelManager.java
```diff
@@ -73,7 +73,7 @@ private ProjectLabelManager() {
 	private ProjectLabelInfoEntry getProjectLabelInfo(IProject project, List<String> types) {
 		String uri = JDTMicroProfileUtils.getProjectURI(project);
 		if (uri != null) {
-			return new ProjectLabelInfoEntry(uri, getProjectLabels(project, types));
+			return new ProjectLabelInfoEntry(uri, project.getName(), getProjectLabels(project, types));
```
xorye, line 76: `project.getName()` returns the last segment of the project folder path:
https://help.eclipse.org/2020-06/nftopic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/core/resources/IResource.html#getName()
which is the folder name.

I think we want `project.getDescription().getName()` instead.

datho7561: Thank you!

datho7561:
ProjectNameAddingExtensions
Still need to fix the tests, although this use case now works successfully

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/core/ProjectLabelManager.java
angelozerr: In Eclipse IDE context, it uses the project.getName() to display the project and not the project name coming from the .project:

image

I think we should follow this idea. @fbricon what do you think about that?

microprofile.jdt/com.redhat.microprofile.jdt.test/src/main/java/com/redhat/microprofile/jdt/core/ProjectLabelTest.java
```diff
@@ -100,6 +120,19 @@ private void assertLabels(List<ProjectLabelInfoEntry> projectLabelEntries, IJava
 		}
 	}
 
+	private void assertName(List<ProjectLabelInfoEntry> projectLabelEntries, IJavaProject javaProject,
```
angelozerr, line 123: add static to show this method have no context of this class fields.

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/core/ProjectLabelManager.java
```diff
@@ -72,8 +72,14 @@ private ProjectLabelManager() {
 	 */
 	private ProjectLabelInfoEntry getProjectLabelInfo(IProject project, List<String> types) {
 		String uri = JDTMicroProfileUtils.getProjectURI(project);
+		String name;
 		if (uri != null) {
-			return new ProjectLabelInfoEntry(uri, getProjectLabels(project, types));
+			try {
```
angelozerr, line 77: replace with 
`return new ProjectLabelInfoEntry(uri, project.getName(), getProjectLabels(project, types));`

angelozerr:
@datho7561 please fix https://travis-ci.org/github/redhat-developer/quarkus-ls/builds/703686845#L1177

It misses .project and .classpath for your new test project (because of .gitignore).

microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/core/ProjectLabelManager.java
datho7561: As of `1a60ddb` I am doing:

```java
project.getName()
```

However, if all this does is get the folder name, then this PR just adds a bunch of overhead to the process, since `vscode-quarkus` was already just using the folder name.

PR #335, opened 2020-07-02
Filter properties snippet with properties instead of dependency

### LINKED ISSUE #312: Filter properties snippet with properties instead of dependency

angelozerr:
Filter properties snippet with properties instead of dependency

Fixes #312

### DIFF
--- a/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/services/MicroProfileCompletions.java
+++ b/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/services/MicroProfileCompletions.java
@@ -232,14 +232,15 @@ class MicroProfileCompletions {
 
 	/**
 	 * Collect Quarkus profiles
+	 * 
 	 * @param offset            the offset where completion was invoked
 	 * @param key               the property key
 	 * @param model             the properties model
 	 * @param markdownSupported boolean determining if markdown is supported
 	 * @param list              the completion list
 	 */
-	private static void collectProfileSuggestions(int offset, PropertyKey key,
-			PropertiesModel model, boolean markdownSupported, CompletionList list) {
+	private static void collectProfileSuggestions(int offset, PropertyKey key, PropertiesModel model,
+			boolean markdownSupported, CompletionList list) {
 
 		Range range = null;
 		Position currPosition = null;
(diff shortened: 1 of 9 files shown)


microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/snippets/SnippetContextForProperties.java
```diff
@@ -37,31 +41,37 @@
...
 	}
 
 	@Override
-	public boolean isMatch(Set<String> dependencies) {
-		return dependencies != null && dependencies.contains(dependency);
+	public boolean isMatch(MicroProfileProjectInfo projectInfo) {
+		for (String propertyName : this.properties) {
+			ItemMetadata metadata = MicroProfilePropertiesUtils.getProperty(propertyName, projectInfo);
+			if (metadata != null) {
+				return true;
```
xorye, line 71: According to this documentation from this PR:
>* for `properties files`: the context/properties can be declared in the snippet to show the snippet only if the declared properties (you can use simple String or String array for that) belong to the project.


If `properties` is an array of String, I think we should if each property from the array exists in the project? In this `isMatch` method, it is only checking if at least one property exists. If this was intended, then I think the documentation should be updated to:

```
* for `properties files`: the context/properties (can be String or String array) can be declared by the snippet to show the
snippet only if the declared property belongs to the project. If an array of properties were provided, the snippet will be 
displayed if at least one property belongs to the project.
```

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/services/MicroProfileCompletions.java
```diff
@@ -255,19 +256,18 @@ private static void collectProfileSuggestions(int offset, PropertyKey key,
...
+			Property property = (Property) n;
+			return property.getProfile();
+		}).filter(Objects::nonNull).filter(not(String::isEmpty)).distinct().collect(Collectors.toSet());
 		// merge existings profiles with default profiles.
 		profiles.addAll(QuarkusModel.getDefaultProfileNames());
 		// Completion on profiles
 		for (String p : profiles) {
-			if (p.equals(key.getProfile())) continue;
+			if (p.equals(key.getProfile()))
+				continue;
```
xorye, line 270: Could we add braces here plesae

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/snippets/SnippetContextForProperties.java
angelozerr: Indeed it was not clear. I change my code to match all properties instead of he least one property strategy. I changed README too explain that.

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/services/MicroProfileCompletions.java
angelozerr: fixed

PR #339, opened 2020-07-04
Run LSP4J callback methods in separate thread

### LINKED ISSUE #338: Deadlock when client is sending burst of request
In IJ Quarkus, when the IDE starts, if several Java files are opened, we will send a didOpen request for each file. The server will send back a getProjectLabels request for each file and this causes a deadlock. The reason is that each LSP request method should not run in the calling thread because it prevents the server to read the input stream of LSP4J is single thread per connection.

jeffmaury:
This is a POC to fix #338 

It is composed on purpose of 2 commits:
- the first one is a unit test that demonstrate the issue and should fail if you cherry-pick only this commit
- the second one is an example on how to update the LSP4J callback methods to make sure they run on a separate thread.

This PR is drafted because I think there are other part of the LS that needs to be updated as well

Fixes #338

### DIFF
--- a/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/ls/MicroProfileTextDocumentService.java
+++ b/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/ls/MicroProfileTextDocumentService.java
@@ -65,6 +65,8 @@ public class MicroProfileTextDocumentService implements TextDocumentService {
 	private final JavaTextDocumentService javaTextDocumentService;
 	private SharedSettings sharedSettings;
 
+	private static final CompletableFuture COMPLETED = CompletableFuture.completedFuture(null);
+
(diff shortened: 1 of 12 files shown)


angelozerr:
Thanks @jeffmaury for your PR. I wonder if this problem is because of IJ LSP client because we have not found some problem in vscode (and for other language server like LemMinx we do like this).

I think it's more a problem with LSP4J and language server should avoid implementing some extra code to avoid having some block problem.

@fbricon what do you think?

jeffmaury:
I don't think it's related to the IJ LSP implementation we have in Quarkus Tools as my sample test does almost nothing in the LSP4J callback method.
I think you did not face the problem in VSCode because you are in the NodeJS world so writing data is performed through an asynchronous call. The problem with LSP4J is that it exposes asynchronous interface in the callback methods (through the CompletableFuture) but when data is returned, write is performed synchronously so the deadlock

angelozerr:
@jeffmaury what about this PR? Have you fixed your problem on IJ side?

jeffmaury:
Yes I did not have it again

PR #340, opened 2020-07-07
Property Expressions

datho7561:
In quite a few implementations of MicroProfile (Thorntail, Quarkus, Helidon), the values of properties can be assigned the value of another property using the syntax `first.property=${other.property}`. This PR ultimately aims to add completion for existing property names when `${` is typed. This PR builds out a model for property expressions to allow the language server to more easily work with them.

- [x] Build out the model for MicroProfile properties files to incorporate property expressions
- [x] Adjust the parser so that it adds the property expressions to the model
- [x] Adjust parser tests to gain confidence the parser is parsing correctly
- [ ] Build an algorithm to detect if there is currently circular dependency
- [ ] Build an algorithm to detect if adding a dependency will create a circular dependency
- [ ] Add completion for `${`
- [ ] Test completion for `${`


xorye:
Will this handle this case?
```
propertyA = 8080
propertyB = test:${propertyA}
```

Have you considered resolving property expressions after parsing application.properties rather than resolving them while parsing which I think this PR is doing?

datho7561:
> Will this handle this case?
> 
> ```
> propertyA = 8080
> propertyB = test:${propertyA}
> ```

The model should be able to support representing that. Once I fix the parser, it should parse that into the model. I haven't added resolving the references yet, or verifying if there is a circular dependency.
 
> Have you considered resolving property expressions after parsing application.properties rather than resolving them while parsing which I think this PR is doing?

I think resolving them after parsing is the way to go. I haven't implemented resolution yet.

My current plan is something like this:

* Parse the references into the model
* When the `getResolvedValue()` is called, build a graph of the interdependencies and use DFS/BFS to check if there any circular relations before trying to sub in the value of the references.
* When calling completion on `${`, try adding each property as a dependency in the graph, and see if it introduces a cyclic dependency. If it doesn't, add it to the suggestion list

This is a very rough plan, and it has lots of flaws. Please let me know if you have any other ideas on how to do this.

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/Node.java
```diff
@@ -34,7 +34,7 @@
 	 *
 	 */
 	public static enum NodeType {
-		DOCUMENT, PROPERTY, PROPERTY_KEY, PROPERTY_VALUE, COMMENTS, ASSIGN;
+		DOCUMENT, PROPERTY, PROPERTY_KEY, PROPERTY_VALUE, PROPERTY_LITERAL, PROPERTY_EXPRESSION, COMMENTS, ASSIGN;
```
angelozerr, line 37: I will rename to `PROPERTY_VALUE_LITERAL` and  `PROPERTY_VALUE_EXPRESSION`

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/PropertyExpression.java
```diff
@@ -0,0 +1,39 @@
+package com.redhat.microprofile.model;
```
angelozerr, line 1: Add header

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/PropertyExpression.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class PropertyExpression extends Node implements ResolvableText {
```
angelozerr, line 3: Add Javadoc

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/ResolvableText.java
```diff
@@ -0,0 +1,25 @@
...
+ * Represents text that has references to other text in it.
+ */
+public interface ResolvableText {
+	
+	/**
+	 * Returns the text such that the references to other text are not replaced with the referenced text.
+	 * 
+	 * @return the text such that the references to other text are not replaced with the referenced text.
+	 */
+	public String getLiteralValue();
```
angelozerr, line 13: remove `public` modifier.

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/parser/PropertiesHandler.java
```diff
@@ -93,12 +93,40 @@
...
+
+	/**
+	 * End of a text literal section in the property value
+	 * 
+	 * @param context the {@link ParseContext}
+	 */
+	void endPropertyValueLiteral(ParseContext context);
+
+	/**
+	 * Start of a property expression
```
angelozerr, line 117: Add a Javadoc comment which explains what is a property expression

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/parser/PropertiesParser.java
```diff
@@ -48,7 +48,9 @@
 	private enum ParseState {
 		Property,
 		PropertyName,
-		PropertyValue
+		PropertyValue,
+		PropertyExpression,
```
angelozerr, line 52: Rename to PropertyValueExpression

microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/model/ResolvableText.java
datho7561: I removed the method entirely

datho7561:
I am migrating this PR to [lsp4mp(https://github.com/eclipse/lsp4mp)

PR #343, opened 2020-07-08
 Change tests so that folder name and project name are distinct

xorye:
Currently, the tests introduced in https://github.com/redhat-developer/quarkus-ls/pull/330 succeed when the project labels command retrieves the project name as both `project.getName()` or `project.getDescription().getName()` here: https://github.com/xorye/quarkus-ls/blob/4a1dae985db0ef0cfa9f7f8bac82e54bb6b878fd/microprofile.jdt/com.redhat.microprofile.jdt.core/src/main/java/com/redhat/microprofile/jdt/core/ProjectLabelManager.java#L83

This PR ensures that tests pass if `project.getDescription().getName()` is used, and fail if `project.getName()` is used.


xorye:
One moment please, I will explain why this PR removes some tests.

microprofile.jdt/com.redhat.microprofile.jdt.test/src/main/java/com/redhat/microprofile/jdt/core/BasePropertiesManagerTest.java
```diff
@@ -187,7 +181,7 @@ public void run(IProgressMonitor monitor) throws CoreException {
 		// property:
 		// deployment-artifact=io.quarkus\:quarkus-hibernate-orm-deployment\:0.21.1
 
-		IJavaProject javaProject = JavaModelManager.getJavaModelManager().getJavaModel().getJavaProject(description.getName());
+		IJavaProject javaProject = JavaModelManager.getJavaModelManager().getJavaModel().getJavaProject(projectName);
```
xorye, line 184: After a discussion with @fbricon, we can stick with Maven tests for now. When loading Maven projects (ie, from VS Code), the IResource path is `P/folderName`, therefore we should use `projectName` (which is the folder name) here instead of `description.getName()`.

This essentially reverts a change that https://github.com/redhat-developer/quarkus-ls/pull/330 did.

microprofile.jdt/com.redhat.microprofile.jdt.test/src/main/java/com/redhat/microprofile/jdt/core/ProjectLabelTest.java
```diff
@@ -82,31 +82,25 @@ public void getProjectLabelMultipleProjects() throws Exception {
...
-		IJavaProject folderNameDifferent = BasePropertiesManagerTest.loadMavenProject(MavenProjectName.folder_name_different_maven);
 		List<ProjectLabelInfoEntry> projectLabelEntries = ProjectLabelManager.getInstance().getProjectLabelInfo();
 		assertName(projectLabelEntries, quarkusMaven, "using-vertx");
 		assertName(projectLabelEntries, maven, "empty-maven-project");
-		assertName(projectLabelEntries, folderNameDifferent, "mostly.empty");
 	}
-
+	
 	@Test
-	public void projectNameSameFolderName() throws Exception {
```
xorye, line 93: This test tested the case where you have two folders in your workspace with the same folder name, but different project names specified in `.project`.

I removed this case because for maven projects, since this the IResource path is their folder name, if two maven projects have the same folder name, they would share the same IProject object.

In other words, this test would try to load two projects, but only one project would be created.

Also, it seems that in Eclipse, by default, you cannot have two maven project folders with the same name in the package explorer even though the two projects have different names in `.project`.

microprofile.jdt/com.redhat.microprofile.jdt.test/src/main/java/com/redhat/microprofile/jdt/core/ProjectLabelTest.java
```diff
@@ -82,31 +82,25 @@ public void getProjectLabelMultipleProjects() throws Exception {
...
 
 	@Test
 	public void projectNameGradle() throws Exception {
 		IJavaProject quarkusGradle = BasePropertiesManagerTest.loadGradleProject(GradleProjectName.quarkus_gradle_project);
 		IJavaProject gradle = BasePropertiesManagerTest.loadGradleProject(GradleProjectName.empty_gradle_project);
-		IJavaProject renamedGradle = BasePropertiesManagerTest.loadGradleProject(GradleProjectName.renamed_quarkus_gradle_project);
 		List<ProjectLabelInfoEntry> projectLabelEntries = ProjectLabelManager.getInstance().getProjectLabelInfo();
 		assertName(projectLabelEntries, quarkusGradle, "quarkus-gradle-project");
 		assertName(projectLabelEntries, gradle, "empty-gradle-project");
-		assertName(projectLabelEntries, renamedGradle, "my-gradle-project");
```
xorye, line 109: When loading Gradle projects (ie, from VS Code), if the folder name is `folder-name` and the project name in .project is `new-name`, unlike loading Maven projects, I've observed that the IResource path is `P/new-name` instead of `P/folder-name`.

Since Maven projects are being focussed on, for now, I removed the Gradle test case where folder name != project name.

xorye: This removed test case will cause a NPE since thid PR loads projects using folder name: `JavaModelManager.getJavaModelManager().getJavaModel().getJavaProject({folder name here});` in BasePropertiesManagerTest.java.

Issue #352, opened 2020-07-15
Adapt README with LSP4MP


angelozerr:
Adapt README with LSP4MP

fbricon:
Make sure all links to lsp4mp are valid, remove all mentions of this repo containing microprofile stuff. Emphasize this is an extension to lsp4mp

PR #355, opened 2020-07-16
Fix links in README files

jmini:


jmini:
Similar change was made with 23e34f9042c1f8b5c0e3911f01a60272179e6001
See also #352

PR #364, opened 2020-09-21
Use LSP4MP 0.1.0

angelozerr:
Use LSP4MP 0.1.0


jeffmaury:
I am lost: the version switched back from 0.9.0 to 0.1.0 should I switched to this version in JBoss Tools ? Is it going to be the first release of LSP4MP ?

fbricon:
Yes it'll be the 1st release of lsp4mp

PR #372, opened 2020-11-20
Automatically infer package names when inserting class snippets

angelozerr:
Automatically infer package names when inserting class snippets

See https://github.com/eclipse/lsp4mp/issues/60


angelozerr:
No need to have this PR since I improved package name replace.

PR #374, opened 2020-12-04
quarkus.hibernate-orm.database.generation values should be enumerated.

rgrunber:
- Fixes redhat-developer/vscode-quarkus#317

### DIFF
--- a/quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus.test/src/main/java/com/redhat/microprofile/jdt/quarkus/QuarkusHibernateORMPropertyTest.java
+++ b/quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus.test/src/main/java/com/redhat/microprofile/jdt/quarkus/QuarkusHibernateORMPropertyTest.java
@@ -1,0 +1,49 @@
+/*******************************************************************************
+* Copyright (c) 2020 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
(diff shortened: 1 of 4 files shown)


angelozerr:
@rgrunber I suggest that you see my draft work at https://github.com/redhat-developer/quarkus-ls/compare/master...angelozerr:hibernate-orm

The idea is to have a new static file for hibernate https://github.com/redhat-developer/quarkus-ls/compare/master...angelozerr:hibernate-orm#diff-c62335edaf1084b7a021d479911a314bee47df26ec1687a228551bca1b2ddb24 and the second thing is to add description in the JSON.

rgrunber:
Ah, so a3621e8 is the change you mentioned that allows defining separate static property files. Let's just merge the draft for hibernate-orm statics if it already exists and we can get rid of this one.

angelozerr:
@rgrunber I would prefer that you do that in order to contribute step by step to quarkus-ls and we need test. Please see this sample https://github.com/redhat-developer/quarkus-ls/blob/master/quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus.test/src/main/java/com/redhat/microprofile/jdt/quarkus/containerimages/QuarkusContainerImagesTest.java

rgrunber:
> @rgrunber I would prefer that you do that in order to contribute step by step to quarkus-ls and we need test. Please see this sample https://github.com/redhat-developer/quarkus-ls/blob/master/quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus.test/src/main/java/com/redhat/microprofile/jdt/quarkus/containerimages/QuarkusContainerImagesTest.java

Alright, sounds good. I'll make the other changes.

PR #375, opened 2020-12-08
Add support for properties defined in Scheduled annotations.

rgrunber:
- Fixes redhat-developer/vscode-quarkus#319
- Add testcases

Note : This needs an addition to the MavenProjectName enum in lsp4mp for the tests to compile. Is that the right approach to test Scheduled annotations ?


angelozerr:
> Note : This needs an addition to the MavenProjectName enum in lsp4mp for the tests to compile. Is that the right approach to test Scheduled annotations ?

Indeed it's very bad. Is there any chance that you could work on that on LSP4MP side too?

angelozerr:
Please fix the conflicts

rgrunber:
> > Note : This needs an addition to the MavenProjectName enum in lsp4mp for the tests to compile. Is that the right approach to test Scheduled annotations ?
> 
> Indeed it's very bad. Is there any chance that you could work on that on LSP4MP side too?

Sure, I'll propose something on the lsp4mp side. If I understand correctly, we would also need to switch to the snapshot releases of lsp4mp in the `quarkus-ls/quarkus.jdt.ext` target platform.

angelozerr:
> Sure, I'll propose something on the lsp4mp side. If I understand correctly, we would also need to switch to the snapshot releases of lsp4mp in the quarkus-ls/quarkus.jdt.ext target platform.

Thanks @rgrunber. An idea is to replace MavenProject enumeration with an interface IMavenProject and the current MavenProject from microprofile should extend IMavenProject. Quarkus test could contribute to test by implementing IMavenProject by a QuarkusMavenProject for instance.

rgrunber:
I should probably bump to the snapshot releases in the target platform, otherwise this won't build.

rgrunber:
I noticed `.classpath` and `.project` weren't uploaded (likely due to gitignore) so hopefully that was the issue.

angelozerr:
Thanks @rgrunber !

PR #379, opened 2021-01-11
Hover support for @Scheduled annoation member values.

### LINKED ISSUE #376: Hover support for @Scheduled/cron
Do the same thing than ConfigProperty/name

rgrunber:
- Fixes #376
- Add testcase

### DIFF
--- a/quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus.test/projects/maven/scheduler-quickstart/src/main/java/org/acme/scheduler/CounterBean.java
+++ b/quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus.test/projects/maven/scheduler-quickstart/src/main/java/org/acme/scheduler/CounterBean.java
@@ -27,7 +27,7 @@ public class CounterBean {
         System.out.println(execution.getScheduledFireTime());
     }
 
-    @Scheduled(cron = "{cron.expr}")
+    @Scheduled(cron = "{cron.expr}", every = "{every.expr}")
     void cronJobWithExpressionInConfig() {
         counter.incrementAndGet();
         System.out.println("Cron expression configured in application.properties");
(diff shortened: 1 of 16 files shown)


rgrunber:
As it turns out, `MicroProfileForJavaAssert` is the only class in `org.eclipse.lsp4mp.jdt.internal.core.java` (not exported), so we might as well just move it into `org.eclipse.lsp4mp.jdt.internal.core` (is exported). The 'java' aspect is at least implicit from the 'jdt' in the namespace.

rgrunber:
I've updated this change with the new class re-locations. It depends on eclipse/lsp4mp#124

angelozerr:
@rgrunber I don't know if it's valid, but when I write 

```
@Scheduled(cron = "{cron.expr}", every = "{cron.expr}" )
void cronJobWithExpressionInConfig() {
```

and I try to hover the cron.expr from the `every` (second member), the hover doesn't work.

If I write just one member:

```
@Scheduled(every = "{cron.expr}" )
void cronJobWithExpressionInConfig() {
```

it works.

rgrunber:
This would require https://github.com/eclipse/lsp4mp/pull/126 and the test cases are adapted to test that behaviour.

quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus.test/projects/maven/scheduler-quickstart/src/main/java/org/acme/scheduler/CounterBean.java
```diff
@@ -27,7 +27,7 @@ void cronJob(ScheduledExecution execution) {
         System.out.println(execution.getScheduledFireTime());
     }
 
-    @Scheduled(cron = "{cron.expr}")
+    @Scheduled(cron = "{cron.expr}", every = "{cron.expr}")
```
angelozerr, line 30: The CI build fails, I think you should write

`every = "{every.expr}` to avoid properties duplicate.

angelozerr:
Thanks @rgrunber !

PR #381, opened 2021-02-10
Provide sourceType for QuarkusScheduledPropertiesProvider.

### LINKED ISSUE #380: Support for properties defined in Scheduled annotations should generate sourceType
When you open the application.properties and you define a scheduled cron property like:

`cron.expr=*/5 * * * * ?`

Ctrl+Click on cron.expr should open the Java class and select the method where this expression is defined. To fix that https://github.com/redhat-developer/quarkus-ls/blob/6d8fbe866d00bc7b6c83e29bc615592a46d56231/quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/providers/QuarkusScheduledPropertiesProvider.java#L66 should generate the sourceType (the class name).

rgrunber:
- Resolves #380


angelozerr:
Thanks @rgrunber !

PR #394, opened 2021-05-03
Remove git commit, push, and tag from the pipeline

### LINKED ISSUE #393: Pushing tags or commits to the repository from the Jenkins agent fails
In these lines of the Jenkinsfile: https://github.com/redhat-developer/quarkus-ls/blob/master/Jenkinsfile#L64-L76

The Jenkins agent will attempt to push to the GitHub repository. Since no GitHub user is configured, this will always fail.

Here is some options that can fix this:
* Set up the Jenkins agent with a bot GitHub account. I think this can be done through the `Jenkinsfile`.
* Remove the tag/commit/push code, and do these steps manually

datho7561:
Since no GitHub bot is configured for the pipeline, don't use `git push` in the pipeline, as this will fail.

As well, remove the `git commit` and `git tag` operations, since the result of these commands cannot be pushed to the GitHub repo.

Closes #393


rgrunber (approved):
Thanks. LGTM.

PR #395, opened 2021-07-12
Depend on lsp4mp 0.3.0 snapshot build

### LINKED ISSUE #388: Consistent use of lsp4mp in target platform
Something to consider in the future.

`com.redhat.quarkus.ls` is building against [lsp4mp 0.1.0](https://github.com/redhat-developer/quarkus-ls/blob/master/quarkus.ls.ext/com.redhat.quarkus.ls/pom.xml#L42) while the `quarkus.jdt.ext` components are using [lsp4mp 0.2.0](https://github.com/redhat-developer/quarkus-ls/blob/master/quarkus.jdt.ext/pom.xml#L20).

Given the tight integration between these, do we ever need to support newer quarkus-ls on an older lsp4mp ? It might make more sense to point to the snapshot releases of lsp4mp except for release tags (where we'd use the release values).

datho7561:
Closes #388


rgrunber (approved):
LGTM.

PR #422, opened 2021-10-19
added diagnostics and member validation for quarkus `@Scheduled` annotation

### LINKED ISSUE #378: Validation support for @Scheduled/cron
Validate syntax of cron with bracket.

AlexXuChen:
Added diagnostics and member validation for quarkus `@Scheduled` annotation from `io.quarkus.scheduler.Scheduled`.

Fixes #378 


quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/scheduler/SchedulerUtils.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.regex.Pattern;
+
+/**
+ * Utility class for quarkus scheduler
+ *
+ */
+public class SchedulerUtils {
+
+	// Day of week and day of month cannot coexist
+	private static String[] cronPatterns = { "[*?0-59]", "[*?0-59]", "[*?0-23]", "[*?1-31]", "[*?1-12]",
```
rgrunber, line 27: What about something like https://quarkus.io/guides/spring-scheduled#creating-a-scheduled-job ?

`@Scheduled(cron="*/5 * * * * ?")`

AlexXuChen: Ah gotcha, what would this mean, every 5 seconds?

rgrunber: If you click through the fields, you can see some of the supported values below.

quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/scheduler/java/QuarkusSchedulerASTVisitor.java
```diff
@@ -0,0 +1,142 @@
...
+
+	@Override
+	public boolean visit(MethodDeclaration node) {
+		try {
+			List modifiers = node.modifiers();
+			for (Object modifier : modifiers) {
+				if (modifier instanceof Annotation) {
+					Annotation annotation = (Annotation) modifier;
+					if (isMatchAnnotation(annotation, QuarkusConstants.SCHEDULED_ANNOTATION)) {
+						validateScheduledAnnotation(node, (NormalAnnotation) annotation);
```
rgrunber, line 63: If a user slowly types out `@Scheduled`, it's not a NormalAnnotation. It's a MarkerAnnotation, which can't be cast to NormalAnnotation.

quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/scheduler/java/QuarkusSchedulerASTVisitor.java
```diff
@@ -0,0 +1,142 @@
...
+	 * @param node       The method declaration to validate
+	 * @param annotation The @Scheduled annotation
+	 * @throws JavaModelException
+	 */
+	private void validateScheduledAnnotation(MethodDeclaration node, NormalAnnotation annotation)
+			throws JavaModelException {
+		Expression cronExpr = getAnnotationMemberValueExpression(annotation,
+				QuarkusConstants.SCHEDULED_ANNOTATION_CRON);
+		if (cronExpr != null) {
+			String cronValue = cronExpr.resolveConstantExpressionValue().toString();
```
rgrunber, line 87: Need to watch out for when `resolveConstantExpressionValue()` returns null. This can happen the instant a user creates the member in an annotation and it has no expression. Eg. `@Scheduled(cron = )`

```
Returns:
the constant expression value, or null if this expression has no constant expression value or if bindings were not requested when the AST was created
```

quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/scheduler/SchedulerUtils.java
```diff
@@ -0,0 +1,113 @@
...
+
+	/**
+	 * Validate the string from the @Scheduled member can be parsed to a Duration
+	 * unit
+	 *
+	 * @param timePattern the member value for the time pattern
+	 * @return the INVALID_DURATION_PARSE_PATTERN error code if invalid and null if
+	 *         valid
+	 */
+//	\\d+[smhSMH]
```
rgrunber, line 69: How about we structure this method as :
```
if (!Pattern.matches(...)) {
  try {
    ...
    ...
  } catch (...) {
    ...
  }
}
return null;
```

quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/scheduler/SchedulerUtils.java
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	/**
+	 * Match the member to an env variable pattern
+	 *
+	 * @param memberValue the member value from expression
+	 * @return true if pattern is well-formed env variable, false otherwise
+	 */
+	private static boolean checkEnvPattern(String memberValue) {
+		return Pattern.matches(envPattern, memberValue);
```
rgrunber, line 101: For the `Pattern.matches(..)` calls, you may want to consider defining the regular expression as :

`private static final Pattern MY_PATTERN = Pattern.compile(SOME_REGEX_PATTERN)` and using it in the form `MY_PATTERN.matcher(inputString).matches()`. The reason for this is that on every keystroke relating to Scheduled annotations, the regular expressions are re-compiled every single time so it's best to avoid it.

rgrunber (changes requested):
Some changes to make, but overall it seems to be working.

quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/scheduler/java/QuarkusSchedulerASTVisitor.java
```diff
@@ -0,0 +1,142 @@
...
+		}
+	}
+
+	/**
+	 * Add diagnostics for members that rely on Duration parser, i.e. every, delayed
+	 *
+	 * @param expr The expression retrieved from the annotation
+	 */
+	private void durationParseDiagnostics(Expression expr) {
+		String value = expr.resolveConstantExpressionValue().toString();
```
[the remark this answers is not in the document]
AlexXuChen: In this case, there's a MemberValueException thrown, and if it's the empty string instead (which is caught atm), this error occurs on runtime:
```
[error]: Build step io.quarkus.arc.deployment.ArcProcessor#generateResources threw an exception: javax.enterprise.inject.spi.DeploymentException: java.lang.IllegalStateException: @Scheduled must declare either cron() or every(): @Scheduled(every = "")
```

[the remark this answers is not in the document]
rgrunber: I mean if the user simply types the expression above and prepares to fill it in. Before they can get to that there will be an exception thrown up to the MultiASTValidator.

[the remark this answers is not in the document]
AlexXuChen: Ah okay gotcha

quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/scheduler/SchedulerUtils.java
```diff
@@ -0,0 +1,126 @@
...
+ * Utility class for quarkus scheduler
+ *
+ */
+public class SchedulerUtils {
+
+	// Patterns for each cron part of the cron member expression in order of
+	// SchedulerErrorCode enum ordinal
+	// Note: DAY_OF_MONTH and DAY_OF_WEEK are mutually exclusive
+
+	private static final String DAY_OF_MONTH_ABBREV = "^[mM][oO][nN]$|^[tT][uU][eE]$|^[wW][eE][dD]$|^[tT][hH][uU]$|^[fF][rR][iI]$|^[sS][aA][tT]$|^[sS][uU][nN]$";
```
AlexXuChen, line 30: It turns out that a cron expression supports 3 character abbreviations of days/months. Not only that, but it supports case mixing too, for example, you can have `jan, JAN, jAn, JAn, etc..` If there's a better way to go about this, please let me know since this regex is quite long.

rgrunber: Try https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/regex/Pattern.html#CASE_INSENSITIVE when compiling the expression.

quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/scheduler/SchedulerErrorCodes.java
```diff
@@ -0,0 +1,56 @@
...
+	INVALID_CRON_LENGTH("The cron expression must contain 6-7 parts, delimited by whitespace."),
+	INVALID_DURATION_PARSE_PATTERN("Text cannot be parsed to a Duration."),
+	INVALID_CHAR_IN_EXPRESSION("Invalid char(s) in expression."), VALID_EXPRESSION("Expression is valid.");
+
+	private final int errorCode;
+
+	private final String errorMessage;
+
+	SchedulerErrorCodes(String errorMessage) {
+		this.errorCode = ordinal();
```
rgrunber, line 43: You don't need this field. You can just return `Integer.toString(ordinal())` below.

quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/scheduler/SchedulerUtils.java
```diff
@@ -0,0 +1,128 @@
...
+	 * @param timePattern the member value for the time pattern
+	 * @return the INVALID_DURATION_PARSE_PATTERN error code if invalid and null if
+	 *         valid
+	 */
+//	\\d+[smhSMH]
+	public static SchedulerErrorCodes validateDurationParse(String timePattern) {
+		if (!TIME_PERIOD_PATTERN.matcher(timePattern).matches()) {
+			try {
+				Duration.parse(timePattern);
+				return null;
```
rgrunber, line 90: You ca remove the extra `return null;` on this line. The one at the bottom will be executed if the line above succeeds.

quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/scheduler/java/QuarkusSchedulerASTVisitor.java
```diff
@@ -0,0 +1,154 @@
...
+				String cronValue = cronExpr.resolveConstantExpressionValue().toString();
+				if (!malformedEnvDiagnostic(cronExpr, cronValue)) {
+					SchedulerErrorCodes cronPartFault = SchedulerUtils.validateCronPattern(cronValue);
+					if (cronPartFault != null) {
+						super.addDiagnostic(cronPartFault.getErrorMessage(), QuarkusConstants.QUARKUS_PREFIX, cronExpr,
+								cronPartFault, DiagnosticSeverity.Warning);
+					}
+				}
+			} catch (NullPointerException e) {
+				;
```
rgrunber, line 101: You'll have to just check `cronExpr.resolveConstantExpressionValue()` is not null prior to using it. Also there is a better method to use than `toString()`. Have a look at the return value of this method call and see if you can find a better alternative.

quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/scheduler/java/QuarkusSchedulerASTVisitor.java
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	/**
+	 * Add diagnostics for members that rely on Duration parser, i.e. every, delayed
+	 *
+	 * @param expr The expression retrieved from the annotation
+	 */
+	private void durationParseDiagnostics(Expression expr) {
+		try {
+			String value = expr.resolveConstantExpressionValue().toString();
```
rgrunber, line 123: Same thing here. Shouldn't be trying to catch an NPE this way.

rgrunber (changes requested):

quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/scheduler/SchedulerUtils.java
```diff
@@ -0,0 +1,127 @@
...
+	// Patterns for each cron part of the cron member expression in order of
+	// SchedulerErrorCode enum ordinal
+	// Note: DAY_OF_MONTH and DAY_OF_WEEK are mutually exclusive
+
+	private static final String DAY_OF_MONTH_ABBREV = "^MON$|^TUE$|^WED$|^THU$|^FRI$|^SAT$|^SUN$";
+	private static final String MONTH_ABBREV = "^JAN$|^FEB$|^MA[RY]$|^APR$|^JU[NL]$|^AUG$|^SEP$|^OCT$|^NOV$|^DEC$";
+	private static final Pattern[] CRON_PATTERNS = {
+			Pattern.compile("(^[1-5]?[0-9]$|\\*|\\?|^\\*\\/\\d+)"), // SECONDS
+			Pattern.compile("(^[1-5]?[0-9]$|\\*|\\?|^\\*\\/\\d+)"), // MINUTES
+			Pattern.compile("(^(2[0-3]|1[0-9]|[1-9])$|\\*|\\?|^\\*\\/\\d+)"), // HOUR
```
rgrunber, line 35: This doesn't support 0, although the diagnostic text does mention 0 is a valid value.

quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/scheduler/SchedulerErrorCodes.java
```diff
@@ -0,0 +1,53 @@
...
+ * enumeration list.
+ */
+@SuppressWarnings("unchecked")
+public enum SchedulerErrorCodes implements IJavaErrorCode {
+	INVALID_CRON_SECOND("Seconds must be within the range 0-59."),
+	INVALID_CRON_MINUTE("Minutes must be within the range 0-59."),
+	INVALID_CRON_HOUR("Hour must be within the range 0-23."),
+	INVALID_CRON_DAY_OF_MONTH("Day of month must be within the range 1-31."),
+	INVALID_CRON_MONTH("Month must be within the range 1-12 or a supported 3 letter abbreviation."),
+	INVALID_CRON_DAY_OF_WEEK("Day of week must be within range 1-7 or a supporeted 3 letter abbreviation."),
```
rgrunber, line 32: Supported

rgrunber:
Overall, I think this is probably ready for merging. Note that we don't support the cron format `3-5`, or `3,4,5` for a given time entry. Could we support this easily in the regular expressions or is it difficult ?

AlexXuChen:
> Overall, I think this is probably ready for merging. Note that we don't support the cron format `3-5`, or `3,4,5` for a given time entry. Could we support this easily in the regular expressions or is it difficult ?

Would that be defined as `cron = "3-5"`?

rgrunber:
> > Overall, I think this is probably ready for merging. Note that we don't support the cron format `3-5`, or `3,4,5` for a given time entry. Could we support this easily in the regular expressions or is it difficult ?
> 
> Would that be defined as `cron = "3-5"`?

I think the format would have to be something like `"3-5 * * * * *"`

AlexXuChen:
> I think the format would have to be something like `"3-5 * * * * *"`

And this would mean run this function at 3 seconds, and stop at 5?

rgrunber:
> > I think the format would have to be something like `"3-5 * * * * *"`
> 
> And this would mean run this function at 3 seconds, and stop at 5?

It would mean the same thing as `3,4,5`. Run this function on the 3rd, 4th and 5th seconds of the minute.

quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/scheduler/SchedulerUtils.java
```diff
@@ -0,0 +1,141 @@
...
+
+	private static final String MONTH_RANGE = "(JAN|FEB|MA[RY]|APR|JU[NL]|AUG|SEP|OCT|NOV|DEC|1[0-2]|[1-9])";
+
+	private static final String DAY_OF_WEEK_RANGE = "(MON|TUE|WED|THU|FRI|SAT|SUN|[1-7])";
+
+	private static final String YEAR_RANGE = "(19[7-9][0-9]|20[0-9][0-9])";
+
+	private static final String CRON_SPECIAL_CHARACTERS = "|\\*|\\?|^\\*\\/\\d+";
+
+	private static final String BASE_CRON_PATTERN = "(%1$s(?:,%1$s)*|%1$s\\-%1$s|%2$s)";
```
AlexXuChen, line 44: This accepts comma separated values and ranges if within the supported range. i.e. `3,4,5 * * * * *` and `3-5 * * * * *`  is accepted, but `3,4,60 * * * * *` and `3-60 * * * * *` is rejected.

rgrunber (approved):

Issue #416, opened 2021-09-10
Add option to work with custom host and `http://` for Quarkus project


AlexXuChen:
As per https://github.com/redhat-developer/quarkus-ls/issues/368#issue-731821303:
- there is no option to work with host differing from localhost. That's a problem.
- there is no option to work with https:// at all. Bad, but can be fixed via SSL termination or/and proxy.

PR #447, opened 2021-11-26
Don't show data model completion inside end tag section

### LINKED ISSUE #446: Don't show data model completion inside tag section
```
{#with data}
{/}

```
proposes data when code completion after {

angelozerr:
Don't show data model completion inside end tag section

Fixes #446

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/expression/ExpressionParser.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/expression/ExpressionParser.java
@@ -33,8 +33,8 @@ public class ExpressionParser {
 		}
 		Template template = expression.getOwnerTemplate();
 		String text = template.getText();
-		int start = expression.getStart() + 1;
-		int end = expression.getEnd() - 1;
+		int start = expression.getStartContentOffset();
+		int end = expression.getEndContentOffset();
 		ExpressionScanner scanner = ExpressionScanner.createScanner(text, start, end);
 		TokenType token = scanner.scan();
 		List<Node> expressionContent = new ArrayList<>();
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/expression/Part.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/expression/Part.java
@@ -63,11 +63,7 @@ public abstract class Part extends Node {
 		return parts.getPartIndex(this) == parts.getChildCount() - 1;
 	}
 
-	/**
-	 * Returns the parent section of the object part and null otherwise.
-	 * 
-	 * @return the parent section of the object part and null otherwise.
-	 */
+	@Override
 	public Section getParentSection() {
 		Parts parts = getParent();
 		Expression expression = parts.getParent();
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/template/Expression.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/template/Expression.java
@@ -40,6 +40,24 @@ public class Expression extends Node {
 		return "#expression";
 	}
 
+	/**
+	 * Returns the start offset of the expression content (after '{')
+	 * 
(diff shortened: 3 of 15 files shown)


jeffmaury (commented):
Completion is fully broken for me ex {data. completion is proposing data after the dot. Seeing a lot of NPE in the LS

angelozerr:
> Completion is fully broken for me ex {data. completion is proposing data after the dot.

Need to write more tests to fix your issue.

>  Seeing a lot of NPE in the LS

Which NPE? I noticed that we have NPE with outline (symbols) when expression is not closed. I need to fix it too. is it this NPE kind?

angelozerr:
> Completion is fully broken for me ex {data. completion is proposing data after the dot. 

I think this issue exists in the master too. If you close the expression it should work. I'm fixing that in this PR.

angelozerr:
@jeffmaury please retry. It should work better now and I have wriiten several test with completion with expression which is not closed. 

I started to write test for symbols (tests which failed with some NPE). Those tests should be improved in the future with another usecases.

fbricon:
Doesn't seem to work. Either I don't get completion at all or the with'ed object is still proposed in completion
Screenshot 2021-11-26 at 11 31 55
Screenshot 2021-11-26 at 11 32 36

fbricon:
My 2nd case is apparently invalid anyway (see #450}

jeffmaury:
Seems to be an issue with stacking #416 

```
{@qute1.MyData data}
{data.available}
{#with data}
{#with stringCollection}
{}
</with>
{/with}
```

image

angelozerr:
> Doesn't seem to work. Either I don't get completion at all 

I suppose you speak about snippets completion with each, for, etc. It's fixed and I wrote some tests.

> or the with'ed object is still proposed in completion

It's vscode completion which shows tokens, this completion doesn't comes from the Qute LS. You have the same problem with vscode-java:

image

angelozerr:
@fbricon @jeffmaury do you think we could merge this PR? There is a problem with #with but it's an another issue.

fbricon:
AFAICT, this PR doesn't fix the issue. item is still proposed inside the tag section
Screenshot 2021-11-29 at 13 28 16

jeffmaury:
> AFAICT, this PR doesn't fix the issue. item is still proposed inside the tag section Screenshot 2021-11-29 at 13 28 16

Which sounds legit to me because it's part of the upper context, no ?

angelozerr:
> Which sounds legit to me because it's part of the upper context, no ?

Exactly you can reusue inside #width the input data model. To avoid some confuse, you can use the data namespace (`data:item`). In other words you can write:

```
{@org.acme.Item item}
{#with item}
    {name}
    {item.name}
    {data:item.name}
{/}
```

fbricon:
ok so what is this PR fixing then?

angelozerr:
`|` is used whre completion is triggered:

 * Case 1:

```
{#with data}
{|/}
```

if you open completion inside `{/}` it should not provide data model.

 * Case 2:

 * `{|}` should show data model
 * `{{|}}` should NOT show data model
 * `{{{|}}}` should show data model

angelozerr:
This PR doesn't show completion inside end tag section, in master it shows:

image

It fixes NPE with outline when your template contains not closed expression. For instance type just:

`{`

and in master you should see NPE:

```
Caused by: java.lang.NullPointerException: Property must not be null: range
	at org.eclipse.lsp4j.util.Preconditions.checkNotNull(Preconditions.java:29)
	at org.eclipse.lsp4j.DocumentSymbol.<init>(DocumentSymbol.java:89)
	at org.eclipse.lsp4j.DocumentSymbol.<init>(DocumentSymbol.java:99)
	at com.redhat.qute.services.QuteSymbolsProvider.findDocumentSymbols(QuteSymbolsProvider.java:56)
	at com.redhat.qute.services.QuteSymbolsProvider.findDocumentSymbols(QuteSymbolsProvider.java:39)
	at com.redhat.qute.services.QuteLanguageService.findDocumentSymbols(QuteLanguageService.java:124)

```

It fixes too problem with property/method which is not displayed when expression is not closed (and you have none content after).

`{item.` should show method, but in master it shows again item:

image

fbricon:
Thanks @angelozerr

PR #458, opened 2021-11-26
added definition support for `@Scheduled` annotation

### LINKED ISSUE #377: Definition support for @Scheduled/cron
Do the same thing than ConfigProperty/name

AlexXuChen:
Added definition support for `@Scheduled` annotation members which have hover support.

Fixes #377

Expected behaviour:
Screenshot from 2021-11-26 18-11-22

Requires https://github.com/eclipse/lsp4mp/pull/214

### DIFF
--- a/quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/scheduler/java/QuarkusScheduledDefinitionParticipant.java
+++ b/quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/scheduler/java/QuarkusScheduledDefinitionParticipant.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright (c) 2020 Red Hat Inc. and others.
+* Copyright (c) 2021 Red Hat Inc. and others.
 *
 * This program and the accompanying materials are made available under the
 * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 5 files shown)


AlexXuChen:
Currently have a bug where the definition hover will show every property below the referenced property as well.

Peek 2021-11-29 09-50

angelozerr:
This bug seems not linked to your PR but without expression there is the same problem.

There are 2 bugs:

 * the hyperlink is not higlighted correctly (it should higlight the full property name, here it higlight only the part after '.')
 * the ranges is not correct (like you have noticed).

Please create an issue to LSP4MP.

AlexXuChen:
> This bug seems not linked to your PR but without expression there is the same problem.
> 
> There are 2 bugs:
> 
>     * the hyperlink is not higlighted correctly (it should higlight the full property name, here it higlight only the part after '.')

Yes you're right, this issue happens on master:
Screenshot from 2021-11-29 10-31-55

quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/scheduler/java/QuarkusScheduledDefinitionParticipant.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Quarkus applocation.properties Definition to navigate from Java
+ * file @Scheduled/cron, every to properties, yaml files where the property is
+ * declared.
+ *
+ */
+public class QuarkusScheduledDefinitionParticipant extends PropertiesDefinitionParticipant {
+
+	public QuarkusScheduledDefinitionParticipant() {
+		super(SCHEDULED_ANNOTATION, new String[] {
```
angelozerr, line 30: Defines this String[] in QuarkusConstants and use it for hover and definition.

angelozerr:
Could you write tests please like https://github.com/eclipse/lsp4mp/blob/master/microprofile.jdt/org.eclipse.lsp4mp.jdt.test/src/main/java/org/eclipse/lsp4mp/jdt/core/config/java/MicroProfileConfigJavaDefinitionTest.java please.

angelozerr:
Thanks @AlexXuChen !

PR #466, opened 2021-12-01
Parameter declaration support improvement

### LINKED ISSUE #459: Parameter declaration support  improvement
There are some NPE when parameter declaration is used (ex : higlight a class which doesn't define an alias). Completion of classes should be improved (ex : today we return all classes when no class is defined which could be very bad for performance).

In some cases, completion of class doesn't work (ex : when a alias is defined and some class is defined)

When an alias is defined (without class),the apply of completion generate a new alias.

This support has been developedquickly, many tests should be written to be have a parameter declaration more robust.

angelozerr:
Parameter declaration support improvement

Fixes #459

### DIFF
--- a/qute.jdt/com.redhat.qute.jdt.test/projects/maven/qute-quickstart/src/main/java/org/acme/qute/NestedClass.java
+++ b/qute.jdt/com.redhat.qute.jdt.test/projects/maven/qute-quickstart/src/main/java/org/acme/qute/NestedClass.java
@@ -1,0 +1,12 @@
+package org.acme.qute;
+
+public class NestedClass {
+
+	public class Foo {
+
+	}
+
+	public class Bar {
+
+	}
+}
--- a/qute.jdt/com.redhat.qute.jdt.test/src/main/java/com/redhat/qute/jdt/template/TemplateGetJavaTypeTest.java
+++ b/qute.jdt/com.redhat.qute.jdt.test/src/main/java/com/redhat/qute/jdt/template/TemplateGetJavaTypeTest.java
@@ -1,0 +1,150 @@
+/*******************************************************************************
+* Copyright (c) 2021 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.qute.jdt.template;
+
+import static com.redhat.qute.jdt.internal.QuteProjectTest.getJDTUtils;
+import static com.redhat.qute.jdt.internal.QuteProjectTest.loadMavenProject;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.eclipse.core.runtime.NullProgressMonitor;
+import org.junit.Assert;
+import org.junit.Test;
+
+import com.redhat.qute.commons.JavaTypeInfo;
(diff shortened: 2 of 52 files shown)


angelozerr:
This PR starts working, here a demo:

QuteParameterDeclarationDemo

I need to improve again:

 * write tests for JDT side
 * try to support generic (generate java.util.List<>
 * manage completion for packages (ex : java. should provide util package etc)

fbricon:
if you type `{@java.util.List<Ite|}` and select the proper Item type, it'll complete with `{@java.util.List<org.acme.Item item}` instead of  `{@java.util.List<org.acme.Item> item}`

angelozerr:
> if you type {@java.util.List<Ite|} and select the proper Item type, it'll complete with {@java.util.List<org.acme.Item item} instead of {@java.util.List<org.acme.Item> item}

It should be fixed

jeffmaury:
Draft and review, sounds silly to me, is it ok to review ?

angelozerr:
> Draft and review, sounds silly to me, is it ok to review ?

I need to write tests again and clean some piece of code (it's the reason why it is in draft). Otherwise features like generic should work:

QuteParameterDeclarationDemo2

angelozerr:
My code is now cleaned and tests are written. I must just write à last test on qute ls side with generic type parameter. 

So you can review it. Thanks.

angelozerr:
Ok all  tests are written now. The PR is ready to review!

jeffmaury:
`{@qute1.MyData }` not reported as error

jeffmaury:
```
{@
{data.available}
{#with data}
{#with stringCollection}
{}
{/with}
{/with}
```

Code assist is NOP after `{@`

angelozerr:
> {@qute1.MyData } not reported as error

Indeed but it's a syntax error validation and the real Qute parser doesn't report this kind of error. I process real Qute parser when validation is done and it should report it, but today it doesn't report it. My idea is that Qute use my parser (see https://github.com/redhat-developer/quarkus-ls/issues/473) and provide validator for that.

In other words I would like to support syntax error in https://github.com/redhat-developer/quarkus-ls/issues/473

mkouba:
> > {@qute1.MyData } not reported as error
> 
> Indeed but it's a syntax error validation and the real Qute parser doesn't report this kind of error. I process real Qute parser when validation is done and it should report it, but today it doesn't report it. My idea is that Qute use my parser (see #473) and provide validator for that.
> 
> In other words I would like to support syntax error in #473 (/cc @mkouba )

Yes, I think that this kind of error should result in a parser error. I'll fix the current parser right away.

angelozerr:
> Yes, I think that this kind of error should result in a parser error. I'll fix the current parser right away.

Thanks! @mkouba let me some time to support all Qute syntaxes with my Qute parser and discuss in  #473 if you think it's a good idea to adopt my parser. The benefit with my parser is that I support start/end offset for each nodes. So it's easy to give the right position.

angelozerr:
> Code assist is NOP after {@

It was a parser problem, it should be fixed with tests.

fbricon:
So if you have a nested class foo.bar.Item.NestedItem, 
{@foo proposes  foo.bar.Item.NestedItem
but {@foo.bar.Item.Nest doesn't

angelozerr:
> So if you have a nested class foo.bar.Item.NestedItem, {@foo proposes foo.bar.Item.NestedItem but {@foo.bar.Item.Nest doesn't

It should be fixed.

fbricon:
@angelozerr  it's better, but now, if you do {@foo.bar.Ite|, only foo.bar.Item is proposed,  missing foo.bar.Item.NestedItem

fbricon:
Thanks @angelozerr!

PR #475, opened 2021-12-06
Template sees parameters leaked from other templates

### LINKED ISSUE #465: Template sees parameters leaked from other templates
I have a resource class declaring/using 2 templates. In each of them, you can see the parameters declared for the other:
Screenshot 2021-12-01 at 12 12 56

Those leaked parameters are then available for completion and are validated as correct.

angelozerr:
Template sees parameters leaked from other templates

Fixes #465

### DIFF
--- a/qute.jdt/com.redhat.qute.jdt.test/projects/maven/qute-quickstart/src/main/java/org/acme/qute/HelloResource.java
+++ b/qute.jdt/com.redhat.qute.jdt.test/projects/maven/qute-quickstart/src/main/java/org/acme/qute/HelloResource.java
@@ -16,6 +16,9 @@ public class HelloResource {
     @Inject
     Template hello;
 
+    @Inject
+    Template goodbye;
+
     @GET
     @Produces(MediaType.TEXT_HTML)
     public TemplateInstance get(@QueryParam("name") String name) {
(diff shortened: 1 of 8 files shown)


angelozerr:
It should work. I need just to write some tests.

qute.jdt/com.redhat.qute.jdt.test/projects/maven/qute-quickstart/src/main/java/org/acme/qute/HelloResource.java
```diff
@@ -16,6 +16,9 @@
     @Inject
     Template hello;
 
+    @Inject
+    Template goodby;
```
fbricon, line 20: goodbye? ;-)

angelozerr: Ooops thanks :) fixed.

angelozerr:
> CI tests fail

fixed.

Now I must clean some code and improve some performance to retrieve referenced field (today I'm searching fieldreference from Java sources folder instead of searching in teh given Java class directly). The only case that I must search in sources folder is when CheckedTemplate is used as class directly (not with inner class).

angelozerr:
The PR is ready now.

fbricon:
@angelozerr thanks!

PR #476, opened 2021-12-07
Filter methods completion with Qute rules

### LINKED ISSUE #463: Filter methods completion with Qute rules
Method completion returns too many completion items which are not valid (ex : wait() method from object is not allowed). 

See rules at https://github.com/quarkusio/quarkus/blob/ce19ff75e9f732ff731bb30c2141b44b42c66050/independent-projects/qute/core/src/main/java/io/quarkus/qute/ReflectionValueResolver.java#L176

It should be nice to:

 * fix completion with those rule
 * report diagnostic to explain why wait is forbidden.

angelozerr:
Filter methods completion with Qute rules

Fixes #463

### DIFF
--- a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/QuteSupportForTemplate.java
+++ b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/QuteSupportForTemplate.java
@@ -13,7 +13,9 @@ package com.redhat.qute.jdt;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
@@ -42,6 +44,7 @@ import org.eclipse.jdt.internal.core.JavaProject;
 import org.eclipse.jdt.internal.corext.util.JavaModelUtil;
 import org.eclipse.lsp4j.Location;
 
+import com.redhat.qute.commons.InvalidMethodReason;
 import com.redhat.qute.commons.JavaFieldInfo;
 import com.redhat.qute.commons.JavaMethodInfo;
 import com.redhat.qute.commons.JavaTypeInfo;
@@ -76,6 +79,8 @@ public class QuteSupportForTemplate {
 	private static final Logger LOGGER = Logger.getLogger(QuteSupportForTemplate.class.getName());
 
 	private static final String JAVA_LANG_ITERABLE = "java.lang.Iterable";
+
+	private static final String JAVA_LANG_OBJECT = "java.lang.Object";
 
 	private static final List<String> COMMONS_ITERABLE_TYPES = Arrays.asList("Iterable", JAVA_LANG_ITERABLE,
 			"java.util.List", "java.util.Set");
@@ -284,11 +289,11 @@ public class QuteSupportForTemplate {
 		if (javaProject == null) {
 			return null;
 		}
-		String className = params.getClassName();
-		int index = className.indexOf('<');
+		String typeName = params.getClassName();
+		int index = typeName.indexOf('<');
 		if (index != -1) {
 			// ex : java.util.List<org.acme.Item>
-			String iterableClassName = className.substring(0, index);
+			String iterableClassName = typeName.substring(0, index);
 			IType iterableType = findType(iterableClassName, javaProject, monitor);
 			if (iterableType == null) {
 				return null;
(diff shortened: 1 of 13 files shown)


angelozerr:
Here a demo:

InvalidMethodDemo

I need to write tests.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/QuteSupportForTemplate.java
```diff
@@ -397,6 +407,33 @@ private static boolean isValidMethod(IMethod method, boolean isInterface) {
...
+	 * 
+	 * @return
+	 * 
+	 * @see https://github.com/quarkusio/quarkus/blob/ce19ff75e9f732ff731bb30c2141b44b42c66050/independent-projects/qute/core/src/main/java/io/quarkus/qute/ReflectionValueResolver.java#L176
+	 */
+	private static InvalidMethodReason getValidMethodForQute(IMethod method, String typeName) {
+		if (method.getNumberOfParameters() > 0) {
+			return InvalidMethodReason.HasParameters;
+		}
+		if ("java.lang.Object".equals(typeName)) {
```
fbricon, line 424: Move Object methods check first

angelozerr: fixed

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/diagnostics/QuteErrorCode.java
```diff
@@ -27,6 +27,10 @@
 	UndefinedVariable("`{0}` cannot be resolved to a variable."), //
 	UnkwownType("`{0}` cannot be resolved to a type."), //
 	UnkwownMethod("`{0}` cannot be resolved or is not a method of `{1}` Java type."), //
+	InvalidMethodVoid("Invalid `{0}` method of `{1}` Java type : void return is not allowed."), //
+	InvalidMethodHasParameters("Invalid `{0}` method of `{1}` Java type : method with parameters is not allowed."), //
+	InvalidMethodFromObject("Invalid `{0}` method of `{1}` Java type : method from java.lang.Object is not allowed."), //
```
fbricon, line 32: Can we drop the "Java type" parts? Feels verbose

angelozerr: fixed

fbricon:
mmm so, given 
```java
public class Item {
    public BigDecimal price;
    public String name;
    public BigDecimal getDiscountedPrice(String percent) {
        return price.subtract(price.multiply(new BigDecimal(percent)));
    }
}
```

I get this error:
Screenshot 2021-12-10 at 12 15 04

But Quarkus runs fine with it, computes and displays the discounted price.

@mkouba isn't Quarkus supposed to prevent using parameters with https://github.com/quarkusio/quarkus/blob/ce19ff75e9f732ff731bb30c2141b44b42c66050/independent-projects/qute/core/src/main/java/io/quarkus/qute/ReflectionValueResolver.java#L178?

angelozerr:
It means that we must remove the check with has parameters?

mkouba:
> It means that we must remove the check with has parameters?

More or less ;-). The `ReflectionValueResolver` is a fallback that works fine in JVM but for native image you'd have to register the classes for reflection. On the other hand, if a value resolver is generated for a class (e.g. all types used in all type-safe expressions) then the `ReflectionValueResolver` is not needed at all (neither in JVM nor for native image).

mkouba:
Note that in Quarkus, you can also define [template extension methods](https://quarkus.io/guides/qute-reference#template_extension_methods) that can be used to extend the set of accessible properties and methods. So it doesn't have to a class property/method. These are not available in qute standalone though.

angelozerr:
> Note that in Quarkus, you can also define template extension methods that can be used to extend the set of accessible properties and methods. So it doesn't have to a class property/method. 

Thanks for the info, but It's already supported :) This issue is about method rules. I think I will remove hasParameters check for now. @fbricon are you OK?

fbricon:
yeah remove it. 

And generally speaking, it's easier to be more lenient than the real Quarkus parser, we'll hear less complaints in that way

fbricon:
We still get a *ton of noise from collection methods for instance:

Screenshot 2021-12-13 at 12 46 14

Not sure how we're supposed to filter them out.

angelozerr:
> yeah remove it.

I did that.

> We still get a *ton of noise from collection methods for instance:

I don't know how to filter those methods, but a thing which is sure is that I don't support generic with those ethod(for completion, validation). But I think it's ananother issue.

angelozerr:
@mkouba it seems that we cannot use static method List.of.  Is it true?

mkouba:
> @mkouba it seems that we cannot use static method List.of. Is it true?

Yes, static members can only be accessed through a namespace or custom template extension methods. See also [Accessing Static Fields and Methods](https://quarkus.io/guides/qute-reference#accessing-static-fields-and-methods).

angelozerr:
> Yes, static members can only be accessed through a namespace or custom template extension methods. See also Accessing Static Fields and Methods.

Thanks for your information!

angelozerr:
@fbricon now I set static method as an error. 

When static method is annotated with TemplateExtension it should work. It misses the case with static method syantx with package name.

mkouba:
> Is that a bug?

I can't see any [template extension method](https://quarkus.io/guides/qute-reference#template_extension_methods) in your snippet. Anyway, I think that we don't handle static methods in the generated value resolver but it seems that we allow invocation of static methods via the fallback reflection resolver. In general, you can invoke a static method upon an instance in java too ...although you get a warning from the compiler and it's a bad practice. Also I think that the type-safe validation check would fail in this case. So perhaps we should remove the support for this kind of incovation from the `ReflectionValueResolver`. WDYT?

angelozerr:
> WDYT?

To be consistent and avoid some confuse, I think it should be safe to removethis kind of invocation. In other words,forbid all use of static methods.

mkouba:
> > WDYT?
> 
> To be consistent and avoid some confuse, I think it should be safe to removethis kind of invocation. In other words,forbid all use of static methods (exepct with TemplateExtension or package name).

Ok, I'll send a PR.

fbricon:
Thanks @angelozerr

mkouba:
> > Ok, I'll send a PR.
> 
> Thank so much!

https://github.com/quarkusio/quarkus/pull/22233

PR #478, opened 2021-12-09
Update o.e.jdt.ls.tp dependency to 1.7.0-SNAPSHOT

rgrunber:


rgrunber:
I think we can adopt the 1.7.0 release once JDT-LS releases in the coming week.

PR #480, opened 2021-12-10
Make clickable CodeLens data model

### LINKED ISSUE #471: Make clickable CodeLens data model
When a Qute template data model comes from Template field or CheckedTemplate, codelens display class and parameters.

It should be nice to provide the capability to click on those codelens and jump to the proper class, method parameters.

angelozerr:
Make clickable CodeLens data model

Fixes #471


angelozerr:
Here a demo with clickable datamodel codelens:

ClickableCodeLensDemo

PR #484, opened 2021-12-13
`this` parameter support

angelozerr:
`this` parameter support

Fixes 483


angelozerr:
Here a demo with `this`:

ThisParameterDemo

@martin could you confirm, that Qute works like this? I have not found some sample in the documentation with `this` (just see that in your tests).

mkouba:
`{this}` should be deprecated. The only reasonable use case is documented for the [`{#with}`](https://quarkus.io/guides/qute-reference#with_section) section which is also a candidate for deprecation and removal.

In general, `{this}` refers to the [current context](https://quarkus.io/version/main/guides/qute-reference#current_context_object), i.e. may be different within different parts of the template.

angelozerr:
> {this} should be deprecated. 

what about the parameter name when you write 

```java
private Template hello;

...
Map data = ...
hello.data(data)
```

This PR supports that. It uses `this` as parameter name

angelozerr:
@mkouba when you will have time, could you answer to my question please.

mkouba:
> > {this} should be deprecated.
> 
> what about the parameter name when you write
> 
> ```java
> private Template hello;
> 
> ...
> Map data = ...
> hello.data(data)
> ```
> 
> This PR supports that. It uses `this` as parameter name

Well, you don't need to use `this` even for `template.data(data)`. I mean you can just use `{name}` instead of `{this.name}`...

fbricon:
should we close that PR?

angelozerr:
As this support seems not very important I stop working on it. When I will have time I will try to rework again

PR #487, opened 2021-12-14
Fix completion with @TemplateExtension

angelozerr:
Fix completion with @TemplateExtension


fbricon:
Completion is there, now I wonder if we shouldn't display the method as halfPrice() instead of halfPrice(Item item)
Screenshot 2021-12-14 at 16 33 55

angelozerr:
> Completion is there, now I wonder if we shouldn't display the method as halfPrice() instead of halfPrice(Item item)

I agree, I create an issue for that at https://github.com/redhat-developer/quarkus-ls/issues/490

angelozerr:
I fixed the conflicts, PR is ready.

angelozerr:
@fbricon can we merge this PR?

PR #494, opened 2021-12-15
Qute validation easily goes out-of-sync when link-editing

### LINKED ISSUE #488: Qute validation easily goes out-of-sync when link-editing
Try renaming `item` in 
```
{#for item in items}
<h2>{item.name}</h2>
{/}
```
Unless you type very slow, the validation will go out-of-sync pretty quick.

Dec-14-2021 19-26-57

angelozerr:
Qute validation easily goes out-of-sync when link-editing

Fixes #488

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/template/TemplateFileTextDocumentService.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/template/TemplateFileTextDocumentService.java
@@ -74,7 +74,7 @@ public class TemplateFileTextDocumentService extends AbstractTextDocumentService
 	public TemplateFileTextDocumentService(QuteLanguageServer quteLanguageServer, SharedSettings sharedSettings) {
 		super(quteLanguageServer, sharedSettings);
 		this.documents = new QuteTextDocuments((document, cancelChecker) -> {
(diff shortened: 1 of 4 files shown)


angelozerr:
The issue should be fixed, the validation should be stopped as soon as the content of Qute template changes.

fbricon:
Nice work @angelozerr!

angelozerr:
> Nice work @angelozerr!

Thanks @fbricon 

The thing that I have noticed is when there is linked editing range is that we have 2 didChange:

 * one didChange with single change coming from the character that user types
 * one didChange with several changes coming from linked editing range.

We trigger validation for each didChange, so there are:

 * one publish diagnostic when the first didChange is reporting diagnostics and didChange comes from at this time. The second didChange cancel the previous diagnostics
 * or 2 publish diagnostics (one per didChange). 

In the last case, you can see that the validationof firdt didChange will report the linked editing range content as error. The second didChange will remove those errors. In other words when validation is done fastly, you can see warning and few ms after warning disappears.

@rgrunber if I understand correctly you manage this usecase in JDT LS, no?

rgrunber:
JDT-LS had a fixed delay (~400ms, now made adaptive based on 1.5 x average validation time) on calls to the validation job. I'm not completely sure of how it works but it looks to me like new calls to the validation simply cancel existing ones (hence the delay, to ensure that occurs). I'm not sure why we aren't using the cancel checker though.

PR #496, opened 2021-12-15
Loop parameter leak into for-else block

### LINKED ISSUE #482: Loop parameter leak into for-else block
When a for loop has an [else](https://quarkus.io/guides/qute-reference#loop_section) bock, the item parameter leaks into that else block, it shouldn't be accessible.
```
{#for item in items}
  {item.name}
{#else}
    {ite|}
  No items.
{/for}
```
Screenshot 2021-12-13 at 12 39 21

Screenshot 2021-12-13 at 12 36 46

angelozerr:
Loop parameter leak into for-else block

Fixes #482

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/template/sections/BaseWhenSection.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/template/sections/BaseWhenSection.java
@@ -11,9 +11,12 @@
 *******************************************************************************/
 package com.redhat.qute.parser.template.sections;
 
+import java.util.List;
+
 import com.redhat.qute.parser.template.Parameter;
 import com.redhat.qute.parser.template.ParametersInfo;
 import com.redhat.qute.parser.template.Section;
+import com.redhat.qute.parser.template.SectionKind;
 
 /**
  * Base class for #switch and #when section.
@@ -46,4 +49,8 @@ public abstract class BaseWhenSection extends Section {
 		return PARAMETER_INFOS;
 	}
 
+	@Override
+	public List<SectionKind> getBlockLabels() {
+		return List.of(SectionKind.IS, SectionKind.CASE, SectionKind.ELSE);
+	}
 }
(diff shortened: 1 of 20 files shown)


angelozerr:
It starts working, need to write tests:

ElseInForDemo

I wonder if metadata like item_count should be ignored in else block like item alias.

fbricon:
> I wonder if metadata like item_count should be ignored in else\ block like item alias.

In that else block, you get : 
> io.quarkus.qute.TemplateException: Entry "item_count" not found in the data map in expression {item_count} in template items on line 10

fbricon:
Loop metadata should not be proposed in that else block:
Screenshot 2021-12-15 at 19 26 54

angelozerr:
> Loop metadata should not be proposed in that else block:

fixed.

This PR too highlight for nested block. See the following demo:

NestedBlockDemo

It's helpful when you write a #else for #for and for #if and you need to see quickly where the #else belongs to.

I did that too for #switch and #case

angelozerr:
@fbricon you can play with the PR. I need to write tests now.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/template/sections/BaseWhenSection.java
```diff
@@ -46,4 +51,8 @@ public ParametersInfo getParametersInfo() {
 		return PARAMETER_INFOS;
 	}
 
+	@Override
+	public List<SectionKind> getBlockLabels() {
+		return ImmutableList.of(SectionKind.IS, SectionKind.CASE, SectionKind.ELSE);
```
fbricon, line 56: use List.of()

angelozerr: thanks fixed

angelozerr:
Tests are written now.

fbricon:
thanks @angelozerr

PR #500, opened 2021-12-16
Changed Qute embedded Java completion to use simple types

### LINKED ISSUE #490: Display simple type for completion
Today completion shows parameter type with packages:

image

It should be better to display as simple type (like Java completion).

There is an utility class which does that at https://github.com/redhat-developer/quarkus-ls/blob/8c06be8766da9c247da5a31536902410e40e5401/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaElementInfo.java#L112

AlexXuChen:
Changed Qute embedded Java completion to display simple types instead of resolved name.

Fixes #490 

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaMethodInfo.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaMethodInfo.java
@@ -13,12 +13,13 @@ package com.redhat.qute.commons;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.StringJoiner;
 
 import org.eclipse.xtext.xbase.lib.util.ToStringBuilder;
 
 /**
  * Java method information.
- * 
+ *
  * @author Angelo ZERR
  *
  */
@@ -34,34 +35,61 @@ public class JavaMethodInfo extends JavaMemberInfo {
 
 	/**
 	 * Returns the Java method signature.
-	 * 
+	 *
 	 * Example:
-	 * 
+	 *
 	 * <code>
 	 * find(query : java.lang.String, params : java.util.Map<java.lang.String,java.lang.Object>) : io.quarkus.hibernate.orm.panache.PanacheQuery<T>
 	 * </code>
-	 * 
+	 *
 	 * @return the Java method signature.
 	 */
+	@Override
 	public String getSignature() {
 		return super.getSignature();
 	}
 
 	/**
-	 * Returns the method name.
-	 * 
+	 * Returns the Java method signature with simple names.
+	 *
 	 * Example:
-	 * 
+	 *
+	 * <code>
+	 * find(query : String, params : Map<String,Object>) : PanacheQuery<T>
+	 * </code>
+	 *
+	 * @return the Java method signature.
+	 */
+	public String getSimpleSignature() {
+		StringBuilder simpleSignatureBuilder = new StringBuilder();
+		simpleSignatureBuilder.append(getName());
+		simpleSignatureBuilder.append("(");
+		List<JavaParameterInfo> parameters = getParameters();
+		StringJoiner commaJoiner = new StringJoiner(", ");
+		for (JavaParameterInfo parameter : parameters) {
+			commaJoiner.add(parameter.getSimpleParameter());
+		}
+		simpleSignatureBuilder.append(commaJoiner.toString());
+		simpleSignatureBuilder.append(") : ");
+		simpleSignatureBuilder.append(super.getSimpleType(getReturnType()));
+		return simpleSignatureBuilder.toString();
+	}
+
+	/**
+	 * Returns the method name.
+	 *
+	 * Example:
+	 *
 	 * <code>
 	 *  find
 	 *  </code>
-	 * 
+	 *
 	 * from the given signature:
-	 * 
+	 *
 	 * <code>
 	 * find(query : java.lang.String, params : java.util.Map<java.lang.String,java.lang.Object>) : io.quarkus.hibernate.orm.panache.PanacheQuery<T>
 	 * </code>
-	 * 
+	 *
 	 * @return the method name.
 	 */
 	@Override
@@ -77,22 +105,22 @@ public class JavaMethodInfo extends JavaMemberInfo {
 		}
 		return methodName;
 	}
-	
+
 	/**
 	 * Returns the method return Java type and null otherwise.
-	 * 
+	 *
 	 * Example:
-	 * 
+	 *
 	 * <code>
 	 *  io.quarkus.hibernate.orm.panache.PanacheQuery<T>
 	 *  </code>
-	 * 
+	 *
 	 * from the given signature:
-	 * 
+	 *
 	 * <code>
 	 * find(query : java.lang.String, params : java.util.Map<java.lang.String,java.lang.Object>) : io.quarkus.hibernate.orm.panache.PanacheQuery<T>
 	 * </code>
-	 * 
+	 *
 	 * @return the method return Java type and null otherwise.
 	 */
 	public String getReturnType() {
@@ -107,7 +135,7 @@ public class JavaMethodInfo extends JavaMemberInfo {
 
 	/**
 	 * Returns the getter name of the method name.
-	 * 
+	 *
 	 * @return the getter name of the method name.
 	 */
 	public String getGetterName() {
@@ -136,7 +164,7 @@ public class JavaMethodInfo extends JavaMemberInfo {
 
 	/**
 	 * Returns true if the method have parameters and false otherwise.
-	 * 
+	 *
 	 * @return true if the method have parameters and false otherwise.
 	 */
 	public boolean hasParameters() {
@@ -148,9 +176,9 @@ public class JavaMethodInfo extends JavaMemberInfo {
 
 	/**
 	 * Returns the Java method parameter at the given index and null otherwise.
-	 * 
+	 *
 	 * @param index the method parameter index
-	 * 
+	 *
 	 * @return the Java method parameter at the given index and null otherwise.
 	 */
 	public JavaParameterInfo getParameterAt(int index) {
@@ -160,7 +188,7 @@ public class JavaMethodInfo extends JavaMemberInfo {
 
 	/**
 	 * Returns the method parameters.
-	 * 
+	 *
 	 * @return the method parameters.
 	 */
 	public List<JavaParameterInfo> getParameters() {
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForExpression.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForExpression.java
@@ -59,7 +59,7 @@ import com.redhat.qute.utils.StringUtils;
 
 /**
  * Qute completion inside Qute expression.
- * 
+ *
  * @author Angelo ZERR
  *
  */
@@ -74,7 +74,7 @@ public class QuteCompletionsForExpression {
 	/**
 	 * Returns the completion result of the given offset inside the given
 	 * expression.
-	 * 
+	 *
 	 * @param expression         the expression where the completion has been
 	 *                           triggered and null otherwise.
 	 * @param nodeExpression     the node expression where the completion has been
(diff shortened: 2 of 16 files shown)


qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForExpression.java
```diff
@@ -251,17 +252,18 @@ private void fillCompletionFields(ResolvedJavaTypeInfo resolvedType, Range range
 			// class A extends B {String foo;}
 			// class B {String foo;}
 			if (!existingProperties.contains(fieldName)) {
+				String simpleFieldName = getSimpleType(fieldName);
```
angelozerr, line 255: Why simpleFieldName? Its the type we must simplified not the name.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForExpression.java
```diff
@@ -251,17 +252,18 @@ private void fillCompletionFields(ResolvedJavaTypeInfo resolvedType, Range range
 			// class A extends B {String foo;}
 			// class B {String foo;}
 			if (!existingProperties.contains(fieldName)) {
+				String simpleFieldName = getSimpleType(fieldName);
 				CompletionItem item = new CompletionItem();
-				item.setLabel(fieldName);
-				item.setLabel(fieldName + " : " + field.getType());
-				item.setFilterText(fieldName);
+				item.setLabel(simpleFieldName);
```
angelozerr, line 257: remove this line

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForExpression.java
```diff
@@ -299,15 +301,16 @@ private void fillCompletionMethods(ResolvedJavaTypeInfo resolvedType, Range rang
 				existingMethodSignatures.add(methodSignature);
 				String property = method.getGetterName();
 				if (property != null && !existingProperties.contains(property)) {
+					String simpleProperty = getSimpleType(property);
```
angelozerr, line 304: Same thing for here. It's just the type that it must simplifed.

angelozerr:
It seems that it misses simplified type of parameters method and return type of method.

AlexXuChen:
> It seems that it misses simplified type of parameters method and return type of method.

The parameters are added in `fillCompletionMethods` and `fillCompletionFields`, as this is where the qute completion list is populated.

angelozerr:
For method you must update the method signature with à simple signature at https://github.com/redhat-developer/quarkus-ls/blob/edaf4cc119554358d4d2629c799c2d6d7ad58b84/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForExpression.java#L344

To do that I suggest you that you create à getSimpleSignature in JavaMethodInfo. This class  provides method like getreturntype getparameyers getname that you van use to implement getsimplesignature

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForExpression.java
AlexXuChen: You can see below that this fix does the following for `CASE_INSENSITIVE`:
Screenshot from 2021-12-16 14-27-37
Which is now of type `Comparator<String>` rather then `java.util.Comparator<java.lang.String>` as it was before.

AlexXuChen: It would be ideal if my feature is tested before review, since I covered the case we mentioned in the original issue.

angelozerr: I think the fix should be just:

`item.setLabel(fieldName + " : " + getSimpleType(field.getType()));`

fieldName doesn't contains some '.', so you don't need to simplify field name.

> It would be ideal if my feature is tested before review, since I covered the case we mentioned in the original issue.

Indeed you cover it, but you don't need to simply field name.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForExpression.java
```diff
@@ -299,15 +301,16 @@ private void fillCompletionMethods(ResolvedJavaTypeInfo resolvedType, Range rang
 				existingMethodSignatures.add(methodSignature);
 				String property = method.getGetterName();
 				if (property != null && !existingProperties.contains(property)) {
+					String simpleProperty = getSimpleType(property);
 					// It's a getter method, create a completion item for simple property (value)
 					// from the method name (getValue)
 					CompletionItem item = new CompletionItem();
-					item.setLabel(property + " : " + method.getReturnType());
-					item.setFilterText(property);
+					item.setLabel(simpleProperty + " : " + getSimpleType(method.getReturnType()));
```
angelozerr, line 308: replace with

`item.setLabel(property + " : " + getSimpleType(method.getReturnType()));`

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForExpression.java
angelozerr, line 304: Remove simpleProperty.

angelozerr:
> The parameters are added in fillCompletionMethods and fillCompletionFields, as this is where the qute completion list is populated.

You will see the problem with 

 * `compareTo` method:

image

 * `codePoints `method:

image

To fix that see https://github.com/redhat-developer/quarkus-ls/pull/500#issuecomment-996088525

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaMethodInfo.java
```diff
@@ -34,34 +35,75 @@
...
+		String parameterString = signature.substring(signature.indexOf('(') + 1, signature.indexOf(')'));
+		if (!parameterString.isEmpty()) {
+			String[] parameters = parameterString.split(",");
+			StringJoiner commaJoiner = new StringJoiner(",");
+			for (String parameter : parameters) {
+				commaJoiner.add(getSimpleParameter(parameter));
+			}
+			simpleSignatureBuilder.append(commaJoiner.toString());
+		}
+		simpleSignatureBuilder.append(") : ");
```
AlexXuChen, line 78: Currently, primitive types have double spaces, but reference types don't

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaMethodInfo.java
```diff
@@ -34,34 +35,75 @@
...
+	 * Example:
+	 *
+	 * <code>
+	 * find(query : String, params : Map<String,Object>) : PanacheQuery<T>
+	 * </code>
+	 *
+	 * @return the Java method signature.
+	 */
+	public String getSimpleSignature() {
+		String signature = super.getSignature();
```
angelozerr, line 64: Don't use signature, bute use method like getName, getParameters, getReturnType.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaMethodInfo.java
```diff
@@ -34,34 +35,75 @@
...
+		StringBuilder simpleSignatureBuilder = new StringBuilder();
+		String methodName = signature.substring(0, signature.indexOf('('));
```
angelozerr, line 66: String methodName = getName();

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaMethodInfo.java
```diff
@@ -34,34 +35,75 @@
...
+		if (!parameterString.isEmpty()) {
+			String[] parameters = parameterString.split(",");
+			StringJoiner commaJoiner = new StringJoiner(",");
+			for (String parameter : parameters) {
+				commaJoiner.add(getSimpleParameter(parameter));
+			}
+			simpleSignatureBuilder.append(commaJoiner.toString());
+		}
+		simpleSignatureBuilder.append(") : ");
+		String returnType = signature.substring(signature.lastIndexOf(':') + 1, signature.length()).trim();
```
angelozerr, line 79: String returnType = getReturnType();

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaMethodInfo.java
```diff
@@ -34,34 +35,75 @@
...
+	 *
+	 * @return the Java method signature.
+	 */
+	public String getSimpleSignature() {
+		String signature = super.getSignature();
+		StringBuilder simpleSignatureBuilder = new StringBuilder();
+		String methodName = signature.substring(0, signature.indexOf('('));
+		simpleSignatureBuilder.append(methodName);
+		simpleSignatureBuilder.append("(");
+		String parameterString = signature.substring(signature.indexOf('(') + 1, signature.indexOf(')'));
```
angelozerr, line 69: use `getParameters()` instead of using signature.

angelozerr:
Add a test in this line https://github.com/redhat-developer/quarkus-ls/blob/edaf4cc119554358d4d2629c799c2d6d7ad58b84/qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/commons/JavaMethodInfoTest.java#L45 to check the value of `getSimpleSignature`.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForExpression.java
```diff
@@ -302,7 +302,7 @@ private void fillCompletionMethods(ResolvedJavaTypeInfo resolvedType, Range rang
 					// It's a getter method, create a completion item for simple property (value)
 					// from the method name (getValue)
 					CompletionItem item = new CompletionItem();
-					item.setLabel(property + " : " + method.getReturnType());
+					item.setLabel(property + " : " + getSimpleType(method.getReturnType()));
```
angelozerr, line 305: `item.setLabel(property + " : " + method.getJavaElementSimpleType());`

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForExpression.java
```diff
@@ -252,8 +253,7 @@ private void fillCompletionFields(ResolvedJavaTypeInfo resolvedType, Range range
 			// class B {String foo;}
 			if (!existingProperties.contains(fieldName)) {
 				CompletionItem item = new CompletionItem();
-				item.setLabel(fieldName);
-				item.setLabel(fieldName + " : " + field.getType());
+				item.setLabel(fieldName + " : " + getSimpleType(field.getType()));
```
angelozerr, line 256: At first you can simplify with

`item.setLabel(fieldName + " : " + field.getJavaElementSimpleType());`

but to be consistent with method, I will move this code in getSimpleSignature in JavaFieldInfo.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaMethodInfo.java
```diff
@@ -34,34 +35,83 @@
...
+	 * <code>
+	 * query : java.lang.String
+	 * </code> becomes <code>
+	 * query : String
+	 * </code>
+	 *
+	 * @param parameter the Java method parameter
+	 * @return the Java method signature with simple type.
+	 */
+	private String getSimpleParameter(JavaParameterInfo parameter) {
```
angelozerr, line 92: Move this code in JavaParameterInfo#getSimpleSignature

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForExpression.java
```diff
@@ -424,7 +424,7 @@ private void doCompleteExpressionForObjectPartWithParentNodes(Node part, Node no
 			return;
 		}
 		if (parent.getKind() == NodeKind.Section) {
-			Section section = (Section) parent;
+			Section section = parent;
```
angelozerr, line 427: Remove Section declaration and use parent variable for the following code

angelozerr:
Nice job @AlexXuChen !

PR #504, opened 2021-12-22
Support for method parameters

### LINKED ISSUE #486: Support for method parameters
Support for method parameters:

 * [ ] validate method parameter type
 * [ ] validate method parameter number
 * [ ] completion inside method parameter
 * [ ] highlight, linked editing range in method parameter

angelozerr:
Support for method parameters

Fixes #486


angelozerr:
It should start working, see the following demo:

MethodDemo

angelozerr:
Please note this PR doesn't take care of methods which have the same name but with different parameters (today it take the first signature found for a given method name when validation is done, when hover is done).

I would like to support that  in an another PR please.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/utils/QuteSearchUtils.java
```diff
@@ -289,11 +292,45 @@ public static void tryToCollectObjectPartOrParameter(String partName, PartNameMa
...
-		ObjectPart objectPart = parameterExpr.getObjectPart();
-		if (isMatch(objectPart, partName, matcher)) {
-			Range range = QutePositionUtility.createRange(objectPart);
-			collector.accept(objectPart, range);
-			return true;
+		boolean result = false;
+		List<Node> nodes = parameterExpr.getExpressionContent();
+		for (Node nodeExpr : nodes) {
+			result |= tryToCollectObjectPart(partName, matcher, nodeExpr, collector);
+		}
```
fbricon, line 299: do we really need to go through all nodes after we found one valid match?

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/parameter/scanner/ParameterScanner.java
```diff
@@ -15,27 +15,40 @@
...
+ *
+ */
 public class ParameterScanner extends AbstractScanner<TokenType, ScannerState> {
 
 	private static final Predicate<Integer> PARAMETER_NAME_OR_VALUE_PREDICATE = ch -> {
 		return ch != ' ' && ch != '=';
 	};
-	
-	public static ParameterScanner createScanner(String input) {
-		return createScanner(input, 0, input.length());
```
fbricon, line 25: or you could keep that method and call `createScanner(input, 0, input.length(), false);`, thus keeping ParameterScannerTest.java simple

fbricon:
why is there a  qute.ls/com.redhat.qute.ls/src/test/java/toremove/ package?

angelozerr:
This PR shows now the method parameters in the hover:

image

angelozerr:
> why is there a qute.ls/com.redhat.qute.ls/src/test/java/toremove/ package?

I used his package to test qute with simple section if , for, etc. I think we should keep those java main test, but perhaps we should change the package name, what do you think @fbricon ?

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/parameter/scanner/ParameterScanner.java
angelozerr: fixed

fbricon:
> > why is there a qute.ls/com.redhat.qute.ls/src/test/java/toremove/ package?
> 
> I used his package to test qute with simple section if , for, etc. I think we should keep those java main test, but perhaps we should change the package name, what do you think @fbricon ?

yes rename it please, it's confusing

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/utils/QuteSearchUtils.java
angelozerr: I have add comments for this code, hope you will understand more. The code collect all object parts inside nodeExpr. The object part can be declared inside an expression or a parameter method, so we need to loop all nodes.

angelozerr:
> yes rename it please, it's confusing

renamed with `sandbox`

fbricon:
thanks @angelozerr!

PR #510, opened 2022-01-04
Validate overloaded methods

angelozerr:
Validate overloaded methods

### DIFF
--- a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaMemberInfo.java
+++ b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaMemberInfo.java
@@ -21,24 +21,24 @@ public abstract class JavaMemberInfo extends JavaElementInfo {
 
 	protected static final String NO_VALUE = "~";
 
-	private transient ResolvedJavaTypeInfo resolvedType;
+	private transient JavaTypeInfo javaType;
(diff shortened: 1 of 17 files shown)


angelozerr:
This PR improves validation:

 * for template extension when it is not applicable for teh base object:
 
image

 * for method which have same name but with several signatures (like String.getBytes):

image

angelozerr:
Please note this PR uses parameter types for validation. Hover and definition doesn't use those parameter types because it will require a lot of refactoring to compute parameter types which is done for the moment in diagnostics only.

When you hover getBytes, you will see type of the first getBytes, so the hover can be wrong for parameters. I would liketo do that in an another PR in the future.

fbricon:
If I have an extension method halfPrice(Item item), when calling {item.halfPrice()} I get an "The virtual method `halfPrice` in the type `TemplateExtensions` is not applicable for the base object `Item` type." error message, but Quarkus is totally fine running with it.

angelozerr:
> If I have an extension method halfPrice(Item item), when calling {item.halfPrice()} I get an "The virtual method halfPrice in the type TemplateExtensions is not applicable for the base object Item type." error message, but Quarkus is totally fine running with it.

Oooops sorry, I introduced a bug , I have noticed that because test doesn't covered this usecase. Now it should be fixed with test.

fbricon:
thanks @angelozerr

PR #513, opened 2022-01-05
Validation fail when iterating over integers

### LINKED ISSUE #511: Validation fail when iterating over integers
According to https://quarkus.io/guides/qute-reference#loop_section

```
The for statement also works with integers, starting from 1. In the example below, considering that total = 3:

{#for i in total}
  {i}:
{/for}
And the output will be:

1:2:3:

```
But the qute ls complains about it:

Screenshot 2022-01-05 at 15 32 09

angelozerr:
Validation fail when iterating over integers

Fixes #511

### DIFF
--- a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/ResolvedJavaTypeInfo.java
+++ b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/ResolvedJavaTypeInfo.java
@@ -144,6 +144,16 @@ public class ResolvedJavaTypeInfo extends JavaTypeInfo {
 		return isIterable.booleanValue();
 	}
 
+	/**
+	 * Returns true if the Java type is an integer and false otherwise.
+	 * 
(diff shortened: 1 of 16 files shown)


angelozerr:
It should work but I need to write test and clean code.

fbricon:
looks like you don't get anything when hovering over `i` in, while hovering over `total` shows `int` 
```
{#for i in total}
  {i}:
{/}
```

angelozerr:
> looks like you don't get anything when hovering over i in, while hovering over total shows int

try to rename i with foo (to have a longer variable), I think it should work.

angelozerr:
> looks like you don't get anything when hovering over i in, while hovering over total shows int

I improved the hover position and now I accept this usecase:

```
{#for i in total}
  {|i}:
{/}
```

when hover/definition is done after a start expression, I include the part.

angelozerr:
Here a demo:

IntegersDemo

fbricon:
thanks @angelozerr !

PR #516, opened 2022-01-06
Literal support as object

angelozerr:
Literal support as object

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/QuteProjectRegistry.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/QuteProjectRegistry.java
@@ -11,6 +11,7 @@
 *******************************************************************************/
 package com.redhat.qute.project;
 
+import static com.redhat.qute.parser.template.LiteralSupport.getPrimitiveObjectType;
 import static com.redhat.qute.services.QuteCompletableFutures.EXTENDED_TEMPLATE_DATAMODEL_NULL_FUTURE;
 import static com.redhat.qute.services.QuteCompletableFutures.RESOLVED_JAVA_CLASSINFO_NULL_FUTURE;
 import static com.redhat.qute.services.QuteCompletableFutures.VALUE_RESOLVERS_NULL_FUTURE;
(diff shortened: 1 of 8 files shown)


angelozerr:
This PR uses java.lang.Integer instead of intfor literal support:

image

I need to clean my code and write some tests again.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/QuteProjectRegistry.java
```diff
@@ -81,6 +84,32 @@ private static void registerPrimitiveType(String type) {
...
 				CompletableFuture.completedFuture(primitiveTypeArray));
+
+		String objectType = getPrmitiveObjectType(type);
+		if (objectType != null) {
+			autoboxing.put(type, type);
+			autoboxing.put(objectType, type);
+		}
+	}
+
+	private static String getPrmitiveObjectType(String type) {
```
fbricon, line 95: getPrimitiveObjectType

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/QuteProjectRegistry.java
```diff
@@ -81,6 +84,32 @@ private static void registerPrimitiveType(String type) {
...
+		case "char":
+			return "java.lang.Character";
+		case "double":
+			return "java.lang.Double";
+		case "float":
+			return "java.lang.Float";
+		case "int":
+			return "java.lang.Integer";
+		case "long":
+			return "java.lang.Long";
```
fbricon, line 110: maybe reference constants from LiteralSupport.java

fbricon (changes requested):

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/QuteProjectRegistry.java
angelozerr: fixed

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/QuteProjectRegistry.java
angelozerr: Very good suggestion, I have moved the function in LiteralSupport class. fixed.

fbricon:
Is it normal to still see `int` for {it} in this case?
```
{#each items.size}
  {it}
{/}
```

angelozerr:
> Is it normal to still see int for {it} in this case?

Yes it's the type returned by JDT.

fbricon:
Thanks @angelozerr

PR #535, opened 2022-01-12
CodeAction to disable validation

### LINKED ISSUE #531: CodeAction to disable validation
When there is an error in qute template it should be nice to provide a code action to disable validation in other words update the proper validation settings like we do for application.properties validation.

When validation settings severity for each error code will be implemented it should be nice to provide an another code action to disable the settings for the given error code.

angelozerr:
CodeAction to disable validation

Fixes #531

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/commons/CodeActionFactory.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/commons/CodeActionFactory.java
@@ -182,12 +182,25 @@ public class CodeActionFactory {
 	 */
 	public static CodeAction createCommand(String title, String commandId, List<Object> commandParams,
 			Diagnostic diagnostic) {
+		return createCommand(title, commandId, commandParams, Collections.singletonList(diagnostic));
+	}
+
+	/**
+	 * Makes a CodeAction to call a command from the available server commands.
+	 *
+	 * @param title         The displayed name of the CodeAction
+	 * @param commandId     The id of the given command to add as CodeAction
+	 * @param commandParams The document URI of the document the command is called
+	 *                      on
+	 * @param diagnostics   The diagnostic list that this CodeAction will fix
+	 */
+	public static CodeAction createCommand(String title, String commandId, List<Object> commandParams,
+			List<Diagnostic> diagnostics) {
 		CodeAction codeAction = new CodeAction(title);
 		Command command = new Command(title, commandId, commandParams);
 		codeAction.setCommand(command);
-		codeAction.setDiagnostics(Collections.singletonList(diagnostic));
+		codeAction.setDiagnostics(diagnostics);
 		codeAction.setKind(CodeActionKind.QuickFix);
-
 		return codeAction;
 	}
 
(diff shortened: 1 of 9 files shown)


angelozerr:
This PR requires the PR on vscode side https://github.com/redhat-developer/vscode-quarkus/pull/435

Please don't review the code (I did that very quickly) but it should work.

angelozerr:
Here a demo:

DisableValidation

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/diagnostics/TemplateCodeActionFactory.java
```diff
@@ -0,0 +1,76 @@
...
+ * @author Angelo ZERR
+ *
+ */
+public class TemplateCodeActionFactory {
+
+	// Enable/Disable Qute validation
+
+	private static final String UNASSIGNED_EXCLUDED_SECTION = "quarkus.tools.qute.validation.enabled";
+
+	private static final String UNASSIGNED_EXCLUDE_CODE_ACTION_TITLE = "Do you want to disable Qute validation?";
```
fbricon, line 39: Disable Qute validation

Not a question

angelozerr: fixed

angelozerr:
@fbricon you can review it now.

fbricon:
So @mkouba suggests we have a validation setting per file instead. The setting would use an array of path/boolean entries instead of a single boolean. Or maybe we can do both (array OR boolean)

angelozerr:
> So @mkouba suggests we have a validation setting per file instead. 

The main problem with that is when your rename template file, it will require to update settings.json too. I think it's more work. Having a global validation enabled settings can be a good start. We can improve it in the future release, no?

fbricon:
I think it's simple enough to support now and having to rename the value in the settings is only a small inconvenience. So I'd prefer having it.

angelozerr:
Ok but I think we should have this global validation settings to disable validation for all template files like we have for microprofile and xml language server.

Could we do that in an another PR?

angelozerr:
And I wonder if we must support pattern file instead of full file name because the question is do we need to store teh full pzth of the template or just the uri of the template (I mean the location uri with the project root basedir)?

angelozerr:
@fbricon have you some idea about the settings name that we can use it to define a set of templates. 

Quickly, I think about quickly `quarkus.tools.qute.validation.files.excluded` String array, but I'm not very happy with this name. More I wonder if we must give the capability to defines some set of files and for each set, we could customize the validation settings (severity for each error code, enabled validation for the declared files template).

angelozerr:
@fbricon could we merge the PR or do you prefer I implement excluded inside this PR?

angelozerr:
> I think it's simple enough to support now and having to rename the value in the settings is only a small inconvenience. So I'd prefer having it.

let's continue the discussion at https://github.com/redhat-developer/quarkus-ls/issues/531#issuecomment-1013342531

PR #542, opened 2022-01-19
Settings from workspace folder

angelozerr:
Settings from workspace folder

This PR provides the capability to manage settings per workspace folder. It means that when you disable the validation with code action, it stores the disable setting in the settings.json of the project (and not in the workspace). It means that:

 * you can have several projects in a workspace and validation can be enabled / disabled for some project. The disabled is not global of the workspace.
 * As the disables setting is store in settings.json of the project, it's very portable (you can push the .vscode/settings.json in the git and share this settings with the team.


angelozerr:
This PR requires https://github.com/redhat-developer/vscode-quarkus/pull/443

fbricon:
Thanks @angelozerr

PR #547, opened 2022-01-31
Generate .html Qute templates by default

fbricon:
Generate .html Qute templates by default to be consistent with https://github.com/redhat-developer/vscode-quarkus/issues/450


angelozerr:
@fbricon could you update please the doc https://github.com/redhat-developer/vscode-quarkus/blob/master/docs/qute/TemplateSupport.md#qute-template-support to remove the sentence `To benefit with this support, we recommend your template files to follow the *.qute.*` and perhaps add a screenshot with notification which says that we enforce the html language  to qute html language please.

angelozerr:
Thanks @fbricon !

PR #562, opened 2022-02-17
Ignore undefined variable diagnostic for user tag

### LINKED ISSUE #561: Ignore undefined variable diagnostic for user tag
Given the `user.html` template  hosted in the src/main/resources/templates/tags folder (it's a user tag)

`{name}`

`name` is a parameter (other template can write `{#user name='ABCD' /}`, it means that:

 * name should be not reported as undefined variable warning.

angelozerr:
Ignore undefined variable diagnostic for user tag

Fixes #561

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
@@ -419,7 +419,7 @@ class QuteDiagnostics {
 
 			case Object: {
 				ObjectPart objectPart = (ObjectPart) current;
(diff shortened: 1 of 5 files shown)


qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
```diff
@@ -530,6 +530,14 @@ private ResolvedJavaTypeInfo validateObjectPart(ObjectPart objectPart, Section o
 				// The data model is not loaded, ignore the error of undefined variable
 				return null;
 			}
+			QuteProject project = template.getProject();
+			if (project != null) {
+				if (project.isUserTag(template)) {
```
fbricon, line 535: can be combined in a single if statement

PR #563, opened 2022-02-17
Added refactoring for changing with section to let

### LINKED ISSUE #489: CodeAction to replace deprecated `#with` into `#let` section
In the future, Qute will deprecate #with section. 

We should support diagnostic and quickfix fir that:

 * report a diagnostic in `#with` with LSP deprecated diagnostic tag.
 * provide a codeaction which transform #with with #let

Given  this Qute template:

```
{@org.acme.qute.Item item}
{#with item}
   {name}
   {price}
{/with}
```
#with should report a deprecated diagnostic and provide a code action to replace with:

```
{@org.acme.qute.Item item}
{#let name=item.name price=item.price}
   {name}
   {price}
{/let}
```

@mkouba what do you think about that?

AlexXuChen:
Added diagnostic for deprecated `#with` section and CodeAction to use `#let`.
replacewithwithlet2

Fixes #489 

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/commons/CodeActionFactory.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/commons/CodeActionFactory.java
@@ -106,6 +106,16 @@ public class CodeActionFactory {
 		return replace(title, Collections.singletonList(replace), document, diagnostic);
 	}
 
+	@SuppressWarnings("null")
+	public static CodeAction replace(String title, List<Range> ranges, String replaceText, TextDocumentItem document,
+			Diagnostic diagnostic) {
+		List<TextEdit> edits = null;
+		for (Range range : ranges) {
+			edits.add(new TextEdit(range, replaceText));
+		}
+		return replace(title, edits, document, diagnostic);
+	}
+
 	public static CodeAction replace(String title, List<TextEdit> replace, TextDocumentItem document,
 			Diagnostic diagnostic) {
 
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
@@ -20,6 +20,7 @@ import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -29,6 +30,7 @@ import org.eclipse.lsp4j.CodeActionContext;
 import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4j.TextEdit;
 
 import com.google.gson.JsonObject;
 import com.redhat.qute.ls.commons.BadLocationException;
@@ -40,8 +42,12 @@ import com.redhat.qute.parser.expression.ObjectPart;
 import com.redhat.qute.parser.template.Expression;
 import com.redhat.qute.parser.template.Node;
 import com.redhat.qute.parser.template.NodeKind;
+import com.redhat.qute.parser.template.ObjectPartASTVisitor;
+import com.redhat.qute.parser.template.Parameter;
 import com.redhat.qute.parser.template.Section;
+import com.redhat.qute.parser.template.SectionKind;
 import com.redhat.qute.parser.template.Template;
+import com.redhat.qute.parser.template.sections.WithSection;
 import com.redhat.qute.project.QuteProject;
 import com.redhat.qute.services.commands.QuteClientCommandConstants;
 import com.redhat.qute.services.diagnostics.QuteErrorCode;
@@ -71,6 +77,8 @@ class QuteCodeActions {
 	private static final String QUTE_VALIDATION_EXCLUDED_SECTION = "qute.validation.excluded";
 
 	private static final String EXCLUDED_VALIDATION_TITLE = "Exclude this file from validation.";
+
+	private static final String QUTE_DEPRICATED_WITH_SECTION = "Replace `#with` with `#let`.";
 
 	public CompletableFuture<List<CodeAction>> doCodeActions(Template template, CodeActionContext context, Range range,
 			SharedSettings sharedSettings) {
@@ -108,6 +116,15 @@ class QuteCodeActions {
 						//
 						// Create `undefinedTag`"
 						doCodeActionsForUndefinedSectionTag(template, diagnostic, codeActions);
+						break;
+					case NotRecommendedWithSection:
+						// The following Qute template:
+						// {#with }
+						//
+						// will provide a quickfix like:
+						//
+						// Replace `with` with `let`.
+						doCodeActionsForNotRecommendedWithSection(template, diagnostic, codeActions);
 						break;
 					default:
 						break;
@@ -184,6 +201,132 @@ class QuteCodeActions {
 		CodeAction disableValidationForTemplateQuickFix = createConfigurationUpdateCodeAction(title, templateUri,
 				QUTE_VALIDATION_EXCLUDED_SECTION, templateUri, ConfigurationItemEditType.add, diagnostics);
 		codeActions.add(disableValidationForTemplateQuickFix);
+	}
+
+	/**
+	 * Create CodeAction for unrecommended `with` Qute syntax.
+	 *
+	 * e.g. <code>
+	 * {#with item}
+	 *   {name}
+	 * {/with}
+	 * </code> becomes <code>
+	 * {#let name=item.name}
+	 *   {name}
+	 * {/let}
+	 * </code>
+	 *
+	 * @param template    the Qute template.
+	 * @param diagnostic  the diagnostic list that this CodeAction will fix.
+	 * @param codeActions the list of CodeActions to perform.
+	 *
+	 */
+	private static void doCodeActionsForNotRecommendedWithSection(Template template, Diagnostic diagnostic,
+			List<CodeAction> codeActions) {
+		Range withSectionRange = diagnostic.getRange();
+		try {
+			// 1. Retrieve the #with section node using diagnostic range
+			int withSectionStart = template.offsetAt(withSectionRange.getStart());
+			WithSection withSection = (WithSection) template.findNodeAt(withSectionStart + 1);
+
+			// 2. Initialize TextEdit array
+			List<TextEdit> edits = new ArrayList<TextEdit>();
+
+			// 2.1 Create text edit to update section start from #with to #let
+			// and collect all object parts from expression for text edit
+			// (e.g. {name} -> {#let name=item.name})
+			edits.add(createWithSectionOpenEdit(template, withSection));
+
+			// 2.2 Create text edit to update section end from /with to /let
+			edits.add(createWithSectionCloseEdit(template, withSection));
+
+			// 3. Create CodeAction
+			CodeAction replaceWithSection = CodeActionFactory.replace(QUTE_DEPRICATED_WITH_SECTION, edits,
+					template.getTextDocument(), diagnostic);
+			codeActions.add(replaceWithSection);
+		} catch (BadLocationException e) {
+			LOGGER.log(Level.SEVERE, "Creation of not recommended with section code action failed", e);
+		}
+	}
+
+	/**
+	 * Create text edit to replace unrecommended with section opening tag
+	 *
+	 * @param template    the Qute template.
+	 * @param withSection the Qute with section
+	 * @return
+	 * @throws BadLocationException
+	 */
+	private static TextEdit createWithSectionOpenEdit(Template template, WithSection withSection)
+			throws BadLocationException {
+		String objectPartName = withSection.getObjectParameter() != null ? withSection.getObjectParameter().getName()
+				: "";
+		// Use set to avoid duplicate parameters
+//		Set<String> withObjectParts = new HashSet<String>();
+
+		ObjectPartASTVisitor visitor = new ObjectPartASTVisitor();
+		withSection.accept(visitor);
+
+		// Retrieve all expressions in #with section
+//		findObjectParts(withSection, withObjectParts);
+		Set<String> withObjectParts = visitor.getObjectPartNames();
+
+		List<String> letObjectPartParameterList = new ArrayList<String>();
+		for (String objectPart : withObjectParts) {
+			letObjectPartParameterList.add(MessageFormat.format("{1}={0}.{1}", objectPartName, objectPart));
+		}
+
+		// Build text edit string
+		String letObjectPartParameters = String.join(" ", letObjectPartParameterList);
+		String withSectionOpenReplacementText = MessageFormat.format("#let {0}", letObjectPartParameters);
+		Range withSectionOpen = new Range(template.positionAt(withSection.getStartTagNameOpenOffset()),
+				template.positionAt(withSection.getStartTagCloseOffset()));
+
+		return new TextEdit(withSectionOpen, withSectionOpenReplacementText);
+	}
+
+	/**
+	 * Find all object parts by traversing AST Nodes, while retrieveing Expressions
+	 * nested in Sections with recursion
+	 *
+	 * @param node        current node in AST traversal
+	 * @param objectParts set of found object parts
+	 */
+	private static void findObjectParts(Node node, Set<String> objectParts) {
+		List<Node> children = node.getChildren();
+		// Base case: Node is an expression
+		if (children.isEmpty()) {
+			if (node.getKind() == NodeKind.Expression) {
+				objectParts.add(((Expression) node).getContent());
+			}
+		}
+		for (Node child : children) {
+			if (child.getKind() == NodeKind.Expression) {
+				objectParts.add(((Expression) child).getContent());
+			} else if (child.getKind() == NodeKind.Section && ((Section) child).getSectionKind() != SectionKind.WITH) {
+				for (Parameter param : ((Section) child).getParameters()) {
+					objectParts.add(param.getValue());
+				}
+				// Recursive call to traverse nested non-WithSection Sections
+				findObjectParts(child, objectParts);
+			}
+		}
+	}
+
+	/**
+	 * Create text edit to replace unrecommended with section closing tag
+	 *
+	 * @param template    the Qute template.
+	 * @param withSection the Qute with section
+	 * @return
+	 * @throws BadLocationException
+	 */
+	private static TextEdit createWithSectionCloseEdit(Template template, WithSection withSection)
+			throws BadLocationException {
+		String withSectionCloseReplacementText = "/let";
+		Range withSectionClose = new Range(template.positionAt(withSection.getEndTagNameOpenOffset()),
+				template.positionAt(withSection.getEndTagCloseOffset()));
+		return new TextEdit(withSectionClose, withSectionCloseReplacementText);
 	}
 
 	/**
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
@@ -54,6 +54,7 @@ import com.redhat.qute.parser.template.SectionKind;
 import com.redhat.qute.parser.template.Template;
 import com.redhat.qute.parser.template.sections.IncludeSection;
 import com.redhat.qute.parser.template.sections.LoopSection;
+import com.redhat.qute.parser.template.sections.WithSection;
 import com.redhat.qute.project.JavaMemberResult;
 import com.redhat.qute.project.QuteProject;
 import com.redhat.qute.project.datamodel.JavaDataModelCache;
@@ -248,6 +249,9 @@ class QuteDiagnostics {
 				case INCLUDE:
 					validateIncludeSection((IncludeSection) section, diagnostics);
 					break;
+				case WITH:
+					validateWithSection((WithSection) section, diagnostics);
+					break;
 				default:
 					validateSectionTag(section, template, resolvingJavaTypeContext, diagnostics);
 				}
@@ -344,6 +348,21 @@ class QuteDiagnostics {
 			// ex: {#include}
 			Range range = QutePositionUtility.selectStartTagName(includeSection);
 			Diagnostic diagnostic = createDiagnostic(range, DiagnosticSeverity.Error, QuteErrorCode.TemplateNotDefined);
+			diagnostics.add(diagnostic);
+		}
+	}
+
+	/**
+	 * Report that `#with` section is deprecated.
+	 *
+	 * @param withSection the with section
+	 * @param diagnostics the diagnostics to fill
+	 */
+	private static void validateWithSection(WithSection withSection, List<Diagnostic> diagnostics) {
+		if (withSection.getObjectParameter() != null) {
+			Range range = QutePositionUtility.createRange(withSection);
+			Diagnostic diagnostic = createDiagnostic(range, DiagnosticSeverity.Warning,
+					QuteErrorCode.NotRecommendedWithSection);
 			diagnostics.add(diagnostic);
 		}
 	}
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/diagnostics/DiagnosticDataFactory.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/diagnostics/DiagnosticDataFactory.java
@@ -24,7 +24,7 @@ import com.google.gson.JsonObject;
 
 /**
  * Diagnostic factory.
- * 
+ *
  * @author Angelo ZERR
  *
  */
@@ -55,4 +55,5 @@ public class DiagnosticDataFactory {
 				errorCode != null ? errorCode.getCode() : null);
 		return diagnostic;
 	}
+
 }
--- a/qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/diagnostics/QuteDiagnosticsInExpressionWithWithSectionTest.java
+++ b/qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/diagnostics/QuteDiagnosticsInExpressionWithWithSectionTest.java
@@ -30,27 +30,62 @@ import org.junit.jupiter.api.Test;
 public class QuteDiagnosticsInExpressionWithWithSectionTest {
 
 	@Test
+	public void missingObject() throws Exception {
+		String template = "{#with}\r\n" + //
+				"{/with}";
+		Diagnostic d = d(0, 6, 0, 6, QuteErrorCode.SyntaxError,
+				"Parser error on line 1: mandatory section parameters not declared for {#with}: [object]",
+				DiagnosticSeverity.Error);
+
+		testDiagnosticsFor(template, d);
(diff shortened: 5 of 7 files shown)


rgrunber:
As discussed, I think a good starting point would be to highlight the entire range as opposed to just the start element. It's likely that we won't need to set the deprecated tag.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
```diff
@@ -345,6 +349,25 @@ private static void validateIncludeSection(IncludeSection includeSection, List<D
...
+	 *
+	 * @param withSection the with section
+	 * @param diagnostics the diagnostics to fill
+	 */
+	private static void validateWithSection(WithSection withSection, List<Diagnostic> diagnostics) {
+		if (withSection != null) {
+//			List<DiagnosticTag> tags = Collections.singletonList(DiagnosticTag.Deprecated);
+//			Range range = QutePositionUtility.selectStartTagName(withSection);
+			Range range = QutePositionUtility.createRange(withSection);
+//			Diagnostic diagnostic = createDiagnosticWithTags(range, DiagnosticSeverity.Warning,
```
AlexXuChen, line 363: No longer marked as deprecated

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +197,41 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
...
+	 *
+	 */
+	private void doCodeActionsForDeprecatedWithSection(Template template, Diagnostic diagnostic,
+			List<CodeAction> codeActions) {
+		String title = MessageFormat.format(QUTE_DEPRICATED_WITH_SECTION, "with");
+		Range withSectionRange = diagnostic.getRange();
+		try {
+			int withSectionStart = template.getTextDocument().offsetAt(withSectionRange.getStart());
+			int withSectionEnd = template.getTextDocument().offsetAt(withSectionRange.getEnd());
+			String replacement = template.getText(withSectionStart, withSectionEnd).replace("#with", "#let")
```
rgrunber, line 216: I noticed that this would also replace any references to those strings that might occur in the body. I think if a comment or some html might be able to mention it that would be unintended.

Maybe `replaceFirst` for the start element and use `lastIndexOf` with `substring` to replace the ending element.

rgrunber (commented):
Overall, works well. Just a few minor fixes and one special case.

angelozerr:
In the demo, the code action doesn't generate properly the #let fix (yo can see warning on name and price.

The code action should generate:

```
{#let name=item.name price=item.price }
  ...
{\let}
```

After that, please write tests with diagnostic and code action.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +199,52 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
...
+			String javaResourceClassName = withSection.getObjectParameter().getName();
+			List<Node> withExpressions = withSection.getChildren().stream().filter(s -> s instanceof Expression)
+					.collect(Collectors.toList());
+			List<String> letExpressionParameterList = new ArrayList<String>();
+			for (Node expression : withExpressions) {
+				letExpressionParameterList.add(MessageFormat.format("{1}={0}.{1}", javaResourceClassName,
+						((Expression) expression).getContent()));
+			}
+			String letExpressionParameters = String.join(" ", letExpressionParameterList);
+			String withSectionText = template.getText(withSectionStart, withSectionEnd);
```
angelozerr, line 227: I think it should be cleaner to create a textedit per update. Don't replace the full text of #with but create several text edit per update:

 * #with -> #let
 * add parameters in the let
 * /with -> /let

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +199,52 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
...
+		Range withSectionRange = diagnostic.getRange();
+		try {
+			int withSectionStart = template.offsetAt(withSectionRange.getStart());
+			int withSectionEnd = template.offsetAt(withSectionRange.getEnd());
+			WithSection withSection = (WithSection) template.findNodeAt(withSectionStart);
+			String javaResourceClassName = withSection.getObjectParameter().getName();
+			List<Node> withExpressions = withSection.getChildren().stream().filter(s -> s instanceof Expression)
+					.collect(Collectors.toList());
+			List<String> letExpressionParameterList = new ArrayList<String>();
+			for (Node expression : withExpressions) {
```
angelozerr, line 222: rename expression with child, because the node can be a text, a section, an expression (not only an expression).

Calling just withSection.getChildren() is not enough because #with can include other section which can include other expresion. You need to call again node.#getChildren()

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
AlexXuChen: Now that I have access to the WithSection, this should be easier to do

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/diagnostics/QuteDiagnosticsInExpressionWithWithSectionTest.java
```diff
@@ -34,23 +34,28 @@ public void undefinedObject() throws Exception {
...
+	public void singleSection() throws Exception {
 		String template = "{@org.acme.Item item}\r\n" + //
 				"{#with item}\r\n" + //
 				"  <h1>{name}</h1>  \r\n" + //
 				"  <p>{price}</p> \r\n" + //
 				"{/with}";
-		testDiagnosticsFor(template);
+		Diagnostic d = d(1, 0, 4, 7, QuteErrorCode.NotRecommendedWithSection,
+				"`with` is not recommended. Use `let` instead.", DiagnosticSeverity.Warning);
+		testDiagnosticsFor(template, d);
```
angelozerr, line 58: Once you will manage code action with severaltext edit,please write test with your code action.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
AlexXuChen: replacewithwithletnested

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +199,79 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
...
+	private List<Node> findExpressions(Node node) {
+		List<Node> children = node.getChildren();
+		if (children.isEmpty()) {
+			if (node instanceof Expression) {
+				return Collections.singletonList(node);
+			}
+		}
+		List<Node> expressions = new ArrayList<Node>();
+		for (Node child : children) {
+			if (child instanceof Expression) {
```
angelozerr, line 257: Don't use instanceof but use node.getKind(), see https://github.com/redhat-developer/quarkus-ls/blob/9c48330144a6eef07205ab4366f868a5bd753498/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java#L192

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +199,79 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
...
+	 * @param diagnostic  the diagnostic list that this CodeAction will fix.
+	 * @param codeActions the list of CodeActions to perform.
+	 * @throws BadLocationException
+	 *
+	 */
+	private void doCodeActionsForNotRecommendedWithSection(Template template, Diagnostic diagnostic,
+			List<CodeAction> codeActions) {
+		Range withSectionRange = diagnostic.getRange();
+		try {
+			int withSectionStart = template.offsetAt(withSectionRange.getStart());
```
angelozerr, line 215: // Get the #with section node

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +199,79 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
...
+			WithSection withSection = (WithSection) template.findNodeAt(withSectionStart + 1);
+			String javaResourceClassName = withSection.getObjectParameter().getName();
```
angelozerr, line 217:
```java

// 2. Collect all object parts from expression (ex: {name}) or section parameter (ex : {#let itemName=name})
```

angelozerr: if your #with doesn't define parameter like item (ex : `{#with}`, you will have a NPE, in other words do:

```java
Parameter object = withSection.getObjectParameter();
if (object != null) {
	String javaResourceClassName = object.getName();
```

please write a test with this usecase and add test with code action and expected text edit.

angelozerr: Please rename `javaResourceClassName` with `objectPartName`

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +199,79 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
...
+			List<String> letExpressionParameterList = new ArrayList<String>();
+			for (Node expression : withExpressions) {
+				letExpressionParameterList.add(MessageFormat.format("{1}={0}.{1}", javaResourceClassName,
+						((Expression) expression).getContent()));
+			}
+			String letExpressionParameters = String.join(" ", letExpressionParameterList);
+
+			List<TextEdit> edits = new ArrayList<TextEdit>();
+
+			String withSectionOpenReplacementText = MessageFormat.format("#let {0}", letExpressionParameters);
```
angelozerr, line 228: I find it's difficult to read the code, because:

 * there are no comments which explain each text edit which are generated
 * code is a litlle mixed (ex : letExpressionParameters is computed, you declare List<TextEdit> edits,and after you create the text edit by using letExpressionParameters ). 

To fix that,I suggest that crate  static methods createWithSectionOpenEdit, createInsertLetParametersEdit, createWithSectionCloseEdit, your organize your code like this:

```java
// 1. Get the #with section node
WithSection withSection = (WithSection) template.findNodeAt(withSectionStart + 1);

// 2. Create required text edits to update #with to #let
List<TextEdit> edits = new ArrayList<TextEdit>();

// 2.1 Create text edit which update start section #with to #let
TextEdit withSectionOpenEdit= createWithSectionOpenEdit(withSection);
edits.add(withSectionOpenEdit);

// 2.2 Define all expression / parameter section in the #let section as parameters
TextEdit insertLetParametersEdit = createInsertLetParametersEdit(withSection);
if (insertLetParametersEdit != null) {
  edits.add(updateSectionStartEdit );
}

// 2.3 Create text edit which updates end section \with to \let
TextEdit withSectionCloseEdit = createWithSectionCloseEdit(withSection);
// Take care of null, because of #with is end with {\} you must not generate a text edit 
// Please write test with that
if (updateSectionEndEdit  != null) {
  edits.add(updateSectionStartEdit );
}

CodeAction replaceWithSection = ...
...
```

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +199,79 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
...
+		Range withSectionRange = diagnostic.getRange();
+		try {
+			int withSectionStart = template.offsetAt(withSectionRange.getStart());
+			WithSection withSection = (WithSection) template.findNodeAt(withSectionStart + 1);
+			String javaResourceClassName = withSection.getObjectParameter().getName();
+			List<Node> withExpressions = findExpressions(withSection);
+			List<String> letExpressionParameterList = new ArrayList<String>();
+			for (Node expression : withExpressions) {
+				letExpressionParameterList.add(MessageFormat.format("{1}={0}.{1}", javaResourceClassName,
+						((Expression) expression).getContent()));
```
angelozerr, line 222: Expression#getContent() returns the content of the expression (ex : `name.toString()` from `{name.toString()}`. More you get only expression, you must take care of too parameter (ex: `{#let foo=name}`.

In another words you need to catch object part from expression or parameter. To fix that, I suggest that you replace your 

`List<Node> withExpressions = findExpressions(withSection);`

with:

`List<String> objectPartNames = new ArrayList<>();
findObjectPart(withSection, objectPartNames);
`

and when you loop for children node, you get object part and collect the name in the given objectPartNames.

Please not set the array list objectPartNames in parameter method, give the capability to avoid creating an array list instance on each loop like https://github.com/redhat-developer/quarkus-ls/pull/563/files#diff-40c1df0b4206ee8c5be37ae62a3d2752b4eb564f080cd67fc2bfdb57728b6081R255

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/diagnostics/QuteDiagnosticsInExpressionWithWithSectionTest.java
AlexXuChen: Is this the appropriate code action: https://github.com/redhat-developer/quarkus-ls/pull/563/files#diff-f58ddd1bbb8c91147063f45a0973b0babf402ec1873f66899f5e89f4aa3d306bR93-R96 ?

angelozerr: perfect!

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +201,101 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
...
+			// 3. Create CodeAction
+			CodeAction replaceWithSection = CodeActionFactory.replace(QUTE_DEPRICATED_WITH_SECTION, edits,
+					template.getTextDocument(), diagnostic);
+			codeActions.add(replaceWithSection);
+		} catch (BadLocationException e) {
+			return;
+		}
+	}
+
+	private TextEdit createWithSectionOpenEdit(Template template, WithSection withSection) throws BadLocationException {
```
angelozerr, line 241: private static

angelozerr, line 241: Add javadoc

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +201,101 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
 		codeActions.add(disableValidationForTemplateQuickFix);
 	}
 
+	/**
+	 * Create CodeAction for deprecated `with` Qute syntax.
+	 *
```
angelozerr, line 206: Please add sample code (inside <code>) with #with and a second <code> section to know what does the code action.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +201,101 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
...
+
+			// 2.2 Create text edit to update section end from /with to /let
+			edits.add(createWithSectionCloseEdit(template, withSection));
+
+			// 3. Create CodeAction
+			CodeAction replaceWithSection = CodeActionFactory.replace(QUTE_DEPRICATED_WITH_SECTION, edits,
+					template.getTextDocument(), diagnostic);
+			codeActions.add(replaceWithSection);
+		} catch (BadLocationException e) {
+			return;
```
angelozerr, line 237: log the error.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +201,101 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
...
+	/**
+	 * Create CodeAction for deprecated `with` Qute syntax.
+	 *
+	 * @param template    the Qute template.
+	 * @param diagnostic  the diagnostic list that this CodeAction will fix.
+	 * @param codeActions the list of CodeActions to perform.
+	 * @throws BadLocationException
+	 *
+	 */
+	private void doCodeActionsForNotRecommendedWithSection(Template template, Diagnostic diagnostic,
```
angelozerr, line 213: private static

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +201,101 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
...
+			return;
+		}
+	}
+
+	private TextEdit createWithSectionOpenEdit(Template template, WithSection withSection) throws BadLocationException {
+		String objectPartName = withSection.getObjectParameter() != null ? withSection.getObjectParameter().getName()
+				: "";
+
+		// Use set to avoid duplicate parameters
+		HashSet<String> withExpressions = new HashSet<String>();
```
angelozerr, line 246: `Set<String> withExpressions`

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +201,101 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
...
+
+		// Retrieve all expressions in #with section
+		findExpressions(withSection, withExpressions);
```
angelozerr, line 249: Its not an expression that you collect, it's an object part. Please use objectPart (findObjectParts, )

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +201,101 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
...
+		List<String> letExpressionParameterList = new ArrayList<String>();
```
angelozerr, line 250: letObjectPartParameterList

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +201,101 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
...
+		for (String expression : withExpressions) {
```
angelozerr, line 251: for (String objectPart : withObjectParts) {

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +201,101 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
...
+		// Build text edit string
+		String letExpressionParameters = String.join(" ", letExpressionParameterList);
+		String withSectionOpenReplacementText = MessageFormat.format("#let {0}", letExpressionParameters);
+		Range withSectionOpen = new Range(template.positionAt(withSection.getStartTagNameOpenOffset()),
+				template.positionAt(withSection.getStartTagCloseOffset()));
+
+		return new TextEdit(withSectionOpen, withSectionOpenReplacementText);
+	}
+
+	private void findExpressions(Node node, HashSet<String> expressions) {
```
angelozerr, line 264: `Set<String>`

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -186,15 +201,101 @@ private static void doCodeActionToDisableValidation(Template template, List<Diag
...
+		List<Node> children = node.getChildren();
+		// Base case: Node is an expression
+		if (children.isEmpty()) {
+			if (node.getKind() == NodeKind.Expression) {
+				expressions.add(((Expression) node).getContent());
```
angelozerr, line 269: In case expression, you need to loop for each node of expressions to extract ExpressionPart (specially an Object Part).

I would use swithc(node.getKind()) to visit each nodes.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
```diff
@@ -345,6 +349,24 @@ private static void validateIncludeSection(IncludeSection includeSection, List<D
...
 
+	/**
+	 * Report that `#with` section is deprecated.
+	 *
+	 * @param withSection the with section
+	 * @param diagnostics the diagnostics to fill
+	 */
+	private static void validateWithSection(WithSection withSection, List<Diagnostic> diagnostics) {
+		if (withSection.getObjectParameter() != null) {
+//			List<DiagnosticTag> tags = Collections.singletonList(DiagnosticTag.Deprecated);
```
angelozerr, line 360: remove comments

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
```diff
@@ -391,7 +413,7 @@ private ResolvedJavaTypeInfo validateExpressionParts(Parts parts, Section ownerS
 		ResolvedJavaTypeInfo resolvedJavaType = null;
 		String namespace = null;
 		for (int i = 0; i < parts.getChildCount(); i++) {
-			Part current = ((Part) parts.getChild(i));
+			Part current = (parts.getChild(i));
```
angelozerr, line 416: `Part current = parts.getChild(i);`

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/diagnostics/DiagnosticDataFactory.java
```diff
@@ -55,4 +58,13 @@ public static Diagnostic createDiagnostic(Range range, String message, Diagnosti
 				errorCode != null ? errorCode.getCode() : null);
 		return diagnostic;
 	}
+
+	public static Diagnostic createDiagnosticWithTags(Range range, DiagnosticSeverity severity,
```
angelozerr, line 62: remove this unused method.

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/diagnostics/QuteDiagnosticsInExpressionWithWithSectionTest.java
angelozerr: Add a test with #with without object parameter.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
angelozerr: Ypu need to takecar of Parameter of sectionwhich can have object part too.

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/diagnostics/QuteDiagnosticsInExpressionWithWithSectionTest.java
AlexXuChen: If there isn't an object parameter, there is simply just a syntax error correct?

angelozerr: There is no code action.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
AlexXuChen: Is an ExpressionPart multiple Nodes? I was under the impression it is just one, i.e. `{name}, {price}, etc..`

angelozerr: See sample at https://github.com/redhat-developer/quarkus-ls/blob/9c48330144a6eef07205ab4366f868a5bd753498/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java#L370

AlexXuChen: I believe the Parameter use case should be handled now:
replacewithwithletnestedparam
I decided against a switch case however since we don't handle every SectionKind case for this code action

angelozerr: @AlexXuChen I suggest that you rebase your PR from master to benefit with ASTVisitor. See https://github.com/redhat-developer/quarkus-ls/blob/b247849e376493117f3ab704d9f3060003ce86f0/qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/parser/template/ASTVisitorTest.java#L59

I suggest that you copy/paste CollectObjectPartNameASTVisitor  (from test) in your code action class.

It will simplify the code and will cover other usecase that you don't take care of like `{item.name}`. In this case you must collect item and not item.name. The ASTVisitor takes care of that.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/commons/CodeActionFactory.java
```diff
@@ -106,6 +106,16 @@ public static CodeAction replace(String title, Range range, String replaceText,
 		return replace(title, Collections.singletonList(replace), document, diagnostic);
 	}
 
+	@SuppressWarnings("null")
+	public static CodeAction replace(String title, List<Range> ranges, String replaceText, TextDocumentItem document,
```
angelozerr, line 110: Add a javadoc

fbricon:
- as per https://github.com/redhat-developer/quarkus-ls/issues/489#issuecomment-1046669808, `#with` is not deprecated, so let's avoid that vocabulary in the code. 
- The warning should only appear when used in a type safe context.
- There should be a preference to disable the warning.

fbricon:
the code action should appear as a refactoring item

PR #564, opened 2022-02-18
Support parameter expression in user tag

angelozerr:
Support parameter expression in user tag


angelozerr:
This PR provides parameter support for user tag, here a little demo:
 
UserTagParameterDemo

PR #565, opened 2022-02-18
Support for untitled uri.

angelozerr:
Support for untitled uri.


angelozerr:
This PR support another URI kind than `file://` (ex : `untitled://`, jdt://jarentry)

Ex: with vscode code, if you do Ctrl+N and you force the language to Qute HTML, in master you have so many errors in the trace because master consider that URI is a file uri. This PR takes care of this usecase.

PR #578, opened 2022-02-28
Support for Injecting Beans Directly In Templates

### LINKED ISSUE #546: Support for Injecting Beans Directly In Templates
See https://quarkus.io/guides/qute-reference#injecting-beans-directly-in-templates

See sample https://github.com/FroMage/quarkus-renarde-todo/blob/5384be6c070d11e5aff200f2469d969c100af088/src/main/resources/templates/tags/input.html#L7

Today we have a warning:

image

angelozerr:
Support for Injecting Beans Directly In Templates

Fixes #546


angelozerr:
Here a demo with inject namespace:

InjectQuteDemo

angelozerr:
@fbricon the PR is ready to review it.

qute.jdt/com.redhat.qute.jdt.test/src/main/java/com/redhat/qute/jdt/template/TemplateGetDataModelProjectTest.java
```diff
@@ -212,24 +215,64 @@ private static void testValueResolvers(DataModelProject<DataModelTemplate<DataMo
...
+				"bar", resolvers);
+
+		// @Named("user")
+		// private String getUser() {...
+		assertValueResolver("inject", "getUser() : java.lang.String", "org.acme.qute.InjectedData", //
+				"user", resolvers);
+
+		// @Named
+		// private String getSystemUser() {...
+		assertValueResolver("inject", "getSystemUser() : java.lang.String", "org.acme.qute.InjectedData", //
```
fbricon, line 254: I don't think that's a valid use case. CDI would expose a bean named getSystemUser or maybe systemUser (but I'm not that familiar with CDI).
I suggest you try running an actual Quarkus app with this @Named beans to check whether it makes sense or not.

fbricon:
I feel like we show too much crap with:
Screenshot 2022-03-04 at 16 43 01

Useful variables and properties are at the bottom of the list, you need to scroll down. Can we lower the `inject:` items priority?

angelozerr:
> Bunch of NPEs I've seen today:

It should be fixed with test.

angelozerr:
> Useful variables and properties are at the bottom of the list, you need to scroll down. Can we lower the inject: items priority?

It should b efixed

PR #581, opened 2022-03-03
Changelog for 0.11.0

rgrunber:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,41 @@
 # Quarkus extension for MicroProfile Language Server Changelog
 
+## 0.11.0 (March 24, 2022)
+
+### Enhancements
+
+ * enhancement - Create a Qute Language Server. See [#176](https://github.com/redhat-developer/quarkus-ls/issues/176).
+ * enhancement - Support for method parameters in Qute templates. See [#486](https://github.com/redhat-developer/quarkus-ls/issues/486).
+ * enhancement - Support Java type hover for Qute templates. See [#502](https://github.com/redhat-developer/quarkus-ls/issues/502).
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,40 @@
 # Quarkus extension for MicroProfile Language Server Changelog
 
+## 0.11.0 (March 17, 2022)
+
+### Enhancements
+
+ * enhancement - Create a Qute Language Server. See [#176](https://github.com/redhat-developer/quarkus-ls/issues/176).
+ * enhancement - Support for method parameters in Qute templates. See [#486](https://github.com/redhat-developer/quarkus-ls/issues/486).
+ * enhancement - Support Java type hover for Qute templates. See [#502](https://github.com/redhat-developer/quarkus-ls/issues/502).
+ * enhancement - Completion for available section after `{` and `{#`. See [#538](https://github.com/redhat-developer/quarkus-ls/issues/538).
```
AlexXuChen, line 10: There actually isn't support for `{` completion anymore, just `{#`. We should probably fix this in the changelog an in the issue title to avoid confusion.

AlexXuChen (changes requested):

CHANGELOG.md
angelozerr: Indeed we change this annoying behavior (because data model was displayed after section in completion). So the section completion are available: 

 * by typing the section name (ex : each).
 * after  {#|

angelozerr:
@rgrunber please add https://github.com/redhat-developer/quarkus-ls/pull/611

CHANGELOG.md
rgrunber: Fixed. updated the title.

PR #586, opened 2022-03-08
Support for uri/uriabs from Quarkus Renarde

### LINKED ISSUE #571: Support for uri/uriabs from Quarkus Renarde
See https://github.com/quarkiverse/quarkus-renarde/blob/main/docs/modules/ROOT/pages/index.adoc#obtaining-a-uri-in-qute-views

angelozerr:
Support for uri/uriabs from Quarkus Renarde

Fixes #571

This PR uses those renarde validation rules:
 * all parameters which are annotated with @RestPath (or @PathParam) are considered as required
 * all parameters which are annotated with @RestQuery (or @QueryParam) are optional
 * all parameters which are annotated with @RestForm (or @FormParam) cannot appear in the method of uri.


angelozerr:
Here a little demo with renarde uri:

RenardeDemo

//c @FroMage

FroMage:
> @mkouba why ReflectionValueResolver forbids the void although renarde value resolver doesn't forbid it? If we want to keep this behavior,it means that we must provide custom validation for renarde.

Actually this doesn't go through `ReflectionValueResolver`, because custom namespaces can do whatever they want.

> How to manage this rule with generic mean?

Well, this is because the method only requires parameters that end up in the URI, such as `@PathParam`, but all the others can be ignored when calling `uri` (those that are context, or headers, cookies, or form params in this case).

So yeah, namespace resolvers are really like primitives and must be hard-coded: you can't guess what they do.

mkouba:
> > @mkouba why ReflectionValueResolver forbids the void although renarde value resolver doesn't forbid it? If we want to keep this behavior,it means that we must provide custom validation for renarde.
> 
> Actually this doesn't go through `ReflectionValueResolver`, because custom namespaces can do whatever they want.

Stef is correct that we only validate output expressions where a namespace is mapped to a template extension method or `@TemplateData` with a namespace. In this case, I assume that `uri:RenardeSecurityController.loginUsingOidc` does not invoke the controller method at all.

FroMage:
Indeed. It's only used to invoke the reverse-router.

angelozerr:
Thanks @FroMage @mkouba for our feedback.

It wil require to support custom validation for a given namespace.

I will do that in an another issues. For the moment we will have those errors.

mkouba:
> It wil require to support custom validation for a given namespace.

In general, an output expression that starts with a namespace should only be validated (java model) if the namespace is `inject`, `cdi` or corresponds to a [`@TemplateEnum`](https://quarkus.io/guides/qute-reference#convenient-annotation-for-enums), [@TemplateData#namespace()](https://quarkus.io/guides/qute-reference#accessing-static-fields-and-methods) and a [namespace extension method](https://quarkus.io/guides/qute-reference#namespace_extension_methods). All other namespaces should be ignored.

angelozerr:
> All other namespaces should be ignored.

I understand the idea, but my idea is to provide the capability to provide custom validation like renarde validation.

fbricon:
So if I'm using the `uri:` or `uriabs:` namespace on a non-renarde project (renarde jars not on the project's classpath), the tool behaves as if the resolver was known and available, as opposed to actually complaining about an unknown namespace.

Screenshot 2022-03-09 at 12 14 21

angelozerr:
> So if I'm using the uri: or uriabs: namespace on a non-renarde project (renarde jars not on the project's classpath), the tool behaves as if the resolver was known and available, as opposed to actually complaining about an unknown namespace.

Good catch! It should be fixed now.

fbricon:
In https://github.com/FroMage/quarkus-renarde-todo/blob/main/src/main/resources/templates/Todos/index.html#L24, the target Controller is resolved:

Screenshot 2022-03-09 at 18 54 21

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/extensions/renarde/UriNamespaceResolverSupport.java
```diff
@@ -0,0 +1,131 @@
...
+
+	@Override
+	public void beginSearch(SearchContext context, IProgressMonitor monitor) {
+		IJavaProject javaProject = context.getJavaProject();
+		IType type = JDTTypeUtils.findType(javaProject, RENARDE_CONTROLLER_TYPE);
+		if (type != null) {
+			try {
+				// Find all classes which extends 'io.quarkiverse.renarde.Controller'
+				ITypeHierarchy typeHierarchy = type.newTypeHierarchy(monitor);
+				IType[] controllerTypes = typeHierarchy.getAllClasses();
```
angelozerr, line 59: @rgrunber here I need to collect all classes which extends "io.quarkiverse.renarde.Controller" class. I have a usecase with : 

 * Application which extends Controller
 * Todos which extends ControllerWithUser which extends Controller.

When Controller and ControllerWithUser comes from sources (I copied/pasted thoses sources in my project), I can get Application and Todos.

But when Controller and ControllerWithUser comes from JAR,it seems (according the problem with @fbricon with Todos), that it get only Application (notTodos which extends ControllerWithUser which extends Controller).

At first do you think my code is correct? Have you some idea about this problem?

rgrunber: Is it the same result if you try using `IType[] getAllSubtypes(IType type);` ?

rgrunber: The first thing I notice from running the test case is that 

```
/qute-renarde/lib/quarkus-renarde-1.0.0-SNAPSHOT.jar[CPE_LIBRARY][K_BINARY][isExported:false][attributes:module=true]
/qute-renarde/lib/quarkus-renarde-oidc-1.0.0-SNAPSHOT.jar[CPE_LIBRARY][K_BINARY][isExported:false][attributes:module=true]
```

don't exist. They're defined in the `.classpath` so they show, but there's no `lib` folder in the qute-renard project so no way to even get the type.

qute.jdt/com.redhat.qute.jdt.test/projects/maven/qute-renarde/.classpath
```diff
@@ -0,0 +1,49 @@
...
+			<attribute name="maven.pomderived" value="true"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry excluding="**" kind="src" output="target/test-classes" path="src/test/resources">
+		<attributes>
+			<attribute name="test" value="true"/>
+			<attribute name="maven.pomderived" value="true"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11">
```
rgrunber, line 27: @angelozerr , I made the following change, and type hierarchy worked for me.

```suggestion
	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8">

```

```
typeHierarchy.getAllSubtypes(type)

Application (not open) {key=Lrest/Application;} [in Application.java [in rest [in src/main/java [in qute-renarde]]]]
ControllerWithUser (not open) {key=Lio/quarkiverse/renarde/oidc/ControllerWithUser<TU;>;} [in ControllerWithUser.class [in io.quarkiverse.renarde.oidc [in lib/quarkus-renarde-oidc-1.0.0-SNAPSHOT.jar [in qute-renarde]]]]
Login (not open) {key=Lrest/Login;} [in Login.java [in rest [in src/main/java [in qute-renarde]]]]
Todos (not open) {key=Lrest/Todos;} [in Todos.java [in rest [in src/main/java [in qute-renarde]]]]
RenardeSecurityController (not open) {key=Lio/quarkiverse/renarde/oidc/impl/RenardeSecurityController;} [in RenardeSecurityController.class [in io.quarkiverse.renarde.oidc.impl [in lib/quarkus-renarde-oidc-1.0.0-SNAPSHOT.jar [in qute-renarde]]]]
```

rgrunber: @fbricon , Occasionally, I see the following errors when I import qute-renarde into an Eclipse workspace :

```
Description	Resource	Path	Location	Type
The project was not built since its build path is incomplete. Cannot find the class file for javax.ws.rs.core.NewCookie. Fix the build path then try building this project	qute-renarde		Unknown	Java Problem
The type javax.ws.rs.core.NewCookie cannot be resolved. It is indirectly referenced from required .class files	Login.java	/qute-renarde/src/main/java/rest	line 1	Java Problem
```

The  above change resolves it. Could this be an Eclipse-specific (maybe m2e-specific) issue that's causing the failure to resolve ?

angelozerr:
@fbricon @FroMage @maxandersen event this PR is not perfect it supports uri and uriabs for renarde. I need to rebase it but what do you think to integrate it master ?

@maxandersen see the demo at https://github.com/redhat-developer/quarkus-ls/pull/586#issuecomment-1061890339

FroMage:
Can you remind me what the problems are with this PR?

fbricon:
@angelozerr tested on https://github.com/ia3andy/renotes, Ctrl+Click on [{uri:Notes.notes} ](https://github.com/ia3andy/renotes/blob/main/src/main/resources/templates/main.html#L23) doesn't work (file not found), and [notes.html](https://github.com/ia3andy/renotes/blob/main/src/main/resources/templates/Notes/notes.html)  and [noteSearch.html](https://github.com/ia3andy/renotes/blob/main/src/main/resources/templates/tags/noteSearch.html) show several errors:
Screenshot 2023-01-25 at 19 32 19

angelozerr:
> , Ctrl+Click on [{uri:Notes.notes} ](https://github.com/ia3andy/renotes/blob/main/src/main/resources/templates/main.html#L23)doesn't work (file not found),

Indeed this problem comes from html ls which generate a document link for href attribute. This behavior comes from html ls. If you use F12 insteaf of ctrl+click it should work

For the other errors Im aware about that because renarde has à validation différent from qute. Ex: qute forbid voir method altjough renarde allow it.

I need to take care of those renarde rôles.

angelozerr:
@fbricon @datho7561 this PR can now be reviewed.

After discussing with @FroMage here the validation rules:

 * all parameters which are annoated with @RestPath (or @PathParam) are considered as required
 * all parameters which are annotaed with @RestQuery (or @QueryParam) are optional
 * all parameters which are annoated with @RestForm (or @FormParam) cannot appear in the method of uri.

Here a de demo with uri/uriabs renarde support:

RenardeUriDemo

datho7561:
> all parameters which are annoated with @RestForm (or @FormParam) cannot appear in the method of uri.

I am a little confused by this. Does this mean that the method should not be able to be invoked from Qute, or does this mean that you skip over the parameter when calling the function from Qute, or something else?

FroMage:
The method isn't invoked from Qute: the `uri:` namespace produces a URI to a method, it doesn't invoke it. And for a URI, only the parameters that take place in building the URI are required. The form parameters don't end up in the URI but in the HTTP body. That's why they're not useful for the `uri:` tag.

datho7561:
> only the parameters that take place in building the URI are required. The form parameters don't end up in the URI but in the HTTP body. That's why they're not useful for the `uri:` tag.

Okay. That makes sense. So if I have:

```java
@GET
public Response foo(String myParam, @PathParam(value = "asdf") String myParam2) {
	System.out.println(myParam2);
	return Response.ok().build();
}
```

Do I need to "pass" both the parameters, or do I only "pass" `myParam2` to build the URI?

FroMage:
Only `myParam2` because `myParam` in this case should be the http request body (which is empty in the case of `GET`)…

… _unless_ you have a `@Path("{myParam}")` on the method or class, which makes it implicitely a path param ;)

datho7561:
Okay makes sense. I did a bit of testing, and it looks like you can do

```html
<a href={uri:Login.foo(null, "1234")}>click me</a>
```

but not
```html
<a href={uri:Login.foo("1234")}>click me</a>
```

@angelozerr do you provide support for validating the number of methods? Currently, if I have the second one in the template, it doesn't list it as invalid, but the template fails to render

FroMage:
Ah, I should check that…

qute.jdt/com.redhat.qute.jdt.test/src/main/java/com/redhat/qute/jdt/QuteAssert.java
```diff
@@ -0,0 +1,73 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
```
datho7561, line 2: update the copyright headers

angelozerr:
> [angelozerr](https://github.com/angelozerr) do you provide support for validating the number of methods?

I suppose you speak about numbers of parameters. If it that yes I tried to support it.

But I supposed that RestParam annotation are set at first. It is not your usecase. I did that because renarde sample uses that.

I think we could do those advanced usecases in a separate PR.

datho7561:
> I think we could do those advanced usecases in a separate PR.

Okay. Makes sense. Be sure to leave a comment or open an issue for that.

datho7561 (changes requested):
Sorry the review took so long, I was trying to wrap my head around what features were added in the PR and learn Renarde.

datho7561:
If I have a method, and I "pass" it more parameters than it has, then there is no error. I don't know if this is intentional

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/extensions/renarde/UriNamespaceResolverSupport.java
```diff
@@ -0,0 +1,150 @@
...
+		IJavaProject javaProject = context.getJavaProject();
+		IType type = JDTTypeUtils.findType(javaProject, RENARDE_CONTROLLER_TYPE);
+		if (type != null) {
+			try {
+				// Find all classes which extends 'io.quarkiverse.renarde.Controller'
+				collectRenardeController(type, context, monitor);
+				// Find all classes which extends
+				// 'io.quarkiverse.renarde.oidc.ControllerWithUser'
+				collectRenardeController(JDTTypeUtils.findType(javaProject, RENARDE_CONTROLLER_WITH_USER_TYPE), context,
+						monitor);
```
rgrunber, line 64: Can you play around with the following snippet ? It seems to fix the test for me :


```diff
diff --git a/qute.jdt/com.redhat.qute.jdt.test/src/main/java/com/redhat/qute/jdt/QuteProjectTest.java b/qute.jdt/com.redhat.qute.jdt.test/src/main/java/com/redhat/qute/jdt/QuteProjectTest.java
index 06a44eb..3c5a2de 100644
--- a/qute.jdt/com.redhat.qute.jdt.test/src/main/java/com/redhat/qute/jdt/QuteProjectTest.java
+++ b/qute.jdt/com.redhat.qute.jdt.test/src/main/java/com/redhat/qute/jdt/QuteProjectTest.java
@@ -120,7 +120,8 @@ public class QuteProjectTest {
 			};
 			IProgressMonitor monitor = new NullProgressMonitor();
 			JavaCore.run(runnable, null, monitor);
-			waitForBackgroundJobs(monitor);			
+			waitForBackgroundJobs(monitor);
+			org.eclipse.jdt.ls.core.internal.JobHelpers.waitUntilIndexesReady();
 		}
 		// Collect Quarkus properties from the "hibernate-orm-resteasy" project. It
 		// should collect Quarkus properties from given JAR:
diff --git a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/extensions/renarde/UriNamespaceResolverSupport.java b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/extensions/renarde/UriNamespaceResolverSupport.java
index 437b680..907fa92 100644
--- a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/extensions/renarde/UriNamespaceResolverSupport.java
+++ b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/extensions/renarde/UriNamespaceResolverSupport.java
@@ -58,10 +58,6 @@ public class UriNamespaceResolverSupport extends AbstractDataModelProvider {
 			try {
 				// Find all classes which extends 'io.quarkiverse.renarde.Controller'
 				collectRenardeController(type, context, monitor);
-				// Find all classes which extends
-				// 'io.quarkiverse.renarde.oidc.ControllerWithUser'
-				collectRenardeController(JDTTypeUtils.findType(javaProject, RENARDE_CONTROLLER_WITH_USER_TYPE), context,
-						monitor);
 			} catch (JavaModelException e) {
 				LOGGER.log(Level.SEVERE, "Error while collecting Renarde Controller.", e);
 			}
```

I'm not sure if I put it in the best place, but the idea is once a project has successfully imported, the index could still be building in the background, so before creating any queries, we should make sure the index loaded completely.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/jaxrs/JaxRsMethodKind.java
```diff
@@ -0,0 +1,42 @@
...
+package com.redhat.qute.commons.jaxrs;
+
+/**
+ * JaxRs method kind.
+ *
+ * @author Angelo ZERR
+ */
+public enum JaxRsMethodKind {
+
+	NONE(0), //
```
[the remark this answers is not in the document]
angelozerr: NONE is now removed.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/extensions/renarde/RenardeResolvedJavaTypeFactory.java
```diff
@@ -0,0 +1,159 @@
...
+
+	@Override
+	protected JavaMethodInfo createMethod(IMethod method, ITypeResolver typeResolver) {
+		JavaMethodInfo info = super.createMethod(method, typeResolver);
+		collectJaxrsInfo(method, info);
+		return info;
+	}
+
+	private static void collectJaxrsInfo(IMethod method, JavaMethodInfo info) {
+		JaxRsMethodKind methodKind = JaxRsMethodKind.GET;
```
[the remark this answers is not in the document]
angelozerr: The main goal is to mark the method as renarde method controllerto manage custom validation on Qute LS. To avoid having a flag JavaMethodInfo#isRenarde(), I use JaxRsMethodKind because perhaps in the future we could have some specific rule for @POST, @GET (ex: you cannot use the @POST method outside an HTML form)

I wanted just to distinguish POST and GET, but other kind are not supported because we don't need it for the moment.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/extensions/renarde/UriNamespaceResolverSupport.java
angelozerr: Thanks so much @rgrunber for your great suggestion. It works like a charm now. No need to use some hack. Thanks!

qute.jdt/com.redhat.qute.jdt.test/src/main/java/com/redhat/qute/jdt/QuteAssert.java
angelozerr: fixed

angelozerr:
> Okay. Makes sense. Be sure to leave a comment or open an issue for that.

See https://github.com/redhat-developer/quarkus-ls/issues/780

angelozerr:
> @angelozerr do you want to handle this in this PR or should I open an issue for it?

I think it is a bug with space. If you write

`<a href={uri:Login.foo(null,"1234")}>click me</a>`

it should work, no?

I would like to do that in a separate PR since it is not linked to this PR

angelozerr:
@datho7561 I need to write JDT test to collect ResolvedJavaTypeInfo for renarde.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/extensions/renarde/RenardeResolvedJavaTypeFactory.java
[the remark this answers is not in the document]
datho7561: Okay. That makes sense to me. In the future it would be interesting to support `@DELETE` and `@PUT` if we want to provide integration with HTMX like in the [renotes sample project](https://github.com/ia3andy/renotes), but for now I am okay with just distinguishing `@POST` from the others.

datho7561:
> I think it is a bug with space.
You're right, I can confirm it works if I don't add spaces.

The qute-ls server is not shutting down properly, but I think that is also not related to this PR.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/extensions/renarde/RenardeResolvedJavaTypeFactory.java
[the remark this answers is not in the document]
angelozerr: fixed

angelozerr:
> @datho7561 I need to write JDT test to collect ResolvedJavaTypeInfo for renarde.

done quickly

angelozerr:
> The qute-ls server is not shutting down properly, but I think that is also not related to this PR.

No, but it is not related with new vscode language client?

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/extensions/renarde/RenardeResolvedJavaTypeFactory.java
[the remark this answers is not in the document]
datho7561: Sorry, I should have been more clear. I meant in the `enum` class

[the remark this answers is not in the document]
angelozerr: ok, fixed

datho7561:
I did some testing, and it looks like you were right: it's related to the new language client.

angelozerr:
> I did some testing, and it looks like you were right: it's related to the new language client.

I removed https://github.com/redhat-developer/vscode-quarkus/pull/506/files#diff-c1b7030f8c5f91b6d8690295e037104c11aecfc6d7eb7a9bd531aa0a9c5547fdL40 because I though this language client fix the problem. If you add again shouldLanguageServerExitOnShutdown: true is shutdown working better?

If it works better,it mean that we have the same problem with vscode-microprofile and vscode-xml -(

angelozerr:
@datho7561 tests pass now, if you think it is good, please merge it.

datho7561 (approved):
Looks good. Thanks, Angelo!

angelozerr:
Thanks so much @datho7561 for your great review!

@FroMage to have been patient and give us great feedback. This PR is à good start for supporting renarde but we need to improve again the support.

@FroMage when you will have time please install vscode quarkus prerelease when it will be available. I think tomorrow prerelease should be available.

Please install too vscode microprofile prerelease to avoid having problem with quarkus support

PR #588, opened 2022-03-08
Support parameter value as String

angelozerr:
Support parameter value as String


angelozerr:
This PR fixes wrong diagnostic with parameter value as String:

Given this Qute template:

`{#formElement name="userName" label="User Name"}
`

Name is higlighted as a warning although it is inside a String:

image

This PR fixes this problem.

fbricon:
thx @angelozerr

PR #593, opened 2022-03-11
Support for `textDocument/InlayHint`

### LINKED ISSUE #595: Support for `textDocument/InlayHint`
Display the Java type of parameter name as inlay hint.

See inlay hints rendererd with gray background:

image

angelozerr:
Support for `textDocument/InlayHint`

Fixes #595

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/lsp4j/proposed/InlayHint.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/lsp4j/proposed/InlayHint.java
@@ -1,0 +1,110 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.lsp4j.proposed;
+
+import java.util.List;
+
+import org.eclipse.lsp4j.Position;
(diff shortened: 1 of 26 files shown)


angelozerr:
This PR requires https://github.com/redhat-developer/vscode-quarkus/pull/474

angelozerr:
Here the result:

image

It display Java type before the alias.

It's just a POC and it doesn't use the real InlayHint of LSP4J but uses the structure like described in the spec.

The inlay hint doesn't appear when the Java type is not resolved. So you must type something in the text editor to refresh the inlay hint.

rgrunber:
Nice :)

Only 2 things immediately stand out.

1. We should think about the best place to put the hint. `Todo todo` or `todo Todo`, or maybe `todo : Todo`, or `todo (Todo)`. Basically it might cause some confusion if it just appears there. 
2. I would play around with typing out the full line and confirm it isn't annoying if the hint appears as someone is typing. Somtimes the logic can resolve a hint at various points that causes the cursor to jump.

angelozerr:
@rgrunber here the new renderer:

image

fbricon:
please don't use `package org.eclipse.lsp4j.proposed;` this is not an eclipse project nor is it code copied from an eclipse project.

angelozerr:
> please don't use package org.eclipse.lsp4j.proposed; this is not an eclipse project nor is it code copied from an eclipse project.

rename with `com.redhat.lsp4j.proposed`. Once LSP4J will provide Inlay Hint, we will able to remove those classes.

angelozerr:
Here a new demo with inlay hint:

InlayHintDemo2

Now I support only section parameter (and method parameter type has been removed, we will see in an another PR the method parameter name support).

This PR requires https://github.com/redhat-developer/vscode-quarkus/pull/474 which provide settings to enable/disable inlay hints + provide inlay hint TypeScript by waiting for the release of vscode-languageclient which will provide this feature.

AlexXuChen (approved):
LGTM, working as intended

PR #596, opened 2022-03-13
Support Java expression for it user tag

angelozerr:
Support Java expression for it user tag

This PR provides completion, validation hover for expression inside user tag which have no parameter name (it):

* `{#user item.| }` should show properties, methods of the item for completion,.
* `{#user item.nam|e }` should show information about the property name
* `{#user item.XXXX }` should report an error to the XXXX

Here a demo:

ItExpressionDemo


qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/diagnostics/QuteDiagnosticsUserTagTest.java
```diff
@@ -60,6 +60,15 @@ public void undefinedVariableInUserTagCall() {
 		testDiagnosticsFor(template, d);
 	}
 
+	@Test
+	public void undefinedVariableInItUserTagCall() {
```
fbricon, line 64: undefinedObject* ?

angelozerr: Good catch! I replace all "variable"with "object"

fbricon:
thx @angelozerr

PR #598, opened 2022-03-14
Qute-LS crashes if settings.json is configured improperly

### LINKED ISSUE #591: Qute-LS crashes if `settings.json` is configured improperly
Similar to https://github.com/eclipse/lemminx/issues/964, if a setting in `settings.json` is improperly configured/malformed, the language server will crash during the client request. There should be a fail-safe way of handling this issue.

angelozerr:
Qute-LS crashes if settings.json is configured improperly

Fixes #591

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/settings/FaultTolerantTypeAdapterFactory.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/settings/FaultTolerantTypeAdapterFactory.java
@@ -1,0 +1,69 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.qute.settings;
+
+import java.io.IOException;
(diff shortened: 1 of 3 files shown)


qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/settings/FaultTolerantTypeAdapterFactory.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+
+/**
+ * Creates TypeAdapters that are wrapped in a try/catch statement, so if
+ * anything goes wrong when deserializing an object, it is set to null and not
+ * parsed.
+ *
+ */
+public class FaultTolerantTypeAdapterFactory implements TypeAdapterFactory {
```
AlexXuChen, line 28: Do we still want the name to be "FaultTolerant"? I know we use this exact same class in lemminx, but fault tolerance is related to mp correct?

AlexXuChen (commented):
Feature is working as intended, language server isn't crashing

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/settings/FaultTolerantTypeAdapterFactory.java
angelozerr: No its not linked to lemminx or mp. It means that the load is fault tolerant.

PR #599, opened 2022-03-15
Rename code action to "Ignore `UndefinedObject` problem."

angelozerr:
Rename code action to "Ignore `UndefinedObject` problem."


fbricon:
thanks @angelozerr

PR #607, opened 2022-03-17
Support for Infix notation

### LINKED ISSUE #519: Support for Infix notation
The PR https://github.com/redhat-developer/quarkus-ls/pull/506 will support or resolver with the following syntax : 

`{name.or('John')}`

The support provides: 

 * completion after` name. `(it will show `or` method)
 * validation for `or` method
 * hover for `or` method (documentation about `or` method)

Qute supports too the another syntax:
 
`{name or 'John'}`

Today we don't support it and validation says that `or` is udenfined:

image

It seems Qute have a generic mechanism and this syntax can be used too for another methods (ex : for get of List), you can write

```
{items.get(0)}
{items get 0}
```

Supporting the syntax {`items get 0`} is not trivial, I wonder if Qute will continue to provide this syntax when it will adopt new parser (with ANTLR, if I understood correctly). 

If this syntax will be supported, we need to suppor it too. The strange thing is that it works only with method which have parameters. Ex if you write 

`{items orEmpty}` this syntax fails.

@mkouba @maxandersen  do you think this syntax will be continued to support it in the future?

angelozerr:
Support for Infix notation

Fixes #519

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/expression/Part.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/expression/Part.java
@@ -27,29 +27,10 @@ public abstract class Part extends Node {
 
 	private String partName;
 
+	private Boolean optional;
+
 	public Part(int start, int end) {
 		super(start, end);
-	}
-
-	/**
-	 * Returns the part name.
-	 * 
-	 * @return the part name.
-	 */
-	public String getPartName() {
-		if (partName != null) {
-			return partName;
-		}
-		return partName = getOwnerTemplate().getText(getStart(), getEndName());
-	}
-
-	/**
-	 * Returns the offset of the part name end.
-	 * 
-	 * @return the offset of the part name end.
-	 */
-	public int getEndName() {
-		return getEnd();
 	}
 
 	@Override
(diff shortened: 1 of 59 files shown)


angelozerr:
Here a demo with infix notation. All features should work (completion only for allowed methods (method which have one parameter), validation, definition,hover):

InfixNotationDemo

fbricon:
if I write `{mom or dad}` and neither mom or dad exist, only `mom` has a UndefinedObject warning

angelozerr:
> if I write {mom or dad} and neither mom or dad exist, only mom has a UndefinedObject warning

At first, this expression is the same than:

`{mom.or(dad)}`

The Qute validator has the same behavior than java editor, in other words it doesn't validate method parameter if object or method cannot be resolved (in your sample, mom cannot be resolved, so it stops the validation for the other parts)

image

fbricon:
while `foo charAt 1` seems valid, I don't think we should provide completion for it. Not sure we should provide completion for any of those methods. @mkouba @maxandersen what say you?

maxandersen:
What would you provide completion for ? Just "or' and "and" or nothing at all ?

fbricon:
there's an `and`?

If we can only provide `or`, I'd do just that, else, I would provide no completion at all. That's what makes most sense to *me*. But you guys tell us what to do.

angelozerr:
> What would you provide completion for ? Just "or' and "and" or nothing at all ?

Because it's generic, I follow the same rule than teh Qute parser from Quarkus. When I build the AST for infix notation or not, I build the same AST and completion works on this AST and provides getChars, etc.

mkouba:
In qute, the infix notation that can be used for _virtual methods_ with single parameter, i.e. `foo.bar(true)` can be written as `foo bar true`. It's about the template syntax, therefore we talk about virtual methods. In other words, completion could make sense for an expression which resolves to a java type which has methods with a single param.

angelozerr:
>  In other words, completion could make sense for an expression which resolves to a java type which has methods with a single param.

It seems that it works too for non virtual method:

```
{@java.lang.String name}
{name charAt 1}
```

is working (it call String.charAt(int index))

The compeltion shows all methods which a have only one parameter, but don't filter for virtual method since the previous sample is woking.

angelozerr:
> if I write {mom or dad} and neither mom or dad exist, only mom has a UndefinedObject warning

It should be fixed now:

image

mkouba:
> > In other words, completion could make sense for an expression which resolves to a java type which has methods with a single param.
> 
> It seems that it works too for non virtual method:

We call these methods "virtual" because they don't have to be backed by a real Java method. In this case, it is backed by a real method though. In other words, a method call in an output expression is called _virtual method_.

angelozerr:
@fbricon according to the @mkouba I think the PR provide the expected behavior, no?

The PR is now ready (code is cleaned and tests has been written).

fbricon:
Screenshot 2022-03-18 at 12 04 39

fbricon:
similar issue with `{items.get(0) or 'toto'}` 
Screenshot 2022-03-18 at 12 08 44

angelozerr:
> similar issue with {items.get(0) or 'toto'}

it should work better.

>  {items[0] or 'toto'}

I need to support this array syntax, in other worfs it doenswork

angelozerr:
Bracket is now supported, see the following demo:

ArrayDemo

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/expression/Part.java
```diff
@@ -49,9 +51,47 @@ public String getPartName() {
...
 		return getEnd();
 	}
 
+	public int getStartName() {
+		return getStart();
+	}
+	
+	private void computeOptionalIfNeeded() {
+		if (canBeOptional()) {
+			if (optional == null) {
```
fbricon, line 64: can be combined

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/expression/PropertyPart.java
```diff
@@ -36,6 +38,54 @@ public PartKind getPartKind() {
...
+
+	private synchronized void computeBracketNotation() {
+		if (bracketNotationComputed != null) {
+			return;
+		}
+		String text = getOwnerTemplate().getText();
+		int start = super.getStartName();
+		int end = super.getEndName() - 1;
+		if (text.charAt(start) == '[') {
+			if ((text.charAt(start + 1) == '\'' || text.charAt(start + 1) == '"' )) {
```
fbricon, line 68: do we need to check text.length if start+1 is out of range?

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/diagnostics/QuteErrorCode.java
```diff
@@ -23,6 +23,9 @@
 	// Error code for namespaces
 	UndefinedNamespace("No namespace resolver found for: `{0}`."), //
 
+	// Error code for operator
+	InvalidOperator("Invalid `{0}` operator for the section `#{1}`. Allowed operators are `{2}`."), //
```
fbricon, line 27: either `for section ``#{1}`` ` or `for the ``#{1}`` section `

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/diagnostics/QuteErrorCode.java
```diff
@@ -34,7 +37,9 @@
 	InvalidMethodParameter("The method `{0}` in the type `{1}` is not applicable for the arguments `{2}`."), //
 	InvalidVirtualMethod(
 			"The virtual method `{0}` in the type `{1}` is not applicable for the base object `{2}` type."), //
-
+	InfixNotationParameterRequired("A parameter for the infix notation method `{0}` is required."), //
+	InvalidMethodInfixNotation("The method `{0}` cannot be used with infix notation, because it has `{1}` parameters."), //
```
fbricon, line 41: because it has more than 1 parameter

fbricon (changes requested):

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/expression/Part.java
angelozerr: fixed

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/diagnostics/QuteErrorCode.java
angelozerr: I changed to `because it has not `1` parameter` because 0 parameter is forbidden too.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/diagnostics/QuteErrorCode.java
angelozerr: fixed

fbricon:
`{name "world"}` doesn't show an error

angelozerr:
Indeed but its à syntax error. I could manage it but I would prefer délégating this syntax to the réal qute parser.

angelozerr:
@mkouba I noticed that there are several syntax errors which is not reported. Do you think you could manage them?

mkouba:
> @mkouba I noticed that there are several syntax errors which is not reported. Do you think you could manage them?
> 
> I could report issues on quarkus github

For sure, pls go ahead and file an issue.

PR #609, opened 2022-03-17
Wrong method parameter resolution (ignores subtypes)

### LINKED ISSUE #608: Wrong method parameter resolution (ignores subtypes)
Writing `{item.name.contains('Apple')}` in a qute template yields a false positive error:
> The method `contains(CharSequence)` in the type `String` is not applicable for the arguments `(String)`.

Screenshot 2022-03-17 at 09 41 01

angelozerr:
Wrong method parameter resolution (ignores subtypes)

Fixes #608


fbricon:
It doesn't work if you have :
```java
public class SmallItem{}
public class Item extends SmallItem{}
public class BigItem extends Item{}
```
and have an `item.convert(SmallItem o)` method. then calling `item.convert(bigItem)` fails with:

```
The method `convert(SmallItem)` in the type `Item` is not applicable for the arguments `(BigItem)`.
```

angelozerr:
> and have an item.convert(SmallItem o) method. then calling item.convert(bigItem) fails with:

It is fixed (with test).

angelozerr:
@fbricon if it works for you,please merge it.

PR #610, opened 2022-03-17
Catch error when collect data model is process for a given provider.

angelozerr:
Catch error when collect data model is process for a given provider.


angelozerr:
When a data model provider (ex : collect data model with @CheckedTemplate) crashes, it breaks all collecting of data model.

This PR catch th eerror an ddoesn't break the collection.

fbricon (approved):

PR #612, opened 2022-03-19
Fix bug with super method.

angelozerr:
Fix bug with super method.

This PR fixes the follwoing problem, if you have :
```java
public class SmallItem{}
public class Item extends SmallItem{}
public class BigItem extends Item{}
```
and have an `smalItem.convert(SmallItem o)` method. then calling `item.convert(bigItem)` fails with:

```
`convert` cannot be resolved or is not a method of `org.acme.Item` Java type.
```

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/QuteProjectRegistry.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/QuteProjectRegistry.java
@@ -13,7 +13,6 @@ package com.redhat.qute.project;
 
 import static com.redhat.qute.parser.template.LiteralSupport.getPrimitiveObjectType;
 import static com.redhat.qute.services.QuteCompletableFutures.EXTENDED_TEMPLATE_DATAMODEL_NULL_FUTURE;
-import static com.redhat.qute.services.QuteCompletableFutures.FIELD_VALUE_RESOLVERS_NULL_FUTURE;
 import static com.redhat.qute.services.QuteCompletableFutures.JAVA_ELEMENT_INFO_NULL_FUTURE;
 import static com.redhat.qute.services.QuteCompletableFutures.METHOD_VALUE_RESOLVERS_NULL_FUTURE;
 import static com.redhat.qute.services.QuteCompletableFutures.NAMESPACE_RESOLVER_INFO_NULL_FUTURE;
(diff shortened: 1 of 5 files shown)


qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/QuteProjectRegistry.java
```diff
@@ -411,14 +401,13 @@ public JavaMemberInfo findMember(ResolvedJavaTypeInfo resolvedType, String prope
...
-		List<JavaFieldInfo> fields = resolvedType.getFields();
-		if (fields == null || fields.isEmpty() || isEmpty(fieldName)) {
-			return null;
-		}
-		for (JavaFieldInfo field : fields) {
-			if (fieldName.equals(field.getName())) {
-				return field;
+	protected static JavaFieldInfo findField(ResolvedJavaTypeInfo baseType, String fieldName) {
+		List<JavaFieldInfo> fields = baseType.getFields();
+		if (!(fields == null || fields.isEmpty() || isEmpty(fieldName))) {
```
fbricon, line 406: previous guard condition is easier to read/understand

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/QuteProjectRegistry.java
```diff
@@ -431,7 +420,7 @@ protected static JavaFieldInfo findField(ResolvedJavaTypeInfo resolvedType, Stri
 	 * @param propertyOrMethodName property or method name
 	 *
 	 * @return the member field retrieved by the given property or method name and
-	 *         null otherwise.
+	 *         null otherwis.
```
fbricon, line 423: otherwise

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/QuteProjectRegistry.java
angelozerr: fixed (like before)

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/QuteProjectRegistry.java
angelozerr: fixed

fbricon:
thx @angelozerr

PR #614, opened 2022-03-24
Added CodeAction to append `??` for `UndefinedObject` and ignored optional object from diagnostic

### LINKED ISSUE #613: CodeAction to add ?? at the end of the object part for UndefinedObject
It should be nice to provide a new code action to set the part as optional with `??`

Given this template:

`{foo}`

foo is higlighted as `UndefinedObject`, it should be nice to provide code action which add `??` like this:

`{foo??}`

AlexXuChen:
Added CodeAction to append ?? for `UndefinedObject` and ignored optional object from diagnostic.
Peek 2022-03-24 17-03

Fixes #613 

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
@@ -557,12 +557,17 @@ class QuteDiagnostics {
 
 		JavaTypeInfoProvider javaTypeInfo = objectPart.resolveJavaType();
 		if (javaTypeInfo == null) {
-			String literalJavaType = LiteralSupport.getLiteralJavaType(objectPart.getPartName()); 
+			if (objectPart.isOptional()) {
+				// The object part is optional (ex : foo??)
+				return null;
+			}
+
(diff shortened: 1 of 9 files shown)


AlexXuChen:
Need to fix tests

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
```diff
@@ -557,12 +557,16 @@ private ResolvedJavaTypeInfo validateObjectPart(String namespace, ObjectPart obj
 
 		JavaTypeInfoProvider javaTypeInfo = objectPart.resolveJavaType();
 		if (javaTypeInfo == null) {
-			String literalJavaType = LiteralSupport.getLiteralJavaType(objectPart.getPartName()); 
+			String literalJavaType = LiteralSupport.getLiteralJavaType(objectPart.getPartName());
+			if (objectPart.isOptional()) {
```
angelozerr, line 561: Move this test before 

`String literalJavaType = LiteralSupport.getLiteralJavaType(objectPart.getPartName());`

because we don't need to check if it is a litteral type if object part is optional.

angelozerr:
Thanks @AlexXuChen !

PR #619, opened 2022-03-30
Don't show undefined variable errors depending on context

### LINKED ISSUE #548: Don't show undefined variable errors depending on context
If a template checks a variable exists in a given context, eg. an if or for block, validation should know the variable exists and therefore should not complain about  undefined variables.eg.

Screenshot 2022-02-03 at 10 11 37

angelozerr:
Don't show undefined variable errors depending on context

Fixes #548


angelozerr:
I need to write tests, but optional parameter should work with diagnostics, definition, compeltion, higlight and linked editing range. See the following demo:

OptionalParameterDemo

angelozerr:
Writing `
{#let bar??=null}
   {#if bar}
    {bar}
   {/if}
{/let}

`
 is a thing which is not allowed even if Qute parser doesn't fail with syntax .{#let bar??=null}. In other words you cannot assign `bar??` to a value. So I think validation has teh correct behavior for UndefinedObject of bar.

@mkouba what do you think about this usecase? We should perhaps display an error when we try to assign parameter which ends with ??, no?

AlexXuChen (commented):
If we have the following:
```
{#let bar=null}
    {#if bar??}
        {bar}
    {/if}
{/let}
```
The following `UnknownType` diagnostic(s) occur:
Screenshot from 2022-03-30 13-19-04
I believe instead there should be no error.

angelozerr:
> I believe instead there should be no error.

Indeed, it's fixed now with a test.

AlexXuChen:
> Indeed, it's fixed now with a test.
Working as intended:
Screenshot from 2022-03-30 13-41-08

AlexXuChen (approved):
LGTM

mkouba:
> @mkouba what do you think about this usecase? We should perhaps display an error when we try to assign parameter which ends with ??, no?

I'm not sure. I think that `bar??` is a valid param name and that's why the parser does not fail.

angelozerr:
Thanks @AlexXuChen for your review.

mkouba:
> > I'm not sure. I think that bar?? is a valid param name and that's why the parser does not fail.
> 
> I tried:
> 
> ```
> {#let foo??='bar'}
>   {foo??}
> {/let}
> ```
> 
> And the result is empty although it should show `bar`. It means that `foo??` as parameter name can never be used.

Yes, it's legal but it can't be used for `{#let}` param names. Maybe we should just forbid the usage of `??`...

angelozerr:
> Maybe we should just forbid the usage of ??

Yes I think we should have an error like ?? cannot be used for parameter name.

PR #621, opened 2022-04-05
Update Quarkus LS to use LSP4MP 0.5.0 Snapshots.

rgrunber:
- Update to the next available snapshot release
- The final release of LSP4MP still gets deployed to the snapshot
  location, resulting in a failure to resolve snapshot artifacts
  post-release


angelozerr:
Thanks @rgrunber !

PR #631, opened 2022-05-05
Support missing attributes for @TemplateData / @RegisterForReflection

angelozerr:
Support missing attributes for @TemplateData / @RegisterForReflection


angelozerr:
@rgrunber @AlexXuChen you can play now with my PR which supports:

 * `@TemplateData/target`
 * `@TemplateData/ignore`
 *  `@RegisterForReflection/targets`

See the following demo:

TargetDemo

angelozerr:
I noticed that in native mode we have thos error with size of List:

image

Although size is avalable as value resolvers (with Map https://quarkus.io/guides/qute-reference#maps but with List too, see https://github.com/quarkusio/quarkus/blob/e24fec4e2c4bd10edbfefd96e59b3a6f3608aaa9/independent-projects/qute/core/src/main/java/io/quarkus/qute/ValueResolvers.java#L389)

We need to declare size as default value resolvers. We can do that in an another PR.

rgrunber:
In the process of understanding `@TemplateData`, I notice something odd. It's not related to this specifically, but might be something needing to be addressed :

According to https://quarkus.io/guides/qute-reference#native_executables, type-safe expressions are safe to use in a native image. So I defined a template through `@Inject` and added some data (eg. `.data("person", new Person(...))` to it. I then made sure the data type was declared as a parameter in the template (eg. `{@org.example.Person person}`). Then I enabled `qute.native.enabled` and noticed that the template file usage of `Person` was marked as invalid. However, the binary works just fine, as confirmed by the reference guide.

AlexXuChen (commented):
Something I noticed in relation to the `qute.native.enabled` setting is that when enabled, although the resolver list isn't populated with the field/method, the definition and hover still works. For example, if we use `java.lang.String.class` in `target`:
Peek 2022-05-17 12-02
This also applies to `ignore`, where the ignored field still has definition/hover support:
Peek 2022-05-17 12-09

angelozerr:
> Something I noticed in relation to the qute.native.enabled setting is that when enabled, although the resolver list isn't populated with the field/method, the definition and hover still works. 

Indeed but I think it is nice to have that because navigation can be helpfull to go the Java class easily and it follows the same behavior of JDT Java when you try to open a private method:

StringDemo

rgrunber (approved):
Type-safe expressions don't cause warnings to be emitted on the type when the native image setting is enabled, and target attribute for `@TemplateData` seems to be working for me.

PR #636, opened 2022-05-12
Added support for `@TemplateGlobal` annotation

### LINKED ISSUE #605: Support for `@TemplateGlobal`
Support for `@TemplateGlobal`. See https://quarkus.io/guides/qute-reference#global_variables

AlexXuChen:
Added support for `@TemplateGlobal` annotation.
templateglobaldemo

Fixes #605 

### DIFF
--- a/qute.jdt/com.redhat.qute.jdt.test/src/main/java/com/redhat/qute/jdt/template/TemplateGetDataModelProjectTest.java
+++ b/qute.jdt/com.redhat.qute.jdt.test/src/main/java/com/redhat/qute/jdt/template/TemplateGetDataModelProjectTest.java
@@ -60,6 +60,7 @@ public class TemplateGetDataModelProjectTest {
 		testValueResolversFromInject(resolvers);
 		testValueResolversFromTemplateData(resolvers);
 		testValueResolversFromTemplateEnum(resolvers);
+		testValueResolversFromTemplateGlobal(resolvers);
 	}
 
 	private static void testTemplates(DataModelProject<DataModelTemplate<DataModelParameter>> project) {
@@ -312,6 +313,30 @@ public class TemplateGetDataModelProjectTest {
 			resolvers);
 	}
 
+	private static void testValueResolversFromTemplateGlobal(List<ValueResolverInfo> resolvers) {
+
+		// @TemplateGlobal
+		// public class Globals {
+
+		// static int age = 40;
+		assertValueResolver(null, "age : int", "org.acme.qute.Globals", null, true, resolvers);
+
+		// static String name;
+		assertValueResolver(null, "name : java.lang.String", "org.acme.qute.Globals", null, true, resolvers);
+
+		// static Color[] myColors() {
+		// return new Color[] { Color.RED, Color.BLUE };
+		// }
+		assertValueResolver(null, "myColors() : org.acme.qute.Color[]", "org.acme.qute.Globals", null, true, resolvers);
+
+		// @TemplateGlobal(name = "currentUser")
+		// static String user() {
+		// return "Mia";
+		// }
+		assertValueResolver(null, "user() : java.lang.String", "org.acme.qute.Globals", "currentUser", true, resolvers);
+		// }
+	}
+
 	private static void assertValueResolver(String namespace, String signature, String sourceType,
 		List<ValueResolverInfo> resolvers) {
 		assertValueResolver(namespace, signature, sourceType, null, resolvers);
@@ -319,6 +344,11 @@ public class TemplateGetDataModelProjectTest {
 
 	private static void assertValueResolver(String namespace, String signature, String sourceType, String named,
 		List<ValueResolverInfo> resolvers) {
+		assertValueResolver(namespace, signature, sourceType, named, false, resolvers);
+	}
+
+	private static void assertValueResolver(String namespace, String signature, String sourceType, String named,
+		boolean globalVariable, List<ValueResolverInfo> resolvers) {
 		Optional<ValueResolverInfo> result = resolvers.stream().filter(r -> signature.equals(r.getSignature()))
 			.findFirst();
 		Assert.assertFalse("Find '" + signature + "' value resolver.", result.isEmpty());
@@ -326,6 +356,7 @@ public class TemplateGetDataModelProjectTest {
 		Assert.assertEquals(namespace, resolver.getNamespace());
 		Assert.assertEquals(signature, resolver.getSignature());
 		Assert.assertEquals(sourceType, resolver.getSourceType());
+		Assert.assertEquals(globalVariable, resolver.isGlobalVariable());
 	}
 
 	private static void assertParameter(String key, String sourceType, boolean dataMethodInvocation,
--- a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/datamodel/resolvers/ValueResolverInfo.java
+++ b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/datamodel/resolvers/ValueResolverInfo.java
@@ -17,7 +17,7 @@ import com.redhat.qute.commons.JavaElementKind;
 
 /**
  * Value resolver information.
- * 
+ *
  * @author Angelo ZERR
  *
  */
@@ -37,7 +37,7 @@ public class ValueResolverInfo {
 
 	/**
 	 * Returns the named of the resolver.
-	 * 
+	 *
 	 * @return the named of the resolver.
 	 */
 	public String getNamed() {
@@ -46,7 +46,7 @@ public class ValueResolverInfo {
 
 	/**
 	 * Set the named of the resolver.
-	 * 
+	 *
 	 * @param named the named of the resolver.
 	 */
 	public void setNamed(String named) {
@@ -64,7 +64,7 @@ public class ValueResolverInfo {
 
 	/**
 	 * Set the namespace of the resolver.
-	 * 
+	 *
 	 * @param namespace the namespace of the resolver.
 	 */
 	public void setNamespace(String namespace) {
@@ -90,7 +90,7 @@ public class ValueResolverInfo {
 
 	/**
 	 * Set the Java element signature.
-	 * 
+	 *
 	 * @param signature the Java element signature.
 	 */
 	public void setSignature(String signature) {
@@ -99,7 +99,7 @@ public class ValueResolverInfo {
 
 	/**
 	 * Returns the java source type and null otherwise.
-	 * 
+	 *
 	 * @return the java source type and null otherwise.
 	 */
 	public String getSourceType() {
@@ -108,7 +108,7 @@ public class ValueResolverInfo {
 
 	/**
 	 * Set the java source type.
-	 * 
+	 *
 	 * @param sourceType the java source type
 	 */
 	public void setSourceType(String sourceType) {
@@ -117,7 +117,7 @@ public class ValueResolverInfo {
 
 	/**
 	 * Returns true if it is a global variable and false otherwise.
-	 * 
+	 *
 	 * @return true if it is a global variable and false otherwise.
 	 */
 	public boolean isGlobalVariable() {
@@ -126,7 +126,7 @@ public class ValueResolverInfo {
 
 	/**
 	 * Set true if it is a global variable and false otherwise.
-	 * 
+	 *
 	 * @param global true if it is a global variable and false otherwise.
 	 */
 	public void setGlobalVariable(boolean global) {
@@ -156,6 +156,7 @@ public class ValueResolverInfo {
 		result = prime * result + ((namespace == null) ? 0 : namespace.hashCode());
 		result = prime * result + ((signature == null) ? 0 : signature.hashCode());
 		result = prime * result + ((sourceType == null) ? 0 : sourceType.hashCode());
+		result = prime * result + ((globalVariable == null) ? 0 : globalVariable.hashCode());
 		return result;
 	}
 
@@ -193,6 +194,11 @@ public class ValueResolverInfo {
 				return false;
 		} else if (!sourceType.equals(other.sourceType))
 			return false;
+		if (globalVariable == null) {
+			if (other.globalVariable != null)
+				return false;
+		} else if (!globalVariable.equals(other.globalVariable))
+			return false;
 		return true;
 	}
 
@@ -204,6 +210,7 @@ public class ValueResolverInfo {
 		b.add("matchName", this.matchName);
 		b.add("signature", signature);
 		b.add("sourceType", sourceType);
+		b.add("globalVariable", globalVariable);
 		return b.toString();
 	}
 }
--- a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
+++ b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
@@ -1,0 +1,202 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0git che
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.qute.jdt.internal.template.datamodel;
+
+import static com.redhat.qute.jdt.internal.QuteJavaConstants.TEMPLATE_GLOBAL_ANNOTATION;
+import static com.redhat.qute.jdt.internal.QuteJavaConstants.TEMPLATE_GLOBAL_ANNOTATION_NAME;
+
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.jdt.core.IAnnotatable;
+import org.eclipse.jdt.core.IAnnotation;
+import org.eclipse.jdt.core.IField;
+import org.eclipse.jdt.core.IJavaElement;
+import org.eclipse.jdt.core.IMember;
+import org.eclipse.jdt.core.IMethod;
+import org.eclipse.jdt.core.IType;
+import org.eclipse.jdt.core.JavaModelException;
+
+import com.redhat.qute.commons.datamodel.resolvers.ValueResolverInfo;
+import com.redhat.qute.jdt.QuteSupportForTemplate;
+import com.redhat.qute.jdt.internal.resolver.ITypeResolver;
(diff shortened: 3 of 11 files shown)


qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
```diff
@@ -0,0 +1,154 @@
...
+	@Override
+	protected void processAnnotation(IJavaElement javaElement, IAnnotation annotation, String annotationName,
+		SearchContext context, IProgressMonitor monitor) throws JavaModelException {
+		if (javaElement.getElementType() != IJavaElement.TYPE) {
+			return;
+		}
+		IType type = (IType) javaElement;
+		String name = null;
+		boolean hasTemplateGlobal = false;
+		for (IAnnotation typeAnnotation : type.getAnnotations()) {
```
angelozerr, line 85: I suggest that you see the @Target of @TemplateGlobal at https://github.com/quarkusio/quarkus/blob/main/independent-projects/qute/core/src/main/java/io/quarkus/qute/TemplateGlobal.java#L30

It means that : 

 * you can annotate type, field and method.
 * you can annotate those java elements with only one annotation.

It means that : 

 * you must not iterate from @TemplateGlobal
 * you can have @TemplateGlobal annotation in type, field, and method. What is the behavior if type is not annotated with @TemplateGlobal but a field is annotated with @TemplateGlobal. I suggest that you test that and you provide the code which have the same behavior of Qute.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
```diff
@@ -0,0 +1,154 @@
...
+			}
+		}
+		if (!hasTemplateGlobal) {
+			return;
+		}
+
+		ITypeResolver typeResolver = QuteSupportForTemplate.createTypeResolver(type);
+
+		// Loop for static fields
+		IField[] fields = type.getFields();
```
angelozerr, line 101: I suggest that you read Javadoc at https://github.com/quarkusio/quarkus/blob/5e6fe630375c96b18228f54c6537f445f3ee1743/independent-projects/qute/core/src/main/java/io/quarkus/qute/TemplateGlobal.java#L10 which explains rules that field and method must match (ex : static, etc) and implement those rules.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
```diff
@@ -0,0 +1,154 @@
...
+				typeResolver, monitor);
+		}
+	}
+
+	private void collectResolversForTemplateGlobal(IMember member, String name, List<ValueResolverInfo> resolvers,
+		ITypeResolver typeResolver, IProgressMonitor monitor) {
+		try {
+			// every non-void non-private static method that declares no parameters and
+			// every non-private static field is considered a global variable
+			if (!Modifier.isPrivate(member.getFlags()) && Modifier.isStatic(member.getFlags())) {
```
angelozerr, line 121: It is not enough, I suggest that you read Javadoc at https://github.com/quarkusio/quarkus/blob/5e6fe630375c96b18228f54c6537f445f3ee1743/independent-projects/qute/core/src/main/java/io/quarkus/qute/TemplateGlobal.java#L10 which explains rules that field and method must match (ex : static, etc) and implement those rules.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
AlexXuChen: So both a field and method must be non private and static, and a method must have no parameters and return void, correct?

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
AlexXuChen: So both a field and method must be non private and static, and a method must have no parameters and return void, correct?

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
```diff
@@ -0,0 +1,186 @@
...
+		SearchContext context, IProgressMonitor monitor) throws JavaModelException {
+		if (!(javaElement instanceof IAnnotatable)) {
+			return;
+		}
+		IAnnotation templateGlobal = AnnotationUtils.getAnnotation((IAnnotatable) javaElement,
+			TEMPLATE_GLOBAL_ANNOTATION);
+		if (templateGlobal == null) {
+			return;
+		}
+		if (javaElement instanceof IType) {
```
angelozerr, line 86: `  if (javaElement.getElementType() == IJavaElement.TYPE) {`

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
```diff
@@ -0,0 +1,186 @@
...
+			IType type = (IType) javaElement;
+			collectResolversForTemplateGlobal(type, templateGlobal, context.getDataModelProject().getValueResolvers(),
+				monitor);
+		} else {
+			ITypeResolver typeResolver = QuteSupportForTemplate.createTypeResolver((IType) javaElement);
+
+			if (javaElement instanceof IMethod) {
```
angelozerr, line 93:
```
  if (javaElement.getElementType() == IJavaElement.FIELD || javaElement.getElementType() == IJavaElement.METHOD) {
    IMember member = (IMember) javaElement;
    collectResolversForTemplateGlobal(member, context.getDataModelProject().getValueResolvers(),
					typeResolver, monitor);
```

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
angelozerr: Yes, it misses the conditions for methods (no parameters and return void)

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
```diff
@@ -0,0 +1,186 @@
...
+	private void collectResolversForTemplateGlobal(IMember member, List<ValueResolverInfo> resolvers,
+		ITypeResolver typeResolver, IProgressMonitor monitor) {
+		// every non-void non-private static method that declares no parameters and
+		// every non-private static field is considered a global variable
+		if (isTemplateGlobalMember(member)) {
+			String sourceType = member.getDeclaringType().getFullyQualifiedName();
+			ValueResolverInfo resolver = new ValueResolverInfo();
+			resolver.setSourceType(sourceType);
+			resolver.setSignature(getSignature(member, typeResolver));
+			// The name is defaulted, i.e. the name of the field/method is used.
```
angelozerr, line 135: `resolver.setGlobalVariable(true);`

angelozerr: It seems that you can override the name of @TemplateGlobal

```
@TemplateGlobal("foo");
private String bar;
```

Your value resolver doesn't take care of that. You should get the "foo" value and if it is not empty, you should set the value to the generated resolver:

```
IAnnotation templateGlobal = AnnotationUtils.getAnnotation(member,TEMPLATE_GLOBAL_ANNOTATION);
if (templateGlobal  != null) {
    // get named from @TemplateGlobal
    String named = ...
    resolver.setNamed(named);
}
```

AlexXuChen:
Template global example:
```java
package org.acme;

import io.quarkus.qute.TemplateGlobal;

enum Color { RED, GREEN, BLUE }

@TemplateGlobal
public class Globals {

    static int age = 40;

    static String name;

    static Color[] myColors() {
      return new Color[] { Color.RED, Color.BLUE };
    }

    @TemplateGlobal(name = "currentUser")
    static String user() {
       return "Mia";
    }
}
```

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
AlexXuChen: Quarkus doesn't seem to support the declaration of `@TemplateGlobal("foo")`, and the compilation fails when I build the binary:
Screenshot from 2022-05-26 11-22-02
Should this be  "name" instead?

AlexXuChen:
`name` attribute is also supported:
templateglobaldemoname

qute.jdt/com.redhat.qute.jdt.test/src/main/java/com/redhat/qute/jdt/template/TemplateGetDataModelProjectTest.java
```diff
@@ -312,6 +313,30 @@ private static void testValueResolversFromTemplateEnum(List<ValueResolverInfo> r
 			resolvers);
 	}
 
+	private static void testValueResolversFromTemplateGlobal(List<ValueResolverInfo> resolvers) {
+
+		// @TemplateGlobal
+		// public class Globals {
+
+		// static int age = 40;
+		assertValueResolver(null, "age : int", "org.acme.qute.Globals", resolvers);
```
angelozerr, line 322: It should be nice to check globalVariable (true).

Your test should look like:

`assertValueResolver(null, "age : int", "org.acme.qute.Globals", true, resolvers);`

Inother words add the globalvariable parameter to assertValueResolver

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/datamodel/resolvers/ValueResolverInfo.java
```diff
@@ -156,6 +156,7 @@ public int hashCode() {
 		result = prime * result + ((namespace == null) ? 0 : namespace.hashCode());
 		result = prime * result + ((signature == null) ? 0 : signature.hashCode());
 		result = prime * result + ((sourceType == null) ? 0 : sourceType.hashCode());
+		result = prime * result + ((globalVariable == null) ? 0 : globalVariable.hashCode());
```
angelozerr, line 159: Good catch! Could you do that too in qute.ls please.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
```diff
@@ -0,0 +1,206 @@
...
+			TEMPLATE_GLOBAL_ANNOTATION);
+		if (templateGlobal == null) {
+			return;
+		}
+		ITypeResolver typeResolver = QuteSupportForTemplate.createTypeResolver((IMember) javaElement);
+		if (javaElement.getElementType() == IJavaElement.TYPE) {
+			IType type = (IType) javaElement;
+			collectResolversForTemplateGlobal(type, templateGlobal, context.getDataModelProject().getValueResolvers(),
+				typeResolver, monitor);
+		} else {
```
angelozerr, line 92:
```java
} else if (javaElement.getElementType() == IJavaElement.FIELD
				|| javaElement.getElementType() == IJavaElement.METHOD) {
...
```

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
```diff
@@ -0,0 +1,206 @@
...
+		return ANNOTATION_NAMES;
+	}
+
+	@Override
+	protected void processAnnotation(IJavaElement javaElement, IAnnotation annotation, String annotationName,
+		SearchContext context, IProgressMonitor monitor) throws JavaModelException {
+		if (!(javaElement instanceof IAnnotatable)) {
+			return;
+		}
+		IAnnotation templateGlobal = AnnotationUtils.getAnnotation((IAnnotatable) javaElement,
```
rgrunber, line 82: I think `templateGlobal` can be removed in favour of just using `annotation` here, unless I'm missing something.

They certainly match, and when I look at `AbstractAnnotationTypeReferenceDataModelProvider#collectDataModel(..)` the java element is derived from calling `getParent()` on the annotation.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
```diff
@@ -0,0 +1,206 @@
...
+	private static boolean isTemplateGlobalMember(IMember member) {
+		try {
+			// every non-void non-private static method that declares no parameters and
+			// every non-private static field is considered a global variable
+			if (!Modifier.isPrivate(member.getFlags()) && Modifier.isStatic(member.getFlags())) {
+				if (member.getElementType() == IJavaElement.FIELD) {
+					return true;
+				} else if (member.getElementType() == IJavaElement.METHOD) {
+					IMethod method = (IMethod) member;
+					return method.getParameters().length == 0 && !"void".equals(method.getReturnType());
```
rgrunber, line 179: I would use `method.getNumberOfParameters()`

Also, try defining a void method like :

`public static void doesNothing () {}`

It shows up in the completion when the class is annotated with `@TemplateGlobal` but it shouldn't.

In fact, the same problem exists with `@TemplateExtension`. It doesn't exclude void values.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
```diff
@@ -0,0 +1,206 @@
...
+			}
+		}
+	}
+
+	private void collectResolversForTemplateGlobal(IType type, IAnnotation templateGlobal,
+		List<ValueResolverInfo> resolvers, ITypeResolver typeResolver, IProgressMonitor monitor) {
+		try {
+			IField[] fields = type.getFields();
+			for (IField field : fields) {
+				if (AnnotationUtils.getAnnotation(field, TEMPLATE_GLOBAL_ANNOTATION) == null
```
rgrunber, line 107: I would use `!AnnotationUtils.hasAnnotation(..)`, here and below.

rgrunber (changes requested):
Works pretty well for me, but noticed a few things worth fixing.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
AlexXuChen: There seems to be an issue with `org.eclipse.jdt.core.Signature`, where it receives a `char[]` of the typeName:

but the resulting string is the first character:

There might be an alternative to `IMethid.getReturnType()` instead?

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
angelozerr: Yes sorry, I meant "name" attribute.

angelozerr:
There is compilation problem https://github.com/redhat-developer/quarkus-ls/runs/6632720210?check_suite_focus=true#step:8:142 you miss to define this method since you add globalVariable parameter (you define a new assertResolver by calling false for globalVariable).

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
```diff
@@ -0,0 +1,204 @@
...
+		return ANNOTATION_NAMES;
+	}
+
+	@Override
+	protected void processAnnotation(IJavaElement javaElement, IAnnotation annotation, String annotationName,
+		SearchContext context, IProgressMonitor monitor) throws JavaModelException {
+		if (!(javaElement instanceof IAnnotatable)) {
+			return;
+		}
+//		IAnnotation tempalteGlobal = AnnotationUtils.getAnnotation((IAnnotatable) javaElement,
```
angelozerr, line 82: remove this comments

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/datamodel/TemplateGlobalAnnotationSupport.java
```diff
@@ -0,0 +1,204 @@
...
+			ValueResolverInfo resolver = new ValueResolverInfo();
+			resolver.setSourceType(sourceType);
+			resolver.setSignature(getSignature(member, typeResolver));
+			// Constant value for {@link #name()} indicating that the field/method name
+			// should be used
+			try {
+				resolver.setNamed(
+					AnnotationUtils.getAnnotationMemberValue(templateGlobal, TEMPLATE_GLOBAL_ANNOTATION_NAME));
+			} catch (JavaModelException e) {
+				LOGGER.log(Level.SEVERE, "Error while getting annotation member value of 'named'.", e);
```
angelozerr, line 135: Error while getting annotation member value of 'name'

angelozerr:
Great job @AlexXuChen !

PR #638, opened 2022-05-17
Linked editing doesn't work if variable is used as a parameter into a section

angelozerr:
Linked editing doesn't work if variable is used as a parameter into a section

Fixes https://github.com/redhat-developer/vscode-quarkus/issues/500

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/definition/QuteDefinitionInForUserTagTest.java
+++ b/qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/definition/QuteDefinitionInForUserTagTest.java
@@ -1,0 +1,54 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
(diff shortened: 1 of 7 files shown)


qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/definition/QuteDefinitionInForUserTagTest.java
```diff
@@ -0,0 +1,54 @@
...
+	public void it() throws Exception {
+		String template = "{#for item in items}\r\n" + //
+				"	{#linkItem it|em name=item.name item /}\r\n" + //
+				"{/for}";
+		testDefinitionFor(template, "test.qute", //
+				ll("test.qute", r(1, 12, 1, 16), r(0, 6, 0, 10)));
+	}
+
+	@Test
+	public void noIt() throws Exception {
```
fbricon, line 47: what do it/noIt mean?

angelozerr: See https://quarkus.io/guides/qute-reference#user_tags

angelozerr: in user tag you can write parameter without value like

`{#user name}` 

in this case name is the it of the user tag.

You can hav ejust one it, so if you write

`{#user name name2}` 

name2 is not an it (it should have an error report for that)

fbricon:
thanks @angelozerr

PR #640, opened 2022-05-17
template validation complains about strings containing spaces

### LINKED ISSUE #639: template validation complains about strings containing spaces
Given the following snippet in a qute template:
```
{name ?: "Quarkus Insights"}
```
validation displays an error: `A parameter for the infix notation method "Insights" is required`

angelozerr:
template validation complains about strings containing spaces

Fixes #639


fbricon:
So in the following case, ` {name or "salt'n peppa" }` both the qute-ls and actual qute parsers consider the `'` to be terminating  the string. That's pretty odd. Quarkus fails with `#1 Parser error in template [hello.html] on line 9: unexpected non-text buffer at the end of the template - unterminated string literal: name or "salt'n peppa" }!` . qute-ls parser fails with `Parser error on line 9: unexpected non-text buffer at the end of the template - unterminated string literal: name or "salt'n peppa" }! </p></body></html>`
@mkouba is this intentional?

fbricon:
Merging the PR regardless of https://github.com/redhat-developer/quarkus-ls/pull/640#issuecomment-1129096464, since it's consistent with actual Qute validation

mkouba:
> is this intentional?

Well, yes according to the current rules. TBH I've never had a need to use an apostrophe inside a string literal but I agree that it could be useful. Pls file an issue.

PR #649, opened 2022-05-30
Varargs are not properly supported in Qute template validation

### LINKED ISSUE #509: Varargs are not properly supported in Qute template validation
Given a template extension method with varargs, such as : 
```java
@TemplateExtension
public class TemplateExtensions {

    static String pretty(Item item, String... elements) { 
        return item.name + " "+ String.join(", ", elements);
    }
}
```
Validation fails in the Qute template for: 

```
{item.pretty("one", "two")}
```
> The method `pretty(String[])` in the type `TemplateExtensions` is not applicable for the arguments `(String, String)`. qute(InvalidMethodParameter)

Screenshot 2022-01-04 at 17 51 57

angelozerr:
Varargs are not properly supported in Qute template validation

Fixes #509

### DIFF
--- a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaMethodInfo.java
+++ b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaMethodInfo.java
@@ -239,6 +239,15 @@ public class JavaMethodInfo extends JavaMemberInfo {
 		}
 		return parameters;
 	}
+	
+	/**
+	 * Returns the parameters length of the method.
+	 * 
+	 * @return the parameters length of the method.
+	 */
+	public int getParameterslength() {
+		return getParameters().size() - (isVirtual() ? 1 : 0);
+	}
 
 	private static List<JavaParameterInfo> parseParameters(String signature) {
 		List<JavaParameterInfo> parameters = new ArrayList<>();
--- a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaParameterInfo.java
+++ b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaParameterInfo.java
@@ -21,11 +21,15 @@ import org.eclipse.xtext.xbase.lib.util.ToStringBuilder;
  */
 public class JavaParameterInfo extends JavaElementInfo {
 
+	private static final String VARARGS = "...";
+
 	private final String name;
 
 	private final String type;
 
 	private transient JavaTypeInfo javaType;
+
+	private String varArgType;
 
 	public JavaParameterInfo(String name, String type) {
 		this.name = name;
@@ -49,6 +53,29 @@ public class JavaParameterInfo extends JavaElementInfo {
 	 */
 	public String getType() {
 		return type;
+	}
+
+	/**
+	 * Returns the Java type parameter as String.
+	 *
+	 * @return the Java type parameter as String.
+	 */
+	public String getVarArgType() {
+		if (varArgType == null && isVarargs()) {
+			varArgType = type.substring(0, type.length() - 3);
+		}
+		return varArgType;
+	}
+
+	/**
+	 * Returns true if the Java type is a varargs (ex : java.lang.String...) and
+	 * false otherwise.
+	 * 
+	 * @return true if the Java type is a varargs (ex : java.lang.String...) and
+	 *         false otherwise.
+	 */
+	public boolean isVarargs() {
+		return type.endsWith(VARARGS);
 	}
 
 	/**
@@ -84,6 +111,15 @@ public class JavaParameterInfo extends JavaElementInfo {
 	}
 
 	@Override
+	public String getJavaElementSimpleType() {
+		if (isVarargs()) {
+			String type = getVarArgType();
+			return getSimpleType(type) + VARARGS;
+		}
+		return super.getJavaElementSimpleType();
+	}
+
+	@Override
 	public JavaElementKind getJavaElementKind() {
 		return JavaElementKind.PARAMETER;
 	}
@@ -98,6 +134,7 @@ public class JavaParameterInfo extends JavaElementInfo {
 		ToStringBuilder b = new ToStringBuilder(this);
 		b.add("name", this.getName());
 		b.add("type", this.getType());
+		b.add("varargs", this.isVarargs());
 		b.add("signature", this.getSignature());
 		return b.toString();
 	}
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaParameterInfo.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaParameterInfo.java
@@ -21,11 +21,15 @@ import org.eclipse.xtext.xbase.lib.util.ToStringBuilder;
  */
 public class JavaParameterInfo extends JavaElementInfo {
 
+	private static final String VARARGS = "...";
+
 	private final String name;
 
 	private final String type;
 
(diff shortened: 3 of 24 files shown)


angelozerr:
This PR should work, I need to clean my code and write some tests again.

Here a demo:

VarargsDemo

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/project/MockQuteProject.java
```diff
@@ -161,4 +162,18 @@ protected static ValueResolverInfo createValueResolver(String namespace, String
...
 
+	public JavaMethodInfo getMethodValueResolver(String typeName, String methodName) {
+		try {
+			List<MethodValueResolver>  resolvers = super.getDataModelProject().get().getMethodValueResolvers();
+			for (MethodValueResolver resolver : resolvers) {
+				if (typeName.equals(resolver.getSourceType()) && methodName.equals(resolver.getMethodName())) {
+					return resolver;
+				}
+			}
+		} catch (Exception e) {
```
fbricon, line 173: would it be useful to at least log the error?

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/project/MockQuteProjectRegistry.java
```diff
@@ -101,28 +105,40 @@ public CompletableFuture<Location> getJavaDefinition(QuteJavaDefinitionParams pa
...
+		
+		if (definitionRange != null) {
+			int index = className.indexOf('<');
+			if (index != -1) {
+				// ex : java.util.List<E>
+				// Remove generic
+				className = className.substring(0, index);
 			}
+			// ex : java/util/List.java
+			String javeFileUri = className.replaceAll("[.]", "/") + ".java";
```
fbricon, line 139: Could you use .replace instead in this case? See https://medium.com/javarevisited/micro-optimizations-in-java-string-replaceall-c6d0edf2ef6

fbricon (changes requested):

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/project/MockQuteProjectRegistry.java
angelozerr: I can omptimize that, but to be honnest with you, it is used just for test. Do you think it is interesting to do that?

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/project/MockQuteProject.java
angelozerr: I can do that, but it is in the test and in this case CompletableFuture is resolved immediatly. Is it really interesting to log that?

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/project/MockQuteProjectRegistry.java
fbricon: just good practice to follow

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/project/MockQuteProject.java
fbricon: that's why I'm asking. if not useful then ignore it.

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/project/MockQuteProjectRegistry.java
angelozerr: You are right, fixed. Thanks for this helpful article!

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaMethodInfo.java
```diff
@@ -239,6 +239,15 @@ public List<JavaParameterInfo> getParameters() {
...
 		return parameters;
 	}
+	
+	/**
+	 * Returns the parameters length of the method.
+	 * 
+	 * @return the parameters length of the method.
+	 */
+	public int getParameterslength() {
+		return getParameters().size() - (isVirtual() ? 1 : 0);
```
AlexXuChen, line 249: Why does the `isVirtual` method simply return false? 
https://github.com/redhat-developer/quarkus-ls/blob/9dd5a92aab304bf55d5808e4daf6ffefa1c836ea/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaMethodInfo.java#L65-L67
If this is the case, we could just use `getParameters().size()`

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaParameterInfo.java
```diff
@@ -21,12 +21,16 @@
  */
 public class JavaParameterInfo extends JavaElementInfo {
 
+	private static final String VARARGS = "...";
```
AlexXuChen, line 24: Would it be better to keep this in `QuteJavaConstants.java`?

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaParameterInfo.java
```diff
@@ -21,12 +21,16 @@
  */
 public class JavaParameterInfo extends JavaElementInfo {
 
+	private static final String VARARGS = "...";
```
AlexXuChen, line 24: Same as above, since you use it twice, except this time on the ls side

AlexXuChen (changes requested):
varargs validation is working for me

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaParameterInfo.java
angelozerr: Qute JDT has this constant but not Qute LS. So we cannot do that, because all classes from `com.redhat.qute.commons` are duplicated in Qute JDT and Qute LS. The clean solution should be have a new project `com.redhat.qute.commons` where Qute JDT and Qute LS depends on it, but it is annoying to do that with OSGi (for Qute JDT).

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaParameterInfo.java
angelozerr: Qute LS has not QuteJavaConstants. More this constants is used only in JavaParameterInfo, so w e don't need to expose it in an Java constants file as public.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaMethodInfo.java
angelozerr: It does that https://github.com/redhat-developer/quarkus-ls/blob/b1271618bc7b308fd88f3c60c6f0f1be3c2c1564/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/datamodel/resolvers/MethodValueResolver.java#L49 no ?

AlexXuChen: Ah sorry, yes you are right, I was looking at JavaMethodInfo still

PR #656, opened 2022-06-09
Improved completion when nested in Qute `#if` section

### LINKED ISSUE #497: Completion for nested block section
Some section like #for, #when, #switch, #if have nested block like #else, #id, #case

We should provide completion for those nested block.

AlexXuChen:
Improved completion with `#else` and `#elseif` when nested in Qute `#if` section.
Peek 2022-06-09 15-40

Part of #497

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
@@ -1,0 +1,106 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.qute.services.snippets;
+
+import java.io.IOException;
+import java.util.Map;
+
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
+import com.google.gson.stream.JsonWriter;
+import com.redhat.qute.parser.template.Node;
+import com.redhat.qute.parser.template.NodeKind;
+import com.redhat.qute.parser.template.Section;
+import com.redhat.qute.services.completions.CompletionRequest;
+
+/**
+ * Snippet context for Qute sections that have parent sections.
+ *
+ * e.g.
+ *
+ * "context": { "parent": "if" }
+ *
+ */
+
+public class QuteParentSnippetContext implements IQuteSnippetContext {
+
+	public static final TypeAdapter<QuteParentSnippetContext> IN_PARENT = new SnippetContextForQuteParentAdapter();
+
+	private final String parent;
+
+	public QuteParentSnippetContext(String parent) {
+		this.parent = parent;
+	}
+
+	@Override
+	public boolean isMatch(CompletionRequest request, Map<String, String> model) {
+		Node node = request.getNode();
+		// Completion is triggered inside a section tag which matches the parent
+		// e.g.
+		// {#if ...}
+		// | --> here {#else} completion is shown, the cursor is nested in #if parent.
+		// {/if}
+		if (isMatchParent(node)) {
+			if ((node.getKind() == NodeKind.Template || node.getKind() == NodeKind.Text)) {
+				return true;
+			} else if (node.getKind() == NodeKind.Section) {
+				Section section = (Section) node;
+				return section.isInStartTagName(request.getOffset());
+			}
+		}
+		return false;
+	}
+
+	private static class SnippetContextForQuteParentAdapter extends TypeAdapter<QuteParentSnippetContext> {
+
+		@Override
+		public QuteParentSnippetContext read(final JsonReader in) throws IOException {
+			JsonToken nextToken = in.peek();
+			if (nextToken == JsonToken.NULL) {
+				return null;
+			}
+
+			in.beginObject();
+			while (in.hasNext()) {
+				String name = in.nextName();
+				switch (name) {
+				case "parent":
+					if (in.peek() == JsonToken.STRING) {
+						return new QuteParentSnippetContext(in.nextString());
+					}
+					break;
+				default:
+					in.skipValue();
+				}
+			}
+			in.endObject();
+			return null;
+		}
+
+		@Override
+		public void write(JsonWriter out, QuteParentSnippetContext value) throws IOException {
+			// Do nothing
+		}
+	}
+
+	/**
+	 * Return true if parent section matches context parent, false otherwise.
+	 *
+	 * @param node the completion request node
+	 * @return true if parent section matches context parent, false otherwise.
+	 */
+	private boolean isMatchParent(Node node) {
+		return (node.getParentSection() != null && parent.equals(node.getParentSection().getTag()));
+	}
+
+}
--- a/qute.ls/com.redhat.qute.ls/src/main/resources/com/redhat/qute/services/snippets/qute-nested-snippets.json
+++ b/qute.ls/com.redhat.qute.ls/src/main/resources/com/redhat/qute/services/snippets/qute-nested-snippets.json
@@ -1,0 +1,22 @@
+{
+    "#else": {
+        "prefix": "else",
+        "body": [
+            "{#else}$0"
+        ],
+        "description": "Else section",
+        "context": {
+            "parent": "if"
+        }
+    },
+    "#elseif": {
+        "prefix": "elseif",
+        "body": [
+            "{#else if ${1:condition}}$0"
+        ],
+        "description": "Else If section",
+        "context": {
+            "parent": "if"
+        }
+    }
+}
--- a/qute.ls/com.redhat.qute.ls/src/main/resources/com/redhat/qute/services/snippets/qute-snippets.json
+++ b/qute.ls/com.redhat.qute.ls/src/main/resources/com/redhat/qute/services/snippets/qute-snippets.json
@@ -33,7 +33,7 @@
 		"description": "If section"
 	},
 	"#else": {
-		"prefix": "else",
+		"prefix": "if-else",
 		"body": [
 			"{#if ${1:condition}}",
 			"\t$2",
@@ -44,7 +44,7 @@
 		"description": "Else section"
 	},
 	"#elseif": {
-		"prefix": "elseif",
+		"prefix": "if-elseif",
 		"body": [
 			"{#if ${1:condition}}",
 			"\t$2",
@@ -98,7 +98,7 @@
 			"{/set}"
 		],
 		"description": "Let section"
-	},		
+	},
 	"#switch": {
 		"prefix": "switch",
 		"body": [
@@ -107,7 +107,7 @@
 			"{/switch}"
 		],
 		"description": "Switch section"
-	},		
+	},
 	"#with": {
 		"prefix": "with",
 		"body": [
@@ -116,7 +116,7 @@
 			"{/with}"
 		],
 		"description": "With section"
-	},		
+	},
 	"#when": {
 		"prefix": "when",
 		"body": [
--- a/qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/completions/QuteSnippetCompletionInIfSectionTest.java
+++ b/qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/completions/QuteSnippetCompletionInIfSectionTest.java
@@ -1,0 +1,92 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.qute.services.completions;
+
+import static com.redhat.qute.QuteAssert.SECTION_SNIPPET_SIZE;
+import static com.redhat.qute.QuteAssert.c;
+import static com.redhat.qute.QuteAssert.r;
+import static com.redhat.qute.QuteAssert.testCompletionFor;
+
+import org.junit.jupiter.api.Test;
+
+/**
+ * Tests for Qute snippet completion in #if section.
+ *
+ */
+public class QuteSnippetCompletionInIfSectionTest {
+
+	@Test
+	public void singleIf() throws Exception {
+		String template = "|";
+		testCompletionFor(template, //
+				true, // snippet support
+				c("if", //
+						"{#if ${1:condition}}" + System.lineSeparator() + //
+								"	$0" + System.lineSeparator() + //
+								"{/if}", //
+						r(0, 0, 0, 0)));
+	}
+
+	@Test
+	public void singleElse() throws Exception {
+		String template = "|";
+		testCompletionFor(template, //
+				true, // snippet support
+				c("if-else", //
+						"{#if ${1:condition}}" + System.lineSeparator() + //
+								"	$2" + System.lineSeparator() + //
+								"{#else}" + System.lineSeparator() + //
+								"	$0" + System.lineSeparator() + //
+								"{/if}", //
+						r(0, 0, 0, 0)));
+	}
+
+	@Test
+	public void singleElseIf() throws Exception {
+		String template = "|";
+		testCompletionFor(template, //
+				true, // snippet support
+				c("if-elseif", //
+						"{#if ${1:condition}}" + System.lineSeparator() + //
+								"	$2" + System.lineSeparator() + //
+								"{#else if ${3:condition}}" + System.lineSeparator() + //
+								"	$4" + System.lineSeparator() + //
+								"{#else}" + System.lineSeparator() + //
+								"	$0" + System.lineSeparator() + //
+								"{/if}", //
+						r(0, 0, 0, 0)));
+	}
+
+	@Test
+	public void snippetCompletionInNestedIf() throws Exception {
+		String template = "{#if condition}\n" + //
+				"|\n" + //
+				"{/if}";
+		testCompletionFor(template, //
+				true, // snippet support
+				SECTION_SNIPPET_SIZE /* #each, #if, ... */ + 2, /* #else #else if */ //
+				c("else", "{#else}$0", r(1, 0, 1, 0)), //
+				c("elseif", "{#else if ${1:condition}}$0", r(1, 0, 1, 0)));
+	}
+
+	@Test
+	public void elseSnippetCompletionInNestedIf() throws Exception {
+		String template = "{#if condition}\n" + //
+				"{#el|}\n" + //
+				"{/if}";
+		testCompletionFor(template, //
+				true, // snippet support
+				SECTION_SNIPPET_SIZE /* #each, #if, ... */ + 2, /* #else #else if */ //
+				c("else", "{#else}$0", r(1, 0, 1, 5)), //
+				c("elseif", "{#else if ${1:condition}}$0", r(1, 0, 1, 5)));
+	}
+}
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteSnippetRegistryLoader.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteSnippetRegistryLoader.java
@@ -25,6 +25,8 @@ public class QuteSnippetRegistryLoader implements ISnippetRegistryLoader<Snippet
 	public void load(SnippetRegistry<Snippet> registry) throws Exception {
 		registry.registerSnippets(QuteSnippetRegistryLoader.class.getResourceAsStream("qute-snippets.json"),
 				QuteSnippetContext.IN_TEXT);
+		registry.registerSnippets(QuteSnippetRegistryLoader.class.getResourceAsStream("qute-nested-snippets.json"),
+				QuteParentSnippetContext.IN_PARENT);
 	}
 
 }


AlexXuChen (commented):
There's a few limitations to my implementation:
if there is indentation before completion occurs, the indentation will remain, i.e. completing
```
{#if condition}
    |
{/if}
```
will give you
```
{#if condition}

    {#else}

{/if}
```
if there is white space between the opening and closing section tags, they will also remain, i.e. completing
```
{#if condition}

|

{/if}
```
will give you (extra line breaks)
```
{#if condition}


{#else}


{/if}
```

qute.ls/com.redhat.qute.ls/src/main/resources/com/redhat/qute/services/snippets/qute-snippets.json
```diff
@@ -41,6 +41,11 @@
 			"\t$0",
 			"{/if}"
 		],
+		"nestedBody": [
+			"\t$2",
```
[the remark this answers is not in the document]
angelozerr: No, but at first please don't update the qute-snippets.json. Those snippets are for "standard" snippet. Managing nested section means that there is a new one condition : parent section. Indeed {#else} must appears in completion only for `#for` and `#if`.

I think it is doable to manage that with snippet but it need some investigation. Perhaps it will be easier to manage that without snippet.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/commons/snippets/Snippet.java
```diff
@@ -95,6 +97,14 @@ public void setSortText(String sortText) {
 		this.sortText = sortText;
 	}
 
+	public String getParent() {
```
angelozerr, line 100: Please dont update classes from commons with qute specific feature.
For managing specific data like parent you must implemt ISnippetContext and déclare in json à context which have à parent. See sample with lsp4mp with context/type

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/commons/snippets/SnippetDeserializer.java
```diff
@@ -130,6 +131,13 @@ public Snippet deserialize(JsonElement json, Type typeOfT, JsonDeserializationCo
 			snippet.setSortText(sortText);
 		}
 
+		// parent
+		JsonElement parentElt = snippetObj.get(PARENT_ELT);
```
angelozerr, line 135: Please dont update commons class with specific qute feature. See LSP4MP  sample

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/commons/snippets/SnippetRegistry.java
```diff
@@ -43,11 +43,13 @@
 import com.google.gson.JsonSyntaxException;
 import com.google.gson.TypeAdapter;
 import com.google.gson.stream.JsonReader;
+import com.redhat.qute.parser.template.Node;
```
angelozerr, line 46: Please dont update commons class with qute specific feature

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/commons/snippets/SnippetRegistry.java
```diff
@@ -303,10 +308,15 @@ private static MarkupContent createDocumentation(Snippet snippet, Map<String, St
 	}
 
 	private static String getInsertText(Snippet snippet, Map<String, String> model, boolean replace,
-			String lineDelimiter) {
+			String lineDelimiter, Node node) {
 		StringBuilder text = new StringBuilder();
 		int i = 0;
+
+		// configure completion when nested in if section
+		boolean hasParent = node.getParent().getKind() == NodeKind.Section && snippet.getParent() != null;
```
angelozerr, line 316: Please dont update çommons cl1ss with qute geature. To check that see https://github.com/eclipse/lsp4mp/blob/master/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/snippets/SnippetContextForJava.java isMatch method. You need to have this similar class but for qute

qute.ls/com.redhat.qute.ls/src/main/resources/com/redhat/qute/services/snippets/qute-snippets.json
```diff
@@ -54,6 +59,13 @@
 			"\t$0",
 			"{/if}"
 		],
+		"nestedBody": [
```
angelozerr, line 62: Please dont update this file

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForSnippets.java
```diff
@@ -108,7 +108,7 @@ public void collectSnippetSuggestions(CompletionRequest completionRequest, Strin
 							}
 						}
 						return null;
-					}, suffixToFind, prefixFilter);
+					}, suffixToFind, prefixFilter, node);
```
angelozerr, line 111: I think this Java file must not be updated.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
```diff
@@ -0,0 +1,91 @@
...
+	private String parent;
+
+	public QuteParentSnippetContext(String parent) {
+		this.parent = parent;
+	}
+
+	@Override
+	public boolean isMatch(CompletionRequest request, Map<String, String> model) {
+		Node node = request.getNode();
+		if node.getParent().
```
angelozerr, line 44: compilation problem?

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
```diff
@@ -0,0 +1,91 @@
...
+			if (nextToken == JsonToken.NULL) {
+				return null;
+			}
+
+			in.beginObject();
+			while (in.hasNext()) {
+				String name = in.nextName();
+				switch (name) {
+				case "parent":
+					if (in.peek() == JsonToken.BEGIN_ARRAY) {
```
angelozerr, line 69: "parent"  in your JSON file snippet is not an array, just a simple string.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
AlexXuChen: Should be fixed now

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
```diff
@@ -0,0 +1,83 @@
...
+	private String parent;
+
+	public QuteParentSnippetContext(String parent) {
+		this.parent = parent;
+	}
+
+	@Override
+	public boolean isMatch(CompletionRequest request, Map<String, String> model) {
+		Node node = request.getNode();
+		if (node.getKind() == NodeKind.Section && node.getParentSection().getTag() == parent) {
```
angelozerr, line 44: There is several problem with `node.getParentSection().getTag() == parent:`

 * node.getParentSection() can be null
 * node.getParentSection().getTag() can be null
 * equals should be used to compare String instead of ==
 * you don't take care of parent of parent, etc

I suggest that you move that in a private static method `isMatchParent(Node node, String parentTag)` :

```java
if (node.getKind() == NodeKind.Section) {
    Section section = (Section) node;
    if (!section.isInStartTagName(request.getOffset())) {
        return false;
    }
    return isMatchParent(node, this.parent);
}
```

isMatchParent could looks like this (not tested):

```
public static boolean isMatchParent(Node node, String parentTag) {
    Section parent = node.getParentSection();
    while (parent != null) {
        if (parentTag.equals(parent.getTag()) {
             return true;
        }
        parent = parent.getParentSection();
    }
    return false;
}
```

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
```diff
@@ -0,0 +1,90 @@
...
+import com.google.gson.stream.JsonToken;
+import com.google.gson.stream.JsonWriter;
+import com.redhat.qute.parser.template.Node;
+import com.redhat.qute.parser.template.NodeKind;
+import com.redhat.qute.parser.template.Section;
+import com.redhat.qute.services.completions.CompletionRequest;
+
+/**
+ * Snippet context for Qute sections that have parent sections.
+ *
```
angelozerr, line 28: It should be nice to add a JSON sample.

angelozerr (changes requested):

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+public class QuteParentSnippetContext implements IQuteSnippetContext {
```
angelozerr, line 31: `implements IQuteSnippetContext<CompletionRequest>`

qute.ls/com.redhat.qute.ls/src/main/resources/com/redhat/qute/services/snippets/qute-nested-snippets.json
```diff
@@ -0,0 +1,28 @@
+{
```
angelozerr, line 1: Please format your JSON with vscode.

qute.ls/com.redhat.qute.ls/src/main/resources/com/redhat/qute/services/snippets/qute-nested-snippets.json
```diff
@@ -0,0 +1,28 @@
...
+            "parent": "if"
+        }
+    },
+    "#elseif": {
+		"prefix": "elseif",
+		"body": [
+			"\t$2",
+			"{#else if ${3:condition}}",
+			"\t$4",
+			"{#else}",
```
angelozerr, line 20: I think we should remove this else, because if it already an else in the template, it will report an error.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
AlexXuChen: I believe we only need to check the immediate parent, and not the nested one:

e.g. say we have
```
{#if condition1}
    {#if condition2}
        |
    {/if}
{/if}
```
the completion should only be for the if with condition2, I believe the if with condition1 doesn't matter here.

angelozerr: > the completion should only be for the if with condition2, I believe the if with condition1 doesn't matter here.

Indeed you are right. Please update my suggestion with isMatchParent

AlexXuChen:
It seems to be working, however there are 2 snippets that appear for if:
Peek 2022-06-16 16-52
Is this behaviour desired?

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
AlexXuChen: should IQuteSnippetContext be paramaterized, since ISnippetContext is paramatarized?

angelozerr:
> Is this behaviour desired?

Indeed it was my fear. I think it is annoying. 

I suggest that you rename the qute-snippet.json 

 * `else` with `ifelse`
 * `elseif` with `ifelseif`

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
```diff
@@ -0,0 +1,92 @@
...
+			return null;
+		}
+
+		@Override
+		public void write(JsonWriter out, QuteParentSnippetContext value) throws IOException {
+			// Do nothing
+		}
+	}
+
+	private boolean isMatchParent(Node node) {
```
angelozerr, line 88: Add javadoc

qute.ls/com.redhat.qute.ls/src/main/resources/com/redhat/qute/services/snippets/qute-nested-snippets.json
```diff
@@ -0,0 +1,26 @@
...
+        ],
+        "description": "Else section",
+        "context": {
+            "parent": "if"
+        }
+    },
+    "#elseif": {
+        "prefix": "elseif",
+        "body": [
+            "\t$2",
```
angelozerr, line 17: I think you can remove this line, no?

qute.ls/com.redhat.qute.ls/src/main/resources/com/redhat/qute/services/snippets/qute-nested-snippets.json
```diff
@@ -0,0 +1,26 @@
...
+            "{#else if ${3:condition}}",
```
angelozerr, line 18: `"{#else if ${1:condition}}",`

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Snippet context for Qute sections that have parent sections.
+ *
+ * e.g.
+ *
+ * "context": { "parent": "if" }
+ *
+ */
+
+public class QuteParentSnippetContext implements IQuteSnippetContext<CompletionRequest> {
```
angelozerr, line 35: Ooops no sorry, please keep like before  `implements IQuteSnippetContext`

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/IQuteSnippetContext.java
```diff
@@ -18,6 +18,6 @@
  * Qute snippet context API.
  *
  */
-public interface IQuteSnippetContext extends ISnippetContext<CompletionRequest> {
+public interface IQuteSnippetContext<T> extends ISnippetContext<CompletionRequest> {
```
angelozerr, line 21: Ooops no sorry, please keep like before  `IQuteSnippetContext extends ISnippetContext<CompletionRequest>`

qute.ls/com.redhat.qute.ls/src/main/resources/com/redhat/qute/services/snippets/qute-nested-snippets.json
AlexXuChen: Same as above

qute.ls/com.redhat.qute.ls/src/main/resources/com/redhat/qute/services/snippets/qute-nested-snippets.json
```diff
@@ -0,0 +1,26 @@
+{
+    "#else": {
+        "prefix": "else",
+        "body": [
+            "\t$2",
```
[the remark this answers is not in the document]
angelozerr: I think space before is annoying. Itis the reason why I suggest to remove `"\t$2"`

qute.ls/com.redhat.qute.ls/src/main/resources/com/redhat/qute/services/snippets/qute-nested-snippets.json
```diff
@@ -0,0 +1,22 @@
...
+        ],
+        "description": "Else section",
+        "context": {
+            "parent": "if"
+        }
+    },
+    "#elseif": {
+        "prefix": "elseif",
+        "body": [
+            "{#else if ${1:condition}}"
```
angelozerr, line 15: Update with `"{#else if ${1:condition}}$0"`

qute.ls/com.redhat.qute.ls/src/main/resources/com/redhat/qute/services/snippets/qute-nested-snippets.json
```diff
@@ -0,0 +1,22 @@
+{
+    "#else": {
+        "prefix": "else",
+        "body": [
+            "{#else}"
```
angelozerr, line 5: Update with `"{#else}$0"`

angelozerr (changes requested):

angelozerr:
Please write tests.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * "context": { "parent": "if" }
+ *
+ */
+
+public class QuteParentSnippetContext implements IQuteSnippetContext {
+
+	public static final TypeAdapter<QuteParentSnippetContext> IN_PARENT = new SnippetContextForQuteParentAdapter();
+
+	private String parent;
```
angelozerr, line 39: `private final String parent;`

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/completions/QuteSnippetCompletionInIfSectionTest.java
```diff
@@ -0,0 +1,88 @@
...
+								"	$4" + System.lineSeparator() + //
+								"{#else}" + System.lineSeparator() + //
+								"	$0" + System.lineSeparator() + //
+								"{/if}", //
+						r(0, 0, 0, 0)));
+	}
+
+	@Test
+	public void nestedElse() throws Exception {
+		String template = "{#if condition}" + System.lineSeparator() + //
```
angelozerr, line 70: Please dont use System.lineSeparator but hard code in the string the separator
"#if ${1:condition}}\n" + //

It is easier to read

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/completions/QuteSnippetCompletionInIfSectionTest.java
```diff
@@ -0,0 +1,88 @@
...
+				"|" + System.lineSeparator() + //
+				"{/if}";
+		testCompletionFor(template, //
+				true, // snippet support
```
angelozerr, line 74: It is important to check that no future nested section will appear in the completion.
To do that you must check the completion items size is the number of snippets + your 2 nested snippets.

See à sample at https://github.com/redhat-developer/quarkus-ls/blob/b5a04317e0363a33b08303d1635ef1a31eb54f30/qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/completions/QuteCompletionInSectionTagTest.java#L123

I suggest you remove your last test and move the check of elseif in this test

angelozerr: In other words:

```java
@Test
public void nestedIf() throws Exception {
String template = "{#if condition}\n" + //
		"|\n"  + //
		"{/if}";
testCompletionFor(template, //
		true, // snippet support,
		SECTION_SNIPPET_SIZE /* #each, #if, ... */+ 2 /* #else, #else if*/,
		c("else", "{#else}$0", r(1, 0, 1, 0)), //
		c("elseif", "{#else if ${1:condition}}$0", r(1, 0, 1, 0)));
}
```

And this test is nice because it higlight a bug here https://github.com/redhat-developer/quarkus-ls/pull/656/files#diff-f8df902c2cf37805c3a21be3609ea2b71e36e8f4db6db3f6f5f3f41594e50dd4R53 indeed the node where completion was triggered is a Text and not a Section, you should support the both. And write an another test like 

```java
@Test
public void nestedIf() throws Exception {
String template = "{#if condition}\n" + //
		"{#el|\n"  + //
		"{/if}";
testCompletionFor(template, //
		true, // snippet support,
		SECTION_SNIPPET_SIZE /* #each, #if, ... */+ 2 /* #else, #else if*/,
		c("else", "{#else}$0", r(1, 0, 1, 0)), //
		c("elseif", "{#else if ${1:condition}}$0", r(1, 0, 1, 0)));
}
```

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/completions/QuteSnippetCompletionInIfSectionTest.java
```diff
@@ -0,0 +1,88 @@
...
+		String template = "{#if condition}" + System.lineSeparator() + //
+				"{#|}" + System.lineSeparator() + //
+				"{/if}";
+		testCompletionFor(template, //
+				true, // snippet support
+				c("else", "{#else}$0", r(1, 0, 1, 3)));
+	}
+
+	@Test
+	public void nestedElseIf() throws Exception {
```
angelozerr, line 79: Please remove this test

angelozerr:
Please write a second test with only | like you did before and you will see
that it doesnt work
In isMatch the node can be a text.

AlexXuChen:
> Please write a second test with only | like you did before and you will see that it doesnt work In isMatch the node can be a text.
Text nodes should work now

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/completions/QuteSnippetCompletionInIfSectionTest.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+public class QuteSnippetCompletionInIfSectionTest {
+
+	@Test
+	public void singleIf() throws Exception {
+		String template = "|";
+		testCompletionFor(template, //
+				true, // snippet support
+				c("if", //
+						"{#if ${1:condition}}\n" + //
```
angelozerr, line 33: This test fails on Windows OS.

As original template content is `""` it doesn't define some line separator, in this case, the line separator comes from OS. So you need to use System.lineSeparator(). Please replace all \n with System.lineSeparator()

```
"{#if ${1:condition}}" + System.lineSperataor() + //
...

```

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/completions/QuteSnippetCompletionInIfSectionTest.java
```diff
@@ -0,0 +1,92 @@
...
+						r(0, 0, 0, 0)));
+	}
+
+	@Test
+	public void singleElse() throws Exception {
+		String template = "|";
+		testCompletionFor(template, //
+				true, // snippet support
+				c("if-else", //
+						"{#if ${1:condition}}\n" + //
```
angelozerr, line 45: As below, replace `\n` with `System.lineSpeparator()`

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/completions/QuteSnippetCompletionInIfSectionTest.java
```diff
@@ -0,0 +1,92 @@
...
+						r(0, 0, 0, 0)));
+	}
+
+	@Test
+	public void singleElseIf() throws Exception {
+		String template = "|";
+		testCompletionFor(template, //
+				true, // snippet support
+				c("if-elseif", //
+						"{#if ${1:condition}}\n" + //
```
angelozerr, line 59: As below, replace `\n` with `System.lineSpeparator()`

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/completions/QuteSnippetCompletionInIfSectionTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+	@Test
+	public void snippetCompletionInNestedIf() throws Exception {
+		String template = "{#if condition}\n" + //
+				"|\n" + //
+				"{/if}";
+		testCompletionFor(template, //
+				true, // snippet support
+				SECTION_SNIPPET_SIZE /* #each, #if, ... */ + 2, /* #else #else if */ //
+				c("else", "{#else}$0", r(1, 0, 1, 0)), //
```
angelozerr, line 77: Your test is correct, but just for your information, if your snippet #else will have a body with several items, it should generate lien separator. In this case it will be \n which should be used because the original template defines a \n

angelozerr:
Many thanks @AlexXuChen !

PR #661, opened 2022-06-20
Improved completion when nested in Qute `for`, `each`, `switch` and `when` sections

### LINKED ISSUE #497: Completion for nested block section

AlexXuChen:
Improved completion when nested in Qute `for`, `each`, `switch` and `when` sections.

Fixes #497 

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/AbstractQuteSnippetContext.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/AbstractQuteSnippetContext.java
@@ -1,0 +1,25 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.qute.services.snippets;
+
+import com.redhat.qute.ls.commons.snippets.AbstractSnippetContext;
+import com.redhat.qute.services.completions.CompletionRequest;
+
+/**
+ * Abstract Qute snippet context.
+ *
+ * @param <T> the value type waited by the snippet context.
+ */
+public abstract class AbstractQuteSnippetContext extends AbstractSnippetContext<CompletionRequest>
+		implements IQuteSnippetContext {
+
+}
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
@@ -28,18 +28,20 @@ import com.redhat.qute.services.completions.CompletionRequest;
  *
  * e.g.
  *
- * "context": { "parent": "if" }
+ * "context": { "parent" : "for", "unique" : true }
  *
  */
 
-public class QuteParentSnippetContext implements IQuteSnippetContext {
+public class QuteParentSnippetContext extends AbstractQuteSnippetContext {
 
 	public static final TypeAdapter<QuteParentSnippetContext> IN_PARENT = new SnippetContextForQuteParentAdapter();
 
 	private final String parent;
+	private final boolean unique;
 
-	public QuteParentSnippetContext(String parent) {
+	public QuteParentSnippetContext(String parent, boolean unique) {
 		this.parent = parent;
+		this.unique = unique;
 	}
 
 	@Override
@@ -50,7 +52,7 @@ public class QuteParentSnippetContext implements IQuteSnippetContext {
 		// {#if ...}
 		// | --> here {#else} completion is shown, the cursor is nested in #if parent.
 		// {/if}
-		if (isMatchParent(node)) {
+		if (isMatchParent(node) && isMatchUnique(node)) {
 			if ((node.getKind() == NodeKind.Template || node.getKind() == NodeKind.Text)) {
 				return true;
 			} else if (node.getKind() == NodeKind.Section) {
@@ -69,14 +71,20 @@ public class QuteParentSnippetContext implements IQuteSnippetContext {
 			if (nextToken == JsonToken.NULL) {
 				return null;
 			}
-
+			String parent = null;
+			boolean unique = false;
 			in.beginObject();
 			while (in.hasNext()) {
 				String name = in.nextName();
 				switch (name) {
 				case "parent":
 					if (in.peek() == JsonToken.STRING) {
-						return new QuteParentSnippetContext(in.nextString());
+						parent = in.nextString();
+					}
+					break;
+				case "unique":
+					if (in.peek() == JsonToken.BOOLEAN) {
+						unique = in.nextBoolean();
 					}
 					break;
 				default:
@@ -84,6 +92,9 @@ public class QuteParentSnippetContext implements IQuteSnippetContext {
 				}
 			}
 			in.endObject();
+			if (parent != null) {
+				return new QuteParentSnippetContext(parent, unique);
+			}
 			return null;
 		}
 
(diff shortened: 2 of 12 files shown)


qute.ls/com.redhat.qute.ls/src/main/resources/com/redhat/qute/services/snippets/qute-nested-snippets.json
```diff
@@ -1,14 +1,34 @@
...
         }
     },
+    "#for-else": {
+        "prefix": "else",
+        "body": [
+            "{#else}$0"
+        ],
+        "description": "Else section for loop section",
+        "context": {
+            "parent": "for"
```
angelozerr, line 29: As we can have just one #else inside #for, I suggest that you write context with 

```json
"context": {
            "parent": "for",
            "unique": true
}
```

AlexXuChen: The same applies to `if`, should we add it as well?

angelozerr: I dont know I suggest that you test that by writing and parse a qute template

AlexXuChen:
Still need to add `unique` check

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
```diff
@@ -75,8 +77,9 @@ public QuteParentSnippetContext read(final JsonReader in) throws IOException {
 				String name = in.nextName();
 				switch (name) {
 				case "parent":
+				case "unique":
```
angelozerr, line 80: I think your code should look like this:

```java
String parent = null;
boolean unique = false;
while (in.hasNext()) {
		String name = in.nextName();
		switch (name) {
		case "parent":
			if (in.peek() == JsonToken.STRING) {
				parent = in.nextString();
			}
			break;
		case "unique":
			if (in.peek() == JsonToken.BOOLEAN) {
				unique = in.nextBoolean();
			}
			break;
		default:
			in.skipValue();
		}
	}
}
if (parent != null) {
    return new QuteParentSnippetContext(parent, unique);
}
return null;
```

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
```diff
@@ -69,21 +73,36 @@ public QuteParentSnippetContext read(final JsonReader in) throws IOException {
...
-						return new QuteParentSnippetContext(in.nextString());
+						parent = in.nextString();
+					}
+					break;
+				case "unique":
+					if (in.peek() == JsonToken.BOOLEAN) {
+						unique = in.nextBoolean();
+					}
+					break;
+				case "tag":
```
AlexXuChen, line 93: I'm looking for a better way to retrieve `prefix` from `QuteParentSnippetContext.isMatch`, since the `tag` field is just a repetition.

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/completions/QuteSnippetCompletionInWhenSectionTest.java
```diff
@@ -0,0 +1,60 @@
...
+import static com.redhat.qute.QuteAssert.testCompletionFor;
+
+import org.junit.jupiter.api.Test;
+
+/**
+ * Tests for Qute snippet completion in #when section.
+ *
+ */
+public class QuteSnippetCompletionInWhenSectionTest {
+
```
datho7561, line 26: it might also be interesting to test the case where you have:
```
{#when value}
  {#is asdf}
  |
{/when}
```
and you open completion at the cursor, since you should be able to complete it many times (this case appears to work as expected right now).

datho7561 (commented):
Seems to be working pretty well from my testing

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
angelozerr: Indeed it should be better to duplicate this tag label.

I suggest that you refactor commons Snippet classes to have that:

```
public interface ISnippetContext<T> {

	void setPrefixes(List<String> prefixes);
	
	List<String> getPrefixes();
```

In Snippet class, update setContext to fill the prefixes of the snippet:

```java
	public void setContext(ISnippetContext<?> context) {
		this.context = context;
		this.context.setPrefixes(getPrefixes());
	}

```

To help implementing ISnippetContext, create the abstract class like:

```java
package com.redhat.qute.ls.commons.snippets;

import java.util.List;

public abstract class AbstractSnippetContext<T> implements ISnippetContext<T> {

	private List<String> prefixes;

	@Override
	public void setPrefixes(List<String> prefixes) {
		this.prefixes = prefixes;
	}

	@Override
	public List<String> getPrefixes() {
		return prefixes;
	}

}

```

Create a Qute abstract class like 

```
package com.redhat.qute.services.snippets;

import com.redhat.qute.ls.commons.snippets.AbstractSnippetContext;
import com.redhat.qute.services.completions.CompletionRequest;

public abstract class AbstractQuteSnippetContext extends AbstractSnippetContext<CompletionRequest>
		implements IQuteSnippetContext {

}
```

Update your Qute snippet contexnt by extening this class:

 * `public class QuteParentSnippetContext extends AbstractQuteSnippetContext {`
 * `public class QuteSnippetContext extends AbstractQuteSnippetContext {`

After that in thei sMatch method, you can access to the prefix with super.getPrefixes(). As you havejust one prefix you must just write

`String tag = super.getPrefixes().get(0);`

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
```diff
@@ -69,21 +73,36 @@ public QuteParentSnippetContext read(final JsonReader in) throws IOException {
...
+					if (in.peek() == JsonToken.STRING) {
+						tag = in.nextString();
```
angelozerr, line 95: Remove tag and use prefixes.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
```diff
@@ -103,4 +117,24 @@ private boolean isMatchParent(Node node) {
...
+	/**
+	 * Return true if unqiue context is true, and if parent node doesn't contain the
+	 * nested snippet, false otherwise.
+	 *
+	 * @param node the completion request node
+	 * @return true if unqiue context is true, and if parent node doesn't contain
+	 *         the nested snippet, false otherwise.
+	 */
+	private boolean isMatchUnique(Node node) {
+		Section parentSection = node.getParentSection();
```
angelozerr, line 129: if (!unique) {
    return true;
}

angelozerr:
CI build failing https://github.com/redhat-developer/quarkus-ls/runs/7049565848?check_suite_focus=true#step:9:4918

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
angelozerr: could you apply please this comment.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/AbstractQuteSnippetContext.java
```diff
@@ -0,0 +1,9 @@
+package com.redhat.qute.services.snippets;
```
angelozerr, line 1: Add header

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/AbstractQuteSnippetContext.java
```diff
@@ -0,0 +1,9 @@
...
+
+import com.redhat.qute.ls.commons.snippets.AbstractSnippetContext;
+import com.redhat.qute.services.completions.CompletionRequest;
+
+public abstract class AbstractQuteSnippetContext extends AbstractSnippetContext<CompletionRequest>
```
angelozerr, line 6: Add Javadoc

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
```diff
@@ -103,4 +117,27 @@ private boolean isMatchParent(Node node) {
...
+	 * @param node the completion request node
+	 * @return true if unqiue context is true, and if parent node doesn't contain
+	 *         the nested snippet, false otherwise.
+	 */
+	private boolean isMatchUnique(Node node) {
+		if (!unique) {
+			return true;
+		}
+		Section parentSection = node.getParentSection();
+		if (unique && parentSection != null) {
```
angelozerr, line 133: if (parentSection != null) {

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/snippets/QuteParentSnippetContext.java
```diff
@@ -28,18 +28,22 @@
...
  */
 
-public class QuteParentSnippetContext implements IQuteSnippetContext {
+public class QuteParentSnippetContext extends AbstractQuteSnippetContext {
 
 	public static final TypeAdapter<QuteParentSnippetContext> IN_PARENT = new SnippetContextForQuteParentAdapter();
 
 	private final String parent;
+	private final boolean unique;
+	private final String tag;
```
angelozerr, line 41: Remove this line code

angelozerr:
@AlexXuChen your test failed because of NPE problem, see my suggestion at https://github.com/redhat-developer/quarkus-ls/pull/661/files#r908216885

PR #666, opened 2022-06-29
Delay revalidation of Java files in qute-ls

datho7561:


qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/commons/RPCValidatorDelayer.java
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Validate a given text document with a delay.
+ *
+ * No model is passed to the validator, since it is assumed that the validation
+ * will be performed by a remote procedure that has access to the model already.
+ *
+ * @see {@ModelValidatorDelayer}
+ */
+public class RPCValidatorDelayer<T extends TextDocument> {
```
angelozerr, line 32: As there is similar code with`ModelValidatorDelayer`, I think it should be good to have an abstract class `AbstractValidatorDelayer` which manages the delayer feature and `ModelValidatorDelayer` and your class could extend it.

datho7561:
Thanks for the review!

PR #667, opened 2022-06-30
Added nested `else` completion support for `when` section

AlexXuChen:
Added nested `else` completion support for `when` section.
nested-else-when

Part of #497 

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/completions/QuteSnippetCompletionInWhenSectionTest.java
+++ b/qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/completions/QuteSnippetCompletionInWhenSectionTest.java
@@ -43,8 +43,9 @@ public class QuteSnippetCompletionInWhenSectionTest {
 				"{/when}";
 		testCompletionFor(template, //
 				true, // snippet support
-				SECTION_SNIPPET_SIZE /* #each, #if, ... */ + 1, /* #is */ //
-				c("is", "{#is ${1:case}}$0", r(1, 1, 1, 1)));
+				SECTION_SNIPPET_SIZE /* #each, #if, ... */ + 2, /* #is, #else */ //
+				c("is", "{#is ${1:case}}$0", r(1, 1, 1, 1)), //
+				c("else", "{#else}$0", r(1, 1, 1, 1)));
 	}
 
 	@Test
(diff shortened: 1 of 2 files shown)


qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/completions/QuteSnippetCompletionInWhenSectionTest.java
```diff
@@ -43,7 +43,7 @@ public void snippetCompletionInNestedWhen() throws Exception {
 				"{/when}";
 		testCompletionFor(template, //
 				true, // snippet support
-				SECTION_SNIPPET_SIZE /* #each, #if, ... */ + 1, /* #is */ //
+				SECTION_SNIPPET_SIZE /* #each, #if, ... */ + 2, /* #is, #else */ //
```
angelozerr, line 46: Add a c(....) like #is but wuth #else

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/completions/QuteSnippetCompletionInWhenSectionTest.java
```diff
@@ -54,19 +55,45 @@ public void isSnippetCompletionInNestedWhen() throws Exception {
...
+	@Test
+	public void elseSnippetCompletionInNestedWhenWithIs() throws Exception {
+		String template = "{#when value}\n" + //
+				"\t{#is case}\n" + //
+				"\t{#el|}\n" + //
+				"{/when}";
+		testCompletionFor(template, //
+				true, // snippet support
+				SECTION_SNIPPET_SIZE /* #each, #if, ... */ + 2, /* #is, #else */ //
+				c("else", "{#else}$0", r(2, 1, 2, 6)));
```
angelozerr, line 84: Add is here too

AlexXuChen: Since the completion is for "\t{#el|}\n", should there only be the "else" completion?

angelozerr: No, because we don't manage filter on server side. Teh filter is done in client side which manages fuzzy filter which is very nice.

angelozerr:
Thanks @AlexXuChen !

PR #668, opened 2022-07-05
Adapt to new version of m2e in JDT-LS

datho7561:
See https://github.com/eclipse/lsp4mp/pull/267, https://github.com/eclipse/lsp4mp/pull/268, https://github.com/eclipse/lsp4mp/pull/269


datho7561:
Okay. I think the tests fail unless we update to Java 17, though.

PR #671, opened 2022-07-08
CodeAction to create field, getter, or template extension for a missing property that's referenced in a template

### LINKED ISSUE #536: CodeActions to create Java field / getter method / template extension
Given this Qute template:

```
{@org.acme.Item item}
{item.XXX}
```

`XXX` property appears as an error. It should be nice to provide 3 code actions to fix the problem:

 * Create Java field `public String XXX` in `org.acme.Item` class.
 * Create Java getter method `public String getXXX() {return this.XXX}` in `org.acme.Item` class.
 * Create the template extension  `@TemplateExtension public String static getXXX(Item item)` in a new class or in an existing class which defines already some template extension.

datho7561:
Generates Java code for a property referenced in a Qute template that doesn't exist yet.

It can generate:
  *  A new public field, or update the visibility of an existing field
  *  A new getter, properly returning the field bearing the corresponding name if it exists
  *  A new template extension in the detected template extensions class
  *  A new template extension in a newly generated template extensions class

Fixes #536

### DIFF
--- a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/ls/QuteSupportForTemplateDelegateCommandHandler.java
+++ b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/ls/QuteSupportForTemplateDelegateCommandHandler.java
@@ -26,7 +26,10 @@ import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.lsp4j.Location;
+import org.eclipse.lsp4j.WorkspaceEdit;
 
+import com.redhat.qute.commons.GenerateMissingJavaMemberParams;
+import com.redhat.qute.commons.GenerateMissingJavaMemberParams.MemberType;
 import com.redhat.qute.commons.JavaTypeInfo;
 import com.redhat.qute.commons.ProjectInfo;
 import com.redhat.qute.commons.QuteJavaDefinitionParams;
@@ -44,7 +47,7 @@ import com.redhat.qute.jdt.QuteSupportForTemplate;
 
 /**
  * JDT LS commands used by Qute template.
- * 
+ *
  * @author Angelo ZERR
  *
  */
@@ -80,21 +83,25 @@ public class QuteSupportForTemplateDelegateCommandHandler extends AbstractQuteDe
 
 	private static final String QUTE_TEMPLATE_RESOLVED_JAVA_TYPE_COMMAND_ID = "qute/template/resolvedJavaType";
 
+	private static final String QUTE_TEMPLATE_GENERATE_MISSING_JAVA_MEMBER = "qute/template/generateMissingJavaMember";
+
 	@Override
 	public Object executeCommand(String commandId, List<Object> arguments, IProgressMonitor monitor) throws Exception {
 		switch (commandId) {
-		case QUTE_TEMPLATE_PROJECT_COMMAND_ID:
-			return getProjectInfo(arguments, commandId, monitor);
-		case QUTE_TEMPLATE_PROJECT_DATA_MODEL_COMMAND_ID:
-			return getProjectDataModel(arguments, commandId, monitor);
-		case QUTE_TEMPLATE_USER_TAGS_COMMAND_ID:
-			return getUserTags(arguments, commandId, monitor);
-		case QUTE_TEMPLATE_JAVA_TYPES_COMMAND_ID:
-			return getJavaTypes(arguments, commandId, monitor);
-		case QUTE_TEMPLATE_RESOLVED_JAVA_TYPE_COMMAND_ID:
-			return getResolvedJavaType(arguments, commandId, monitor);
-		case QUTE_TEMPLATE_JAVA_DEFINITION_COMMAND_ID:
-			return getJavaDefinition(arguments, commandId, monitor);
+			case QUTE_TEMPLATE_PROJECT_COMMAND_ID:
+				return getProjectInfo(arguments, commandId, monitor);
+			case QUTE_TEMPLATE_PROJECT_DATA_MODEL_COMMAND_ID:
+				return getProjectDataModel(arguments, commandId, monitor);
+			case QUTE_TEMPLATE_USER_TAGS_COMMAND_ID:
+				return getUserTags(arguments, commandId, monitor);
+			case QUTE_TEMPLATE_JAVA_TYPES_COMMAND_ID:
+				return getJavaTypes(arguments, commandId, monitor);
+			case QUTE_TEMPLATE_RESOLVED_JAVA_TYPE_COMMAND_ID:
+				return getResolvedJavaType(arguments, commandId, monitor);
+			case QUTE_TEMPLATE_JAVA_DEFINITION_COMMAND_ID:
+				return getJavaDefinition(arguments, commandId, monitor);
+			case QUTE_TEMPLATE_GENERATE_MISSING_JAVA_MEMBER:
+				return generateMissingJavaMember(arguments, commandId, monitor);
 		}
 		return null;
 	}
@@ -167,13 +174,13 @@ public class QuteSupportForTemplateDelegateCommandHandler extends AbstractQuteDe
 
 	/**
 	 * Collect user tags from a given project Uri.
-	 * 
+	 *
 	 * @param arguments
 	 * @param commandId
 	 * @param monitor
-	 * 
+	 *
 	 * @return user tags from a given project Uri.
-	 * 
+	 *
 	 * @throws CoreException
 	 */
 	private static List<UserTagInfo> getUserTags(List<Object> arguments, String commandId, IProgressMonitor monitor)
@@ -309,6 +316,62 @@ public class QuteSupportForTemplateDelegateCommandHandler extends AbstractQuteDe
 		boolean dataMethodInvocation = getBoolean(obj, DATA_METHOD_INVOCATION_ATTR);
 		params.setDataMethodInvocation(dataMethodInvocation);
 		return params;
+	}
+
+	private static WorkspaceEdit generateMissingJavaMember(List<Object> arguments, String commandId,
+			IProgressMonitor monitor)
+			throws JavaModelException, CoreException {
+
+		// Create java definition parameter
+		GenerateMissingJavaMemberParams params = createGenerateMissingJavaMemberParams(arguments, commandId);
+		// Return file information from the parameter
+		return QuteSupportForTemplate.getInstance().generateMissingJavaMember(params, JDTUtilsLSImpl.getInstance(),
+				monitor);
+	}
+
+	private static GenerateMissingJavaMemberParams createGenerateMissingJavaMemberParams(List<Object> arguments,
+			String commandId) {
+
+		Map<String, Object> obj = getFirst(arguments);
+		if (obj == null) {
+			throw new UnsupportedOperationException(String
+					.format("Command '%s' must be called with one QuteTemplateCodeActionParams argument!", commandId));
+		}
+
+		MemberType memberType = null;
+		try {
+			memberType = MemberType.forValue(ArgumentUtils.getInt(obj, "memberType"));
+		} catch (IllegalArgumentException e) {
+			// ignored
+		}
+		if (memberType == null) {
+			throw new UnsupportedOperationException(String.format(
+					"Command '%s' must be called with required GenerateMissingJavaMemberParams.memberType !",
+					commandId));
+		}
+
+		String missingProperty = getString(obj, "missingProperty");
+		if (missingProperty == null) {
+			throw new UnsupportedOperationException(String.format(
+					"Command '%s' must be called with required GenerateMissingJavaMemberParams.missingProperty !",
+					commandId));
+		}
+
+		String javaType = getString(obj, "javaType");
+		if (javaType == null) {
+			throw new UnsupportedOperationException(String.format(
+					"Command '%s' must be called with required GenerateMissingJavaMemberParams.javaType !", commandId));
+		}
+
+		String projectUri = getString(obj, "projectUri");
+		if (projectUri == null) {
+			throw new UnsupportedOperationException(String.format(
+					"Command '%s' must be called with required GenerateMissingJavaMemberParams.projectUri !",
+					commandId));
+		}
+
+		return new GenerateMissingJavaMemberParams(memberType, missingProperty, javaType, projectUri);
+
 	}
 
 }
--- a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/QuteSupportForTemplateGenerateMissingJavaMemberHandler.java
+++ b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/QuteSupportForTemplateGenerateMissingJavaMemberHandler.java
@@ -1,0 +1,683 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.qute.jdt.internal.template;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.jdt.core.Flags;
+import org.eclipse.jdt.core.IAnnotatable;
+import org.eclipse.jdt.core.IAnnotation;
+import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IField;
+import org.eclipse.jdt.core.IJavaElement;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.IPackageFragment;
+import org.eclipse.jdt.core.IPackageFragmentRoot;
+import org.eclipse.jdt.core.IType;
+import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.jdt.core.dom.AST;
+import org.eclipse.jdt.core.dom.ASTNode;
+import org.eclipse.jdt.core.dom.ASTParser;
+import org.eclipse.jdt.core.dom.ASTVisitor;
+import org.eclipse.jdt.core.dom.Block;
+import org.eclipse.jdt.core.dom.ChildListPropertyDescriptor;
+import org.eclipse.jdt.core.dom.CompilationUnit;
+import org.eclipse.jdt.core.dom.Expression;
+import org.eclipse.jdt.core.dom.FieldAccess;
+import org.eclipse.jdt.core.dom.FieldDeclaration;
+import org.eclipse.jdt.core.dom.MethodDeclaration;
+import org.eclipse.jdt.core.dom.Modifier;
+import org.eclipse.jdt.core.dom.Modifier.ModifierKeyword;
+import org.eclipse.jdt.core.dom.ReturnStatement;
+import org.eclipse.jdt.core.dom.SingleVariableDeclaration;
+import org.eclipse.jdt.core.dom.Type;
+import org.eclipse.jdt.core.dom.TypeDeclaration;
+import org.eclipse.jdt.core.dom.VariableDeclarationFragment;
+import org.eclipse.jdt.core.dom.rewrite.ASTRewrite;
+import org.eclipse.jdt.core.dom.rewrite.ListRewrite;
+import org.eclipse.jdt.core.manipulation.CodeGeneration;
+import org.eclipse.jdt.core.search.IJavaSearchConstants;
+import org.eclipse.jdt.core.search.IJavaSearchScope;
+import org.eclipse.jdt.core.search.SearchEngine;
+import org.eclipse.jdt.core.search.SearchMatch;
+import org.eclipse.jdt.core.search.SearchParticipant;
+import org.eclipse.jdt.core.search.SearchPattern;
+import org.eclipse.jdt.core.search.SearchRequestor;
+import org.eclipse.jdt.internal.core.JavaModelManager;
+import org.eclipse.jdt.internal.core.manipulation.StubUtility;
+import org.eclipse.jdt.internal.core.manipulation.dom.ASTResolving;
+import org.eclipse.jdt.internal.corext.dom.ASTNodes;
+import org.eclipse.jdt.internal.corext.dom.IASTSharedValues;
+import org.eclipse.lsp4j.CreateFile;
+import org.eclipse.lsp4j.Position;
+import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4j.ResourceOperation;
+import org.eclipse.lsp4j.TextDocumentEdit;
+import org.eclipse.lsp4j.TextEdit;
+import org.eclipse.lsp4j.VersionedTextDocumentIdentifier;
+import org.eclipse.lsp4j.WorkspaceEdit;
+import org.eclipse.lsp4j.jsonrpc.messages.Either;
+
+import com.redhat.qute.commons.GenerateMissingJavaMemberParams;
+import com.redhat.qute.jdt.internal.QuteJavaConstants;
+import com.redhat.qute.jdt.utils.IJDTUtils;
+import com.redhat.qute.jdt.utils.TextEditConverter;
+
+/**
+ * Resolves workspace edits for generating missing java members.
+ *
+ * @author datho7561
+ */
+public class QuteSupportForTemplateGenerateMissingJavaMemberHandler {
+
+	private static final Logger LOGGER = Logger
+			.getLogger(QuteSupportForTemplateGenerateMissingJavaMemberHandler.class.getName());
+
+	private static final Pattern BROKEN_FILE_PROTOCOL = Pattern.compile("file:/(?!//)");
+
+	private static final Range PREPEND_RANGE = new Range(new Position(0, 0), new Position(0, 0));
+
+	private QuteSupportForTemplateGenerateMissingJavaMemberHandler() {
+
+	}
+
+	/**
+	 * Returns the WorkspaceEdit needed to generate the requested member, or null if
+	 * it cannot be computed.
+	 *
+	 * @param params  the parameters needed to construct the workspace edit
+	 * @param utils   the jdt utils
+	 * @param monitor the progress monitor
+	 * @return the WorkspaceEdit needed to generate the requested member, or null if
+	 *         it cannot be computed
+	 */
+	public static WorkspaceEdit handleGenerateMissingJavaMember(GenerateMissingJavaMemberParams params, IJDTUtils utils,
+			IProgressMonitor monitor) {
+		switch (params.getMemberType()) {
+		case Field:
+			return handleMissingField(params, utils, monitor);
+		case Getter:
+			return handleCreateMissingGetterCodeAction(params, utils, monitor);
+		case AppendTemplateExtension:
+			return handleCreateMissingTemplateExtension(params, utils, monitor);
+		case CreateTemplateExtension:
+			return createNewTemplateExtensionsFile(params, utils, monitor);
+		default:
+			return null;
+		}
+	}
+
+	private static WorkspaceEdit handleMissingField(GenerateMissingJavaMemberParams params, IJDTUtils utils,
+			IProgressMonitor monitor) {
+		IJavaProject project = getJavaProjectFromProjectUri(params.getProjectUri());
+		IType javaType;
+		try {
+			javaType = project.findType(params.getJavaType());
+		} catch (JavaModelException e) {
+			return null;
+		}
+
+		IField currentlyField = javaType.getField(params.getMissingProperty());
+		if (currentlyField != null && currentlyField.exists()) {
+			return handleUpdatePermissionsOfExistingField(params, utils, project, javaType, monitor);
+		} else {
+			return handleCreateMissingField(params, utils, project, javaType, monitor);
+		}
+
+	}
+
+	private static WorkspaceEdit handleCreateMissingField(GenerateMissingJavaMemberParams params, IJDTUtils utils,
+			IJavaProject project, IType javaType, IProgressMonitor monitor) {
+		CompilationUnit cu = createQuickFixAST(javaType);
+		if (cu == null) {
+			return null;
+		}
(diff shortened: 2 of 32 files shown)


qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/QuteTemplateCodeActionParams.java
```diff
@@ -0,0 +1,74 @@
...
+ * @author datho7561
+ */
+public class QuteTemplateCodeActionParams {
+
+	public static final String CREATE_MISSING_FIELD = "createMissingField";
+	public static final String CREATE_MISSING_GETTER = "createMissingGetter";
+	public static final String CREATE_MISSING_TEMPLATE_EXTENSION = "createMissingTemplateExtension";
+
+	private String id;
+	private String resolvedType;
```
angelozerr, line 26: I would rename with javaType

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/QuteTemplateCodeActionParams.java
```diff
@@ -0,0 +1,74 @@
...
+	private String missingProperty;
+	private String projectUri;
+
+	public QuteTemplateCodeActionParams() {
+		this(null, null, null, null);
+	}
+
+	public QuteTemplateCodeActionParams(String id, String projectUri, String resolvedType,
```
angelozerr, line 34: I would sort parameters with id missingProperty javaType and projectUri

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/QuteTemplateCodeActionParams.java
```diff
@@ -0,0 +1,74 @@
...
+
+package com.redhat.qute.commons;
+
+/**
+ * Parameters for a request from qute-ls to the JDT component to resolve a
+ * CodeAction that requires knowledge of the Java project.
+ *
+ * @author datho7561
+ */
+public class QuteTemplateCodeActionParams {
```
angelozerr, line 19: I would not try to do a generic codeaction for template in jdt side because your code action parameters could grow with new parameters like deleteJavaFileUri to delete a java file.  It will be hard to unterstand your code action parameters. 

I suggest that you keep your parameyer like this but rename your class like GenerateJavaMemberParameters ans replace in jdt side codeaction with your new name. More on jdt you return à workspaceEdit and not à codeaction so it is better to use a name like GenerateJavaMember instead of codeaction. Fell free to use ana noter name but more specific than codeaction.

angelozerr:
I think the direction is good but please rénale codeaction with other name like generateJavaElements or generateJavaMambers or an another name.

It seems it misses the pr on vscode quzrkus to bind the request with jdt command

datho7561:
Thanks for the review, Angelo! It was really helpful.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/ls/QuteSupportForTemplateDelegateCommandHandler.java
```diff
@@ -311,4 +318,27 @@ private static QuteJavaDefinitionParams createQuteJavaDefinitionParams(List<Obje
...
+
+	private static GenerateMissingJavaMemberParams createGenerateMissingJavaMemberParams(List<Object> arguments,
+			String commandId) {
+
+		Map<String, Object> obj = getFirst(arguments);
+		if (obj == null) {
+			throw new UnsupportedOperationException(String
+					.format("Command '%s' must be called with one QuteTemplateCodeActionParams argument!", commandId));
+		}
+		return GsonInstance.getInstance().fromJson(GsonInstance.getInstance().toJson(obj),
```
angelozerr, line 340: Have you the intention do do tha for other create method because it is not consistent with other create method.

datho7561: I am planning on converting it over to the same strategy as the other methods. I did this because I got lazy when I was coding and wanted to focus on the JDT code and the interface.

datho7561:
See the needed client change at redhat-developer/vscode-quarkus#520

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
```diff
@@ -701,8 +702,14 @@ private ResolvedJavaTypeInfo validatePropertyPart(PropertyPart part, Section own
 			// ex : {@org.acme.Item item}
 			// "{item.XXXX}
 			Range range = QutePositionUtility.createRange(part);
+			String signature = baseType.getSignature();
+			String property = part.getPartName();
 			Diagnostic diagnostic = createDiagnostic(range, DiagnosticSeverity.Error, QuteErrorCode.UnknownProperty,
-					part.getPartName(), baseType.getSignature());
+					property, signature);
+			Map<String, String> data = new HashMap<>();
```
angelozerr, line 709: I suggest that you do like Lemminx 

 * use JsonObject instead of map in order to support eclipse ide because iy works like this
 * * use an utility class to create the data that you need to store and get it

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -100,39 +105,89 @@ public CompletableFuture<List<CodeAction>> doCodeActions(Template template, Code
...
+				projectUri);
+		GenerateMissingJavaMemberParams getterParams = new GenerateMissingJavaMemberParams(
+				MemberType.Getter, missingProperty, resolvedType,
+				projectUri);
+		GenerateMissingJavaMemberParams templateExtensionParams = new GenerateMissingJavaMemberParams(
+				MemberType.TemplateExtension, missingProperty, resolvedType,
+				projectUri);
+
+		CodeAction createPublicField = createCodeActionWithData(
+				"Create public field \"" + missingProperty + "\" in " + resolvedType,
```
angelozerr, line 165: Use messageFormat to create error message

angelozerr:
Why you cannot use existing qute qickstart test project?

datho7561:
> Why you cannot use existing qute qickstart test project?

I need to have one project with a type that is annotated with `@TemplateExtensions` and one project that doesn't have any class annotated with `@TemplateExtensions` for my tests.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -133,6 +141,46 @@ public CompletableFuture<List<CodeAction>> doCodeActions(Template template, Code
...
+
+		GenerateMissingJavaMemberParams publicFieldParams = new GenerateMissingJavaMemberParams(MemberType.Field,
+				missingProperty, resolvedType, projectUri);
+		GenerateMissingJavaMemberParams getterParams = new GenerateMissingJavaMemberParams(MemberType.Getter,
+				missingProperty, resolvedType, projectUri);
+		GenerateMissingJavaMemberParams templateExtensionParams = new GenerateMissingJavaMemberParams(
+				MemberType.TemplateExtension, missingProperty, resolvedType, projectUri);
+
+		CodeAction createPublicField = createCodeActionWithData(
+				String.format("Create public field \"%s\" in %s", missingProperty, resolvedType), publicFieldParams,
```
angelozerr, line 161: Move the string message as private static field

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -133,6 +141,46 @@ public CompletableFuture<List<CodeAction>> doCodeActions(Template template, Code
...
+				Collections.singletonList(diagnostic));
+		CodeAction createGetter = createCodeActionWithData(
+				String.format("Create getter \"get%s()\" in %s", propertyCapitalized, resolvedType), getterParams,
+				Collections.singletonList(diagnostic));
+		CodeAction createTemplateExtension = createCodeActionWithData(String
+				.format("Create template extension \"%s()\" in detected template extensions class", missingProperty),
+				templateExtensionParams, Collections.singletonList(diagnostic));
+
+		try {
+			createPublicField.setEdit(resolver.resolveTemplateCodeAction(publicFieldParams).get());
```
angelozerr, line 171: Using .get could block the thread.  I would try to use .getNow(null) by checking if result != null

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -133,6 +141,46 @@ public CompletableFuture<List<CodeAction>> doCodeActions(Template template, Code
 		return CompletableFuture.completedFuture(codeActions);
 	}
 
+	private void doCodeActionToCreateProperty(Template template, Diagnostic diagnostic, QuteErrorCode errorCode,
+			List<CodeAction> codeActions, QuteTemplateGenerateMissingJavaMember resolver, SharedSettings settings) {
+		JsonObject data = (JsonObject) diagnostic.getData();
+		String missingProperty = data.get("property").getAsString();
```
angelozerr, line 147: If client doesnt support data, the data will be null. You must take care of this npe and get in this case the resolved type. For the last part you could do that in an another pr since vscode and eclipse ide support data

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -133,6 +141,46 @@ public CompletableFuture<List<CodeAction>> doCodeActions(Template template, Code
 		return CompletableFuture.completedFuture(codeActions);
 	}
 
+	private void doCodeActionToCreateProperty(Template template, Diagnostic diagnostic, QuteErrorCode errorCode,
+			List<CodeAction> codeActions, QuteTemplateGenerateMissingJavaMember resolver, SharedSettings settings) {
+		JsonObject data = (JsonObject) diagnostic.getData();
+		String missingProperty = data.get("property").getAsString();
+		String resolvedType = data.get("signature").getAsString();
```
angelozerr, line 148: Dont use string for property like "signature" but declare a static method where you create the JsonObject. See a sample inlemminx

angelozerr:
> I need to have one project with a type that is annotated with @TemplateExtensions and one project that doesn't have any class annotated with @TemplateExtensions for my tests.

I wonder if you could provide your code action which create the template extension   class even if there are a template extension in the class.

If you do that you dont need a new test project. I would like to avoid having à lot of test project.

angelozerr:
@datho7561 @AlexXuChen I have the impression that you need the same feature : get the java type signature from the basé type when a property or method is wrong. Storing the signature in data is a good idea but please keep in minid that  some client doesnt support diagnostic data support.

We could take care of this usecase in an another pr.

We should perhzps having à codeactioneequest which provides à method like

 ResolvedJavaType getJavaType (Duagnostic diagnostic)

I think more and more too that we should try to use a class with signature znd property fields instead of JsonObject and use JsonHelper to deserialize it.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
datho7561: I'll see if there is a way I can do this without blocking. Unfortunately, I can't use `.getNow(null)`, since the result will always be `null`.

datho7561:
> We could take care of this usecase in an another pr.

Sounds good, I'll make sure to open a PR for that once this is merged

> I think more and more too that we should try to use a class with signature znd property fields instead of JsonObject and use JsonHelper to deserialize it.

I'll add that in this PR, I think it will be helpful and make the code easier to read.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
datho7561: I tried a slightly different method to prevent blocking in QuteCodeActions. When you get a change, please let me know what you think about it.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/diagnostics/DiagnosticDataFactory.java
```diff
@@ -43,6 +43,13 @@ public static JsonObject createUndefinedSectionTagData(String tagName) {
 		return data;
 	}
 
+	public static JsonObject createMissingMemberData(String signature, String property) {
+		JsonObject data = new JsonObject();
+		data.addProperty("signature", signature);
```
angelozerr: Try to create a class like UnkwownProperryData with signature and property fields instead of using JsonObject

datho7561: sorry, I did that, I just forgot to squash the commit

angelozerr: I think UnkwownProperryData is more generic for nzme znd @AlexXuChen could use it for the similar codeaction

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -100,37 +114,115 @@ public CompletableFuture<List<CodeAction>> doCodeActions(Template template, Code
...
+	private void doCodeActionToCreateProperty(Template template, Diagnostic diagnostic, QuteErrorCode errorCode,
+			List<CodeAction> codeActions, List<CompletableFuture<Void>> registrations,
+			QuteTemplateGenerateMissingJavaMember resolver, SharedSettings settings) {
+		JsonObject data = (JsonObject) diagnostic.getData();
+		if (data == null) {
+			// TODO: the client may not support diagnostic data; resolve the needed
+			// information some other way
+			return;
+		}
+		MissingMemberData missingMemberData = DiagnosticDataFactory.getMissingMemberData(data);
```
angelozerr, line 169: I think it should be better to have

UnkwonwPropertyData data = DiagnosticDataFactory.getUnkwonwPropertyData(diagnostic, template, javaCache);

In this method you could manage the case when diagnostic has a data and when diagnostic has no the data. In this case  node part must be found by the diagnostic index and must be resolved by using javaCache

angelozerr: @AlexXuChen could use it this method too for similar code action.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
angelozerr: Yes it is better thanks. But I think we should update the edit in à resolve code action. We can do that in an another pr

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -100,37 +114,115 @@ public CompletableFuture<List<CodeAction>> doCodeActions(Template template, Code
...
+					if (workspaceEdit == null) {
+						return;
+					}
+					createPublicField.setEdit(workspaceEdit);
+					synchronized (codeActions) {
+						codeActions.add(createPublicField);
+					}
+				}));
+
+		registrations.add(resolver.resolveTemplateCodeAction(getterParams) //
```
angelozerr, line 204: I think it should be better that you create the codeaction without the edit and you add it in codeActions.add(createGetter

The thenApply will update the edit code action and that is it.

If you do that you will control the order of codeaction add in

datho7561:
> I wonder if you could provide your code action which create the template extension class even if there are a template extension in the class.

I think this is a good idea. I still need to set aside some time to implement this.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
datho7561: CodeActions don't have access to the java cache in this PR, so I didn't include that parameter

angelozerr: Yes sure, let's do that in the PR which will support no diagnostic data. For the moment you can do:

`UnkwonwPropertyData data = DiagnosticDataFactory.getUnkwonwPropertyData(diagnostic);`

datho7561:
I've separated the "Create new Template extensions class" and "Add to existing template extension class" code actions, and removed the extra test project.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/QuteSupportForTemplateGenerateMissingJavaMemberHandler.java
```diff
@@ -0,0 +1,668 @@
...
+		buf.append(name);
+		buf.append(" {").append(lineDelimiter); //$NON-NLS-1$
+		buf.append(constructMethodStub(parentCU, methodName, methodParamFullyQualifiedType, lineDelimiter));
+		buf.append('}').append(lineDelimiter);
+		return buf.toString();
+	}
+
+	private static String constructMethodStub(ICompilationUnit compilationUnit, String methodName,
+			String methodParamFullyQualifiedType, String lineDelimiter) {
+		StringBuffer buf = new StringBuffer("\tpublic static String ");
```
angelozerr, line 521: Use StringBuilder

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/QuteSupportForTemplateGenerateMissingJavaMemberHandler.java
```diff
@@ -0,0 +1,668 @@
...
+import org.eclipse.jdt.core.search.SearchEngine;
+import org.eclipse.jdt.core.search.SearchMatch;
+import org.eclipse.jdt.core.search.SearchParticipant;
+import org.eclipse.jdt.core.search.SearchPattern;
+import org.eclipse.jdt.core.search.SearchRequestor;
+import org.eclipse.jdt.internal.core.JavaModelManager;
+import org.eclipse.jdt.internal.core.manipulation.StubUtility;
+import org.eclipse.jdt.internal.corext.dom.ASTNodes;
+import org.eclipse.jdt.internal.corext.dom.IASTSharedValues;
+import org.eclipse.jdt.ls.core.internal.ChangeUtil;
```
angelozerr, line 68: We must have no dependency to jdt ls because jboss tools which integrate qute in eçlipse ide has no dependency to jdt ls.

The solution are

 * copy paste jdt ls ChangeUtils etc in qute jdt like I did in lsp4mp
 *   * provide an interface like ijdtutils to wrap implemntation of jdt ls. I think it was hard to do that.

datho7561: Okay. Sounds good. I found that after cleaning up some of the code, I didn't need ChangeUtils. Let me know if my additions to IJDTUtils make sense.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/utils/IJDTUtils.java
```diff
@@ -85,5 +86,26 @@ public interface IJDTUtils {
...
 
-	//String getJavadoc(IMember member, DocumentFormat documentFormat) throws JavaModelException;
+	/**
+	 * Returns the JDT text edit as an LSP text edit
+	 *
+	 * @param compulationUnit the compilation unit of the file that was edited
+	 * @param jdtTextEdit     the JDT text edit
+	 * @return the JDT text edit as an LSP text edit
+	 */
+	TextDocumentEdit convertEdit(ICompilationUnit compulationUnit, org.eclipse.text.edits.TextEdit jdtTextEdit);
```
angelozerr, line 96: After seeing the code of TextEditConverter of JDT LS, I think you can copy/paste the TextEditConverter and adapt it to use it with our own IJDTUtils. It will avoid for @jeffmaury to implement this method.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/utils/IJDTUtils.java
```diff
@@ -85,5 +86,26 @@ public interface IJDTUtils {
...
+	/**
+	 * Returns an AST that can be used to create a quick fix for the given
+	 * compilation unit.
+	 *
+	 * @param compliationUnit the compilation that a quick fix is being generated
+	 *                        for
+	 * @return an AST that can be used to create a quick fix for the given
+	 *         compilation unit
+	 */
+	CompilationUnit createQuickFixAST(ICompilationUnit compliationUnit);
```
angelozerr, line 107: createQuickFixAST implementation uses JDT manipulation and not JDT LS, so you can remove this method. It will avoid for @jeffmaury to implement this method.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/utils/IJDTUtils.java
```diff
@@ -21,8 +20,10 @@
 import org.eclipse.jdt.core.IOpenable;
 import org.eclipse.jdt.core.ITypeRoot;
 import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.jdt.core.dom.CompilationUnit;
```
angelozerr, line 23: Don't update this file.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/api/QuteTemplateGenerateMissingJavaMember.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Request additional information from the associated java language server in
+ * order to resolve the workspace edit to create a field, getter, or template
+ * extension for a given type.
+ *
+ * @author datho7561
+ */
+public interface QuteTemplateGenerateMissingJavaMember {
+	@JsonRequest("qute/template/generateMissingJavaMember")
+	default CompletableFuture<WorkspaceEdit> resolveTemplateCodeAction(GenerateMissingJavaMemberParams params) {
```
angelozerr, line 28: rename method name with `generateMissingJavaMember`

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -100,37 +116,120 @@ public CompletableFuture<List<CodeAction>> doCodeActions(Template template, Code
...
+		CodeAction appendToTemplateExtensions = createCodeActionWithData(String
+				.format(APPEND_TO_TEMPLATE_EXTENSIONS, missingProperty),
+				appendToTemplateExtensionsParams, Collections.singletonList(diagnostic));
+		CodeAction createTemplateExtensions = createCodeActionWithData(String.format(CREATE_TEMPLATE_EXTENSIONS, missingProperty), createTemplateExtensionsParams, Collections.singletonList(diagnostic));
+		codeActions.add(createPublicField);
+		codeActions.add(createGetter);
+		codeActions.add(appendToTemplateExtensions);
+		codeActions.add(createTemplateExtensions);
+
+		registrations.add(resolver.resolveTemplateCodeAction(publicFieldParams) //
```
angelozerr, line 201: You can do that in an another PR, but it should be really nice to support resolve code actions. Please create an issue for that.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/QuteSupportForTemplateGenerateMissingJavaMemberHandler.java
angelozerr: Thanks for the change.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/ls/JDTUtilsLSImpl.java
```diff
@@ -25,13 +25,18 @@
 import org.eclipse.jdt.core.IOpenable;
 import org.eclipse.jdt.core.ITypeRoot;
 import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.jdt.core.dom.CompilationUnit;
```
angelozerr, line 28: Please don't update this file.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/QuteSupportForTemplateGenerateMissingJavaMemberHandler.java
```diff
@@ -0,0 +1,671 @@
...
+		StringBuilder buf = new StringBuilder();
+		if (!pack.isDefaultPackage()) {
+			buf.append("package ").append(pack.getElementName()).append(';'); //$NON-NLS-1$
+		}
+		buf.append(lineDelimiter).append(lineDelimiter);
+		buf.append(typeContent);
+		return buf.toString();
+	}
+
+	private static CompilationUnit createQuickFixAST(ICompilationUnit compliationUnit) {
```
angelozerr, line 559: There is ClasCastException error when this method is called with binary like java.lang.String (it is a IClassFile and not a ICompilationUnit).

I think createQuickFixAST should return null if it is a binary (ex : we try to add a new getter, property in java.lang.String) and otherwise the CompilationUnit.

```java
CompilationUnit createQuickFixAST(IType javaType) {
   if (javaType.isBinary()) {
      return null;
   }
   return ASTResolving.createQuickFixAST((ICompilationUnit) javaType, null);
} 
```

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/QuteSupportForTemplateGenerateMissingJavaMemberHandler.java
```diff
@@ -0,0 +1,671 @@
...
+
+	private static WorkspaceEdit createNewTemplateExtensionsFile(GenerateMissingJavaMemberParams params,
+			IJDTUtils utils, IProgressMonitor monitor) {
+		IJavaProject project = getJavaProjectFromProjectUri(params.getProjectUri());
+		return createNewTemplateExtensionFile(params, utils, project, monitor);
+	}
+
+	private static WorkspaceEdit addTemplateExtensionToFile(GenerateMissingJavaMemberParams params, IJDTUtils utils,
+			IJavaProject project, IType templateExtensionType, IProgressMonitor monitor) {
+		CompilationUnit cu = createQuickFixAST((ICompilationUnit) templateExtensionType.getTypeRoot());
```
angelozerr, line 399:
```java
CompilationUnit cu = createQuickFixAST(templateExtensionType.getTypeRoot());
if (cu == null) {
  // It is a binary java type (ex : java.lang.String)
   // I think we should return a WorkspaceEdit which does nothing
   return ????
}
```

angelozerr:
@datho7561 after fixing my comments (we cannot create method/property for binary like java.lang.String), I'm OK to merge your PR, but it will require some improvements again (please one issue for that):

 * code actions should not appear if the Java type is a binary (the main idea is to add a Boolean (and not boolean) source field in ResolvedJavaType to know this information) like we did for LSP4MP.
 * resolve code actions should be implemented if client support it : the idea is to create code actions like you did without text edit, the text edit will be resolved on resolveCodeActions.
 * the getter generates a method which returns null. Indeally we should generate a private field (if it doesn't exists) which returns this field as getter.
 * we have one code action which generates the method in existing TemplateExtensions, but we should generate one code action per existing TemplateExtensions (which are not binary).

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
datho7561: https://github.com/redhat-developer/quarkus-ls/issues/604

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -81,10 +86,21 @@ class QuteCodeActions {
 
 	private static final String UNDEFINED_NAMESPACE_SEVERITY_SETTING = "qute.validation.undefinedNamespace.severity";
 
+	private static final String APPEND_TO_TEMPLATE_EXTENSIONS = "Create template extension \"%s()\" in detected template extensions class";
+	
+	private static final String CREATE_TEMPLATE_EXTENSIONS = "Create template extension \"%s()\" in a new template extensions class";
+
+	private static final String CREATE_GETTER = "Create getter \"get%s()\" in %s";
```
angelozerr, line 93: Use ` instead of `"` and ends with . your message to be consistent with other code action title:

image

angelozerr: It is not fixed, no?

datho7561:
Okay, I have created the issues you mentioned

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -81,10 +86,21 @@ class QuteCodeActions {
 
 	private static final String UNDEFINED_NAMESPACE_SEVERITY_SETTING = "qute.validation.undefinedNamespace.severity";
 
+	private static final String APPEND_TO_TEMPLATE_EXTENSIONS = "Create template extension `%s()` in detected template extensions class";
+	
+	private static final String CREATE_TEMPLATE_EXTENSIONS = "Create template extension `%s()` in a new template extensions class";
+
+	private static final String CREATE_GETTER = "Create getter `get%s()` in %s";
```
angelozerr, line 93: The second %s must be enclosed with tild. The message must be ended with dot. Do that for the other message.

angelozerr:
@datho7561 I think it misses the issues with redolve code actions which I think it is the more important for performance reason.

datho7561:
> @datho7561 I think it misses the issues with redolve code actions which I think it is the more important for performance reason.

https://github.com/redhat-developer/quarkus-ls/issues/604

angelozerr:
Great job @datho7561 ! Lets improve this feature in other issues.

@AlexXuChen you can rebase your PR to benefit with the feature of this pr

PR #679, opened 2022-07-28
Fix cancel support with CompletableFuture compose

angelozerr:
Fix cancel support with CompletableFuture compose


datho7561 (approved):
Works for me. Good catch! We should probably also copy this over to other language servers.

PR #686, opened 2022-07-28
Don't show "add field/getter" CodeActions in Qute for binary types

### LINKED ISSUE #676: Base "generate missing member" CodeActions on if the type is a source or binary file
The "add field" and "add getter" code actions for missing members in Qute templates should not appear if the Java type is a binary, since the source for the class can't be modified.

The "add to existing template extensions" and "add to new template extensions" code actions for missing members in Qute templates should not appear if the Java type is a source file, since it makes more sense to add it as a field or getter to the source file.

If we know if the type is a source file or a binary file before we generate the text edits associated with the code action, we can generate only the code actions that are needed.

See https://github.com/redhat-developer/quarkus-ls/pull/671#issuecomment-1197058836

datho7561:
When generating CodeActions for missing properties that are referenced in a Qute template, don't suggest adding a field or adding a getter when the type that is being referred to is a binary type (i.e. built-in class or class from an external library).

Similarly, don't suggest creating a template extension for a type whose source code is available.

Fixes #676

### DIFF
--- a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/GenerateMissingJavaMemberParams.java
+++ b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/GenerateMissingJavaMemberParams.java
@@ -9,6 +9,8 @@
 *******************************************************************************/
 package com.redhat.qute.commons;
 
+import org.eclipse.xtext.xbase.lib.util.ToStringBuilder;
+
 /**
  * Parameters required for generating a field, getter, or template extension in
  * a Java type.
@@ -96,6 +98,29 @@ public class GenerateMissingJavaMemberParams {
 
 	public void setProjectUri(String projectUri) {
 		this.projectUri = projectUri;
+	}
+
+	@Override
+	public String toString() {
+		ToStringBuilder builder = new ToStringBuilder(this);
+		builder.add("memberType", memberType);
+		builder.add("missingProperty", missingProperty);
+		builder.add("javaType", javaType);
+		builder.add("projectUri", projectUri);
+		return builder.toString();
+	}
+
+	@Override
+	public boolean equals(Object other) {
+		if (other == null || !(other instanceof GenerateMissingJavaMemberParams)) {
+			return false;
+		}
+		GenerateMissingJavaMemberParams that = (GenerateMissingJavaMemberParams) other;
+		return (this.memberType == that.memberType) //
+				&& ((this.missingProperty == null && that.missingProperty == null)
+						|| this.missingProperty.equals(that.missingProperty)) //
+				&& ((this.javaType == null && that.javaType == null) || this.javaType.equals(that.javaType)) //
+				&& ((this.projectUri == null && that.projectUri == null) || this.projectUri.equals(that.projectUri));
 	}
 
 }
--- a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/ResolvedJavaTypeInfo.java
+++ b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/ResolvedJavaTypeInfo.java
@@ -43,6 +43,8 @@ public class ResolvedJavaTypeInfo extends JavaTypeInfo {
 
 	private Boolean isIterable;
 
+	private Boolean source;
+
 	private RegisterForReflectionAnnotation registerForReflectionAnnotation;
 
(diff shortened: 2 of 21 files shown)


datho7561:
@AlexXuChen in this PR I added a new `isSource` property to `UnknownPropertyData`: https://github.com/redhat-developer/quarkus-ls/pull/686/files#diff-57af4a93ff85289d099d6400ad0aa2ac0a53bdaf4e19d523dfbbcd9a6f63fb38

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/ResolvedJavaTypeInfo.java
```diff
@@ -43,6 +43,8 @@ public class ResolvedJavaTypeInfo extends JavaTypeInfo {
 
 	private Boolean isIterable;
 
+	private Boolean isSource;
```
angelozerr, line 46: Rename to source

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/ResolvedJavaTypeInfo.java
```diff
@@ -151,6 +153,26 @@ public boolean isIterable() {
...
+	public boolean isSource() {
+		return isSource != null && isSource.booleanValue();
+	}
+	
+	/**
+	 * Set if this type comes from a source file.
+	 * 
+	 * @param isSource true if the type comes from a source file, false otherwise
+	 */
+	public void setIsSource(Boolean isSource) {
```
angelozerr, line 172: Rename to setSource

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/GenerateMissingJavaMemberParams.java
```diff
@@ -98,4 +98,28 @@ public void setProjectUri(String projectUri) {
 		this.projectUri = projectUri;
 	}
 
+	@Override
+	public String toString() {
+		StringBuilder buf = new StringBuilder("GenereateMissingJavaMemberParams [\n");
```
angelozerr, line 103: Why not using ToStringBuilder?

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteCodeActions.java
```diff
@@ -173,62 +171,63 @@ private void doCodeActionToCreateProperty(Template template, Diagnostic diagnost
...
+					}));
+
+			registrations.add(resolver.generateMissingJavaMember(getterParams) //
+					.thenAccept((workspaceEdit) -> {
+						if (workspaceEdit == null) {
+							return;
+						}
+						createGetter.setEdit(workspaceEdit);
+					}));
+		} else {
```
angelozerr, line 202: Why else? Even if type is source it should have à codeaction which cteate a template extension.

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/GenerateMissingJavaMemberParams.java
datho7561: Thanks! Didn't know about this method.

angelozerr: you did it here https://github.com/redhat-developer/quarkus-ls/pull/686/files#diff-70b7e08819155fab891313cee90b7612e06e38ebfda05864c4ed37df00621807R259 :)

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/ResolvedJavaTypeInfo.java
```diff
@@ -234,6 +256,7 @@ public String toString() {
 		ToStringBuilder b = new ToStringBuilder(this);
 		b.add("name", this.getName());
 		b.add("signature", this.getSignature());
+		b.add("isSource", this.isSource() ? "SOURCE" : "BINARY");
```
angelozerr, line 259: b.add("source", this.isSource() ? "SOURCE" : "BINARY");

angelozerr:
It works like a charm. Thanks @datho7561 !

PR #688, opened 2022-07-29
Create a code action class per error code.

angelozerr:
Create a code action class per error code.


angelozerr:
This PR clean Qute code actions:

 * there is a code action class per error code to avoid having a big class file QuteCodeActions which is hard to read. Splitting code action per error code will simplify the contribution to code action because it will not impact the single QuteCodeAction class.
 * CodeActionRequest wraps all input parameters (template, diagnostic). We ca add another parameters in the future without breaking API. CodeActionRequest provides :
   * getCoveredNode which is the covered node higlighted by the diagnostic range.  
   *  getJavaTypeOfCoveredNode which is the Java type of the covered part.
   * @datho7561 you could add in this class  `canSupportResolve`  method likehttps://github.com/eclipse/lemminx/blob/18e3bdf48a26e0cae6d00fb3f5edb6a4fe1b6708/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/codeaction/ICodeActionRequest.java#L49 when you will support resolve of code action.
 * I simplify data which stores only the Java base type signature (not the source, not the property name) because those informations can be retrieved easily with the convered node (ex : property name) and the Java type of the covered part (ex : source).

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/diagnostics/QuteDiagnosticContants.java
```diff
@@ -11,17 +11,17 @@
 *******************************************************************************/
 package com.redhat.qute.services.diagnostics;
 
+/**
+ * Constants for Qute diagnostics.
```
datho7561, line 15: I think you misspelled your name

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/codeactions/AbstractQuteCodeAction.java
```diff
@@ -0,0 +1,130 @@
...
+
+	/**
+	 * Generate code action(s) sor similar values (Did you mean ...?)
+	 * 
+	 * @param part            the part node.
+	 * @param availableValues the available values.
+	 * @param template        the Qute template.
+	 * @param diagnostic      the diagnostic.
+	 * @param codeActions     list of CodeActions.s
+	 */
```
datho7561, line 64: Small typos:
* sor -> for
* The "s" after the "." for the param codeActions

datho7561 (changes requested):
Looks really good. I like the code structure, and I think this will help when we need to implement more code actions. I caught a few small documentations things, and once they are addressed, I think it should be good to merge.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/codeactions/AbstractQuteCodeAction.java
angelozerr: fixed

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/diagnostics/QuteDiagnosticContants.java
angelozerr: fixed

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/codeactions/AbstractQuteCodeAction.java
```diff
@@ -0,0 +1,130 @@
...
+				codeActions.add(similarCodeAction);
+			}
+		}
+	}
+
+	/**
+	 * CodeAction to change severity setting value to "ignore"
+	 *
+	 * @param template        the Qute template.
+	 * @param diagnostics     a singleton list diagnostic to set to ignore.
```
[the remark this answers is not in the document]
angelozerr: fixed

AlexXuChen (commented):
LGTM, can confirm all CodeActions are still working. One thing I noticed is that there is a flicker for "Quick Fix..." pnly when hovering any`UnknownProperty` diagnostic:
Peek 2022-08-02 09-58
However, all `UnknownProperty` CodeActions are available and working, seems to be a small performance issue when generating the code action.

qute.jdt/com.redhat.qute.jdt.test/projects/maven/qute-quickstart/src/main/resources/templates/hello.qute.html
```diff
@@ -2,18 +2,9 @@
 <html>
 <head>
 <meta charset="UTF-8">
-<title>Qute Hello World</title>
+<title>Qute Hello World</title> 
 </head>
 <body>
     <p>Hello {name ?: "world"}!</p>
 </body>
-</html>
```
datho7561, line 10: Any reason this is being changed?

AlexXuChen: I accidentally pushed a change to this test file in my CodeAction pr, I just asked Angelo if he could revert it.

datho7561 (approved):

PR #732, opened 2022-09-01
Detect cycles in inheritance graph during completion

### LINKED ISSUE #725: Detect cyclical class hierarchy and prevent infinite recursion when calculating all fields and methods of an object
Let's say we have the following classes:
* `ClassA` which extends `ClassC`
* `ClassB` which extends `ClassA`
* `ClassC` which extends `ClassB`

JDT reports an error for this since it can't be compiled. If you make a Qute template that includes one of these classes as a parameter, then open up completion:

```qute
{@org.acme.ClassA objectA}
{objectA.|}
```

Then you won't ever get completion back, since qute-ls uses recursion in order to collect the fields and methods of the subtypes, and will be stuck cycling between the 3 types. I'd be nice to prevent the infinite recursion in this case.

datho7561:
If a cycle is detected, don't return any of the properties or methods of the class. Prevent ResolvedJavaTypeInfo from being fetched from qute.jdt multiple times. This prevents completion from freezing when you try to complete the properties and methods of an object whose class is part of an inheritance cycle.

Closes #725


datho7561:
Nevermind, this PR needs a lot more work

PR #735, opened 2022-09-07
Prevent infinite recursion in validation, completion, code actions

### LINKED ISSUE #725: Detect cyclical class hierarchy and prevent infinite recursion when calculating all fields and methods of an object

datho7561:
Prevent types from being visited twice while collecting type information. This prevents validation, completion, and code actions from freezing when working with objects of classes that have cyclic inheritance.

Closes #725

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/QuteProjectRegistry.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/QuteProjectRegistry.java
@@ -22,6 +22,7 @@ import static com.redhat.qute.services.QuteCompletableFutures.VALUE_RESOLVERS_NU
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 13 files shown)


qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/QuteProjectRegistry.java
```diff
@@ -211,6 +213,17 @@ public void onDidCloseTextDocument(TemplateInfoProvider document) {
...
 
 	public CompletableFuture<ResolvedJavaTypeInfo> resolveJavaType(String javaTypeName, String projectUri) {
+		return resolveJavaType(javaTypeName, projectUri, new HashSet<>());
+	}
+
+	private CompletableFuture<ResolvedJavaTypeInfo> resolveJavaType(String javaTypeName, String projectUri,
+			Set<String> visited) {
+
+		if (visited.contains(javaTypeName)) {
+			return null;
```
angelozerr, line 223: return RESOLVED_JAVA_CLASSINFO_NULL_FUTURE;

angelozerr:
Thanks @datho7561 !

PR #739, opened 2022-09-09
Fix missing definition for #case section Enum with operator name

### LINKED ISSUE #738: Definition for Enum in #case section fails for Enum with operator name
When trying to get the definition of a Enum in a #case section with the same name as a case operator, it will not work in the case that it is the only parameter. 

Ex: {#case i|n} will not work, while {#case in i|n} will.

Peek 2022-09-09 11-01

JessicaJHee:
Peek 2022-09-09 11-08

Fixes #738 

Signed-off-by: Jessica He <jhe@redhat.com>

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDefinition.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDefinition.java
@@ -350,7 +350,7 @@ class QuteDefinition {
 		if (!isWhenSection(whenSection)) {
 			return NO_DEFINITION;
 		}
-		if (((CaseSection) caseSection).isCaseOperator(parameter)) {
+		if (((CaseSection) caseSection).isCaseOperator(parameter) && caseSection.getParameters().size() > 1) {
 			return NO_DEFINITION;
 		}
 
--- a/qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/definition/QuteDefinitionInWhenSectionTest.java
+++ b/qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/definition/QuteDefinitionInWhenSectionTest.java
@@ -72,7 +72,7 @@ public class QuteDefinitionInWhenSectionTest {
 				"  {#is O|N }";
 		testDefinitionFor(template);
 	}
-	
+
 	@Test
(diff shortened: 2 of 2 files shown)


qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDefinition.java
```diff
@@ -295,7 +295,7 @@ private CompletableFuture<List<? extends LocationLink>> findDefinitionFromParame
 		if (!isWhenSection(whenSection)) {
 			return NO_DEFINITION;
 		}
-		if (((CaseSection) caseSection).isCaseOperator(parameter)) {
+		if (((CaseSection) caseSection).isCaseOperator(parameter) && caseSection.getParameters().size() > 1) {
```
angelozerr, line 298: what about?

`{#is in i|n}`

where `in` is an enum value.

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/services/definition/QuteDefinitionInWhenSectionTest.java
```diff
@@ -131,6 +131,16 @@ public void definitionInEnumOnOperatorAsEnum() throws Exception {
 				ll("org/acme/MachineStatus.java", r(2, 10, 2, 12), MockQuteProjectRegistry.JAVA_FIELD_RANGE));
 	}
 
+	@Test
+	public void definitionInEnumOnOperatorAsEnumOnly() throws Exception {
+		String template = "{@org.acme.Machine Machine}\r\n" + //
+				"{#when Machine.status}\r\n" + //
+				"  {#is i|n}\r\n" + //
```
angelozerr, line 138: It should be nice to have 2 new tests:

 * `{#is i|n in}` : it should return nothing.
 * `{#is in i|n}` : it should return a definition.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDefinition.java
JessicaJHee: In that case isCaseOperator() will return false, and it should find the definition

angelozerr:
Thanks @JessicaJHee !

PR #743, opened 2022-09-14
Generic support for Java data model

### LINKED ISSUE #503: Support for Map java data model
```
{@java.util.Map<java.lang.String,org.acme.Item> map}

[map.| <-- we should have completion, validation, hover, etc)
```

angelozerr:
Generic support for Java data model

Fixes #503

### DIFF
--- a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaElementInfo.java
+++ b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaElementInfo.java
@@ -11,8 +11,6 @@
 *******************************************************************************/
 package com.redhat.qute.commons;
 
-import java.util.StringJoiner;
-
 /**
  * Base class for Java element (Java types, methods, fields, parameters)
  *
@@ -112,33 +110,29 @@ public abstract class JavaElementInfo {
 	 * @return the simple type of the given type.
 	 */
 	public static String getSimpleType(String type) {
-		if (type == null) {
-			return null;
+		if (type == null || type.indexOf('.') == -1) {
+			// ex: List
+			return type;
 		}
-		int startBracketIndex = type.indexOf('<');
-		if (startBracketIndex != -1) {
-			int endBracketIndex = type.indexOf('>', startBracketIndex);
-			// Main type
-			StringBuilder simpleType = new StringBuilder(
-					getSimpleTypeWithoutGeneric(type.substring(0, startBracketIndex)));
-			// Generic type
-			simpleType.append('<');
-			String[] generics = type.substring(startBracketIndex + 1, endBracketIndex).split(",");
-			StringJoiner commaJoiner = new StringJoiner(",");
-			for (String generic : generics) {
-				commaJoiner.add(getSimpleTypeWithoutGeneric(generic));
-			}
-			simpleType.append(commaJoiner.toString());
-			simpleType.append('>');
-
-			return simpleType.toString();
+		int lastIndex = 0;
+		StringBuilder result = new StringBuilder();
+		for (int i = 0; i < type.length(); i++) {
+			char c = type.charAt(i);
+			switch(c) {
+			case '.':
+				lastIndex = i;
+				break;
+			case '<':
+			case ',':
+				result.append(type.substring(lastIndex + 1, i + 1));
+				lastIndex = i;
+				break;
+			}			
 		}
-		return getSimpleTypeWithoutGeneric(type);
-	}
-
-	private static String getSimpleTypeWithoutGeneric(String type) {
-		int index = type.lastIndexOf('.');
-		return index != -1 ? type.substring(index + 1, type.length()) : type;
+		if (lastIndex != type.length()) {
+			result.append(type.substring(lastIndex + 1, type.length()));
+		}
+		return result.toString();
 	}
 
 }
--- a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaFieldInfo.java
+++ b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaFieldInfo.java
@@ -11,6 +11,8 @@
 *******************************************************************************/
 package com.redhat.qute.commons;
 
+import java.util.Map;
+
 import org.eclipse.xtext.xbase.lib.util.ToStringBuilder;
 
 /**
@@ -124,6 +126,42 @@ public class JavaFieldInfo extends JavaMemberInfo {
 		return getType();
 	}
 
+	/**
+	 * Returns a new field with the generic type information applied to the
+	 * signature.
+	 * 
+	 * <code>
+	 *    field : T
+	 * </code>
+	 * 
+	 * becomes:
+	 * 
+	 * <code>
+	 *    field : java.lang.String
+	 * </code>
+	 * 
+	 * where 'T' is filled in the generic Map with the 'java.lang.String' value.
+	 * 
+	 * @param field      the Java field.
+	 * @param genericMap the generic Map
+	 * 
+	 * @return a new field with the generic type information applied to the
+	 *         signature.
+	 */
+	public static JavaFieldInfo applyGenericTypeInvocation(JavaFieldInfo field, Map<String, String> genericMap) {
+		// ex :
+		// - field : java.lang.String
+		// - field : T
+		// - field : java.util.List<T>
+		JavaFieldInfo newField = new JavaFieldInfo();
+		StringBuilder newSignature = new StringBuilder();
+		newSignature.append(field.getName());
+		newSignature.append(" : ");
+		JavaTypeInfo.applyGenericTypeInvocation(field.getType(), genericMap, newSignature);
+		newField.setSignature(newSignature.toString());
+		return newField;
+	}
+
 	@Override
 	public String toString() {
 		ToStringBuilder b = new ToStringBuilder(this);
@@ -132,4 +170,5 @@ public class JavaFieldInfo extends JavaMemberInfo {
 		b.add("signature", this.getSignature());
 		return b.toString();
 	}
+
 }
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaElementInfo.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaElementInfo.java
@@ -11,8 +11,6 @@
 *******************************************************************************/
 package com.redhat.qute.commons;
 
-import java.util.StringJoiner;
-
 /**
  * Base class for Java element (Java types, methods, fields, parameters)
  *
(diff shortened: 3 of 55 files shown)


angelozerr:
I need to write tests again, but it should start working. Here a demo

MapDemo

It is very important to spend some time to test this PR (even if we have a lot of tests) because the resolve of JavaType is now generic and it should work with Map and List (like before). For instance it is important to play with List with #for.

datho7561:
If I have:
```qute
{@java.util.ArrayList<java.lang.String> myStringArrayList}

{#for item in myStringArrayList}
    {item.name}
{/for}
```

... it says `myStringArrayList` is not iterable:

ArrayListStringIsNotIterable

rgrunber:
I tried the same for `java.util.HashSet`. However when I represented the injected element as `java.util.Set`, the error went away! So maybe @datho7561 's issue is the same, and `java.util.List` might fix it. Maybe the base type of the element is the only thing that is checked ?

angelozerr:
Good catch! I never declared parameter declaration with iterable implementation. Just with interface like List.

I think this bug exists too in Master.

Let me time to investigate the problem.

angelozerr:
@datho7561 @rgrunber I need to clean my code, but HashMap should work now.

angelozerr:
After playing with my PR, I noticed a big bug with cache when you write

```
{@java.util.HashMap<Integer,org.acme.Item> map}
{@java.util.HashMap<Integer,String> map2}
```

When you loop for map, the value is String instead of Item. Need to fix that and write more tests...

rgrunber:
This doesn't seem to be working for me. Relevant parts :

```
ArrayList<Person> people = new ArrayList<>();
...
...
return hello.data("people", people).render();
```

where `hello` is :
```
@Location("RootResource/hello.html")
@Inject
Template hello;
```

In the template I define :

```
{@java.util.ArrayList<org.example.Person> people}
...
...
{#for person in people}
...
...
```

`people` still has a diagnostic of `Iteration error: {people} resolved to [java.util.ArrayList<org.example.Person>] which is not iterable. qute(IterationError)`

angelozerr:
@rgrunber I pushed my  work because the jdt change was big and it should work.I need to write tests and clean the code.

Non I need to finish the qute ls side. Tests should pass but there is à problem with cache that I need to fix and write too à test with this cache which breaks totally the qute java data model features like you noticed.

Please wait I ping you when you could tests again.

Il sorry that you lost your time to test this PR which is not ready.

angelozerr:
@rgrunber I push a new commit and it should work better (I hope), but I play quickly so perhaps there is some bugs again.

If you have time you can play with the PR but I need to clean the code and write some tests.

rgrunber:
This is working well for me now :+1: .

datho7561:
One thing I caught is handling cyclic dependencies:

```java
public ClassA<T> extends ClassC<T> {
}
// ...
public ClassB<T> extends ClassA<T> {
}
// ...
public ClassC<T> extends ClassB<T> {
}
```

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaElementInfo.java
```diff
@@ -112,33 +110,27 @@ public String getJavaElementSimpleType() {
...
-			simpleType.append('<');
-			String[] generics = type.substring(startBracketIndex + 1, endBracketIndex).split(",");
-			StringJoiner commaJoiner = new StringJoiner(",");
-			for (String generic : generics) {
-				commaJoiner.add(getSimpleTypeWithoutGeneric(generic));
+		StringBuilder result = new StringBuilder();
+		boolean parsingClassName = false;
+		for (int i = 0; i < type.length(); i++) {
+			char c = type.charAt(i);
+			if (!parsingClassName && Character.isUpperCase(c)) {
```
datho7561, line 121: Similar comment about the class name

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
```diff
@@ -532,10 +524,15 @@ private ResolvedJavaTypeInfo validateExpressionParts(Parts parts, Section ownerS
 					// {items.size()}
 					// Property, method to validate must be done for iterable type (ex :
 					// java.util.List
-					String iterableType = baseType.getIterableType();
-					baseType = resolveJavaType(iterableType, projectUri, resolvingJavaTypeContext);
-					if (baseType == null || isResolvingJavaType(baseType)) {
-						// The java type doesn't exists or it is resolving, stop the validation
+					// String iterableType = baseType.getIterableType();
```
datho7561, line 527: If this code in the comments is unused, remove it

qute.ls/com.redhat.qute.ls/src/test/java/com/redhat/qute/commons/ResolvedJavaTypeInfoTest.java
```diff
@@ -0,0 +1,181 @@
+/*******************************************************************************
+* Copyright (c) 2021 Red Hat Inc. and others.
```
datho7561, line 2: 2022

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/QuteProjectRegistry.java
```diff
@@ -229,48 +239,109 @@ private CompletableFuture<ResolvedJavaTypeInfo> resolveJavaType(String javaTypeN
 			// It's a primitive type like boolean, double, float, etc
 			return primitiveType;
 		}
-		if (StringUtils.isEmpty(javaTypeName) || StringUtils.isEmpty(projectUri)) {
-			return RESOLVED_JAVA_CLASSINFO_NULL_FUTURE;
+
+		String projectUri = project.getUri();
```
datho7561, line 243: I think it might be a good idea to split this method up into multiple methods to make it more legible

datho7561 (changes requested):

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaElementInfo.java
```diff
@@ -112,33 +110,27 @@ public String getJavaElementSimpleType() {
...
-			simpleType.append('<');
-			String[] generics = type.substring(startBracketIndex + 1, endBracketIndex).split(",");
-			StringJoiner commaJoiner = new StringJoiner(",");
-			for (String generic : generics) {
-				commaJoiner.add(getSimpleTypeWithoutGeneric(generic));
+		StringBuilder result = new StringBuilder();
+		boolean parsingClassName = false;
+		for (int i = 0; i < type.length(); i++) {
+			char c = type.charAt(i);
+			if (!parsingClassName && Character.isUpperCase(c)) {
```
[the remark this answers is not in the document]
angelozerr: I know it is a limitation, but I have not found a better solution for the case of  inner class/interface like Map.Entry

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaFieldInfo.java
```diff
@@ -124,6 +126,41 @@ public String getJavaElementType() {
 		return getType();
 	}
 
+	/**
+	 * Returns a new field with the apply of generic type invocation.
```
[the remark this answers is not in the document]
angelozerr: fixed

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaElementInfo.java
angelozerr: See https://github.com/redhat-developer/quarkus-ls/pull/743#discussion_r980219970

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/QuteProjectRegistry.java
angelozerr: Is it better?

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaElementInfo.java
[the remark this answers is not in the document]
datho7561: Is there a way to get the fully qualified name of innerclasses using the same for class `.class.getName()`? eg. so it would be `java.util.Map$Entry`?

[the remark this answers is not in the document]
mkouba: It should be `java.util.Map$Entry`.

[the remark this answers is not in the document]
angelozerr: Thanks @mkouba for your feedback!

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/JavaElementInfo.java
angelozerr: fixed, now I'm using $

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/JavaElementInfo.java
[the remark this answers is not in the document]
angelozerr: fixed, now I'm using $

angelozerr:
This PR should be good, however I noticed several problem with parameter declaration that we could fix in another PRs:

 * `{@java.util.Map$|}` completion after $ doesn't work.
 * with `{@java.util.List<java.util.List<XXXX>> list}` the `XXXX` is not highlighted as an error although  `{@java.util.List<XXXX> list}` is correctly highlighted. In other words I don't take care of genereic of generic for validation.
 * same problem with definition `{@java.util.List<java.util.List<java.lan|g.String>> list}`  cannot find definition for java.lang.String
 * `{@java.util.List<java.lang.String list}` it misses the end of `>`, the validation ignore this error.
 * `{@java.util.List<java.lang.String,java.lang.Integer> list}` in this case `java.lang.Integer` must be highlighted as an error because List expected just one generic type parameter and not 2.

datho7561:
I had some server crashes that I want to investigate, and it looks like some tests are failing.

angelozerr:
> I had some server crashes that I want to investigate, and it looks like some tests are failing.

Indeed I notice that there were a problem wiith signature of entrySet of Map. I push a new commit and hope tests will work.

datho7561 (approved):
Works pretty well for me, the only things that I've found are things related to the limitations of the implementation

angelozerr:
@datho7561 now you can use Map in the CheckedTemplate, it should work now.

datho7561 (approved):

Issue #742, opened 2022-09-13
Qute parser does not parse operator parameters with '=' correctly


JessicaJHee:
Operators used in #case or #is sections that include the `=` character does not get parsed correctly.

Ex: 
`{#case <= 10}`

In this case, for operators `<=` or `>=`, the parser only scans `<` or `>` as the parameter since `=` is assumed to be the end of a token in the parser and is cut off from the parameter.

PR #744, opened 2022-09-14
Support hover for operators in #case section

### LINKED ISSUE #716: Hover operator for #case #is
Given this Qute template:

```
{#switch name}
	{#case in "abcd" "efgh"}
{/switch}
```

The hover in `in` `{#case i|n` should show some documentation.

JessicaJHee:
Fixes #716 

Signed-off-by: Jessica He <jhe@redhat.com>

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/template/DocumentableItem.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/template/DocumentableItem.java
@@ -1,0 +1,42 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.qute.parser.template;
+
+import java.util.List;
+
+/**
+ * DocumentableItem.
+ *
+ */
+public interface DocumentableItem {
+
+	/**
+	 * Returns the description of the item and null otherwise.
+	 * 
+	 * @return the description of the item and null otherwise.
+	 */
+	String getDescription();
+
+	/**
+	 * Returns a sample with the item and null otherwise.
+	 * 
+	 * @return a sample with the item and null otherwise.
+	 */
+	List<String> getSample();
+
+	/**
+	 * Returns the documentation Url of the item and null otherwise.
+	 * 
+	 * @return the documentation Url of the item and null otherwise.
+	 */
+	String getUrl();
+}
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/template/sections/CaseSection.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/template/sections/CaseSection.java
@@ -11,6 +11,7 @@
 *******************************************************************************/
 package com.redhat.qute.parser.template.sections;
 
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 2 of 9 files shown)


JessicaJHee:
There is an issue with the parser that prevents hover for operators `<=` and` >=` to work correctly here: https://github.com/redhat-developer/quarkus-ls/issues/742. Will continue this PR after the fix.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteHover.java
```diff
@@ -279,8 +281,22 @@ private CompletableFuture<Hover> doHoverForObjectPart(Part part, String projectU
 			}
 		}
 
-		// Check if part is a literal (ex: true, null, 123, 'abc', etc)
 		Expression expression = part.getParent().getParent();
+
+		if (CaseSection.isCaseSection(expression.getOwnerSection())) {
+			CaseSection caseSection = (CaseSection) expression.getOwnerSection();
```
angelozerr, line 287: Add a sample of this hover usecase

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/utils/DocumentationUtils.java
```diff
@@ -199,6 +200,26 @@ public static MarkupContent getDocumentation(UserTag userTag, boolean markdown)
 
 		return createMarkupContent(documentation, markdown);
 	}
+	
+	public static MarkupContent getDocumentation(CaseOperator operator, boolean markdown) {
```
angelozerr, line 204: Add Javadoc

angelozerr: Call this utility method too for completion at https://github.com/redhat-developer/quarkus-ls/blob/bdafad0b1fdacbe4d450a0cb0b24462ee44a27b2/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForExpression.java#L522 to have the same renderer in completion and hover .

angelozerr: It should be really nice if we could improve the documentation like we have for `or` method value resolver :

image

To do that, you need : 

 * a description
 * a sample (list of String)
 * an URL

In other words Operation should store those information. I suggest that you create a `DocumentableItem` interface like this:

```java
package com.redhat.qute.parser.template;

import java.util.List;

public interface DocumentableItem {

	String getDescription();
	
	List<String> getSample();
	
	String getUrl();
}

```

This interface should be implemented with:

 * Operator
 * MethodValueResolver

After that you can extract this code 
https://github.com/redhat-developer/quarkus-ls/blob/bdafad0b1fdacbe4d450a0cb0b24462ee44a27b2/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/utils/DocumentationUtils.java#L77 in a new method which have DocumentableItem as parameter

JessicaJHee:
Peek 2022-09-15 12-36

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/template/DocumentableItem.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.List;
+
+/**
+ * DocumentableItem.
+ *
+ */
+public interface DocumentableItem {
+
+	/**
+	 * Returns the description of the value resolver and null otherwise.
```
angelozerr, line 23: It is not only value resolver now. You can change with item.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForExpression.java
```diff
@@ -519,7 +520,7 @@ private void fillCaseOperators(CaseSection caseSection, boolean onlyMulti, Range
 				item.setLabel(operator.getName() + " : Operator");
 				item.setFilterText(operator.getName());
 				item.setKind(CompletionItemKind.Operator);
-				item.setDocumentation(operator.getDocumentation());
+				item.setDocumentation(DocumentationUtils.getDocumentation(operator, true));
```
angelozerr, line 523: Instead of hard coding with true, you should use the CompletionRequest#canSupportMarkupKind like you do with hover. Compute this flag before the loop.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/template/sections/CaseSection.java
```diff
@@ -57,22 +58,26 @@ public static enum CompletionCaseResult {
...
+	private static void registerOperator(String name, String documentation, List<String> sample, boolean isMulti,
+			String... aliases) {
 		CaseOperator operator = new CaseOperator(name, documentation, null, isMulti);
 		caseOperators.put(operator.getName(), operator);
+		operator.setUrl("https://quarkus.io/guides/qute-reference#when_section");
+		operator.setSample(sample);
 		if (aliases != null) {
 			for (String alias : aliases) {
 				CaseOperator aliasOperator = operator = new CaseOperator(alias, documentation, name, isMulti);
+				aliasOperator.setUrl("https://quarkus.io/guides/qute-reference#when_section");
```
angelozerr, line 79: I think the url should be https://quarkus.io/guides/qute-reference#when_operators

angelozerr:
Great job @JessicaJHee !

PR #745, opened 2022-09-14
Qute parser does not parse operator parameters with '=' correctly

### LINKED ISSUE #742: Qute parser does not parse operator parameters with '=' correctly

angelozerr:
Qute parser does not parse operator parameters with '=' correctly

Fixes #742


JessicaJHee (approved):
Fixes the issue, operators with '=' in case sections are now parsed in properly without cutting off the '=' character.

PR #746, opened 2022-09-15
Update Target Platform to 1.16.0-SNAPSHOT version of JDT-LS target.

rgrunber:
- Use a proper release of JDT-LS 1.15.0 as it should be compatible
- Fix buildAll.sh on Linux so it runs all statements without halting


datho7561 (approved):
Looks good to me! Can confirm the server starts up after these changes

PR #747, opened 2022-09-19
Documentation when hovering object members in Qute

### LINKED ISSUE #452: Display Javadoc on hover for property/method
Display Javadoc on hover for property/method. This get of javadoc should be done only when the hover require it (aka lazy loading) and not when JavaMethodIndo, JavaFiledInfo are getted to keep proper performance and avoid having memory problem.

datho7561:
Closes #452

### DIFF
--- a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/DocumentFormat.java
+++ b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/DocumentFormat.java
@@ -1,0 +1,42 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
+*
+* This program and the accompanying materials are made available under the
+* terms of the Eclipse Public License v. 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+* which is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.qute.commons;
+
+/**
+ * Document format.
+ *
+ * @author Angelo ZERR
+ */
+public enum DocumentFormat {
+
+	PlainText(1), Markdown(2);
+
+	private final int value;
+
+	DocumentFormat(int value) {
+		this.value = value;
+	}
+
+	public int getValue() {
+		return value;
+	}
+
+	public static DocumentFormat forValue(int value) {
+		DocumentFormat[] allValues = DocumentFormat.values();
+		if (value < 1 || value > allValues.length)
+			throw new IllegalArgumentException("Illegal enum value: " + value);
+		return allValues[value - 1];
+	}
+
+}
--- a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/QuteJavadocParams.java
+++ b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/QuteJavadocParams.java
@@ -1,0 +1,92 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package com.redhat.qute.commons;
+
+/**
+ * Represents the parameters needed to fetch the javadocs from the Java language
+ * server component.
+ * 
+ * @author datho7561
+ */
+public class QuteJavadocParams {
+
+	private String sourceType;
+	private String projectUri;
+	private String memberName;
+	private String signature;
+	private DocumentFormat documentFormat;
+
+	public QuteJavadocParams() {
+		// needed for gson
+	}
+
+	public QuteJavadocParams(String sourceType, String projectUri, String memberName, String signature,
+			DocumentFormat documentFormat) {
+		this.sourceType = sourceType;
+		this.projectUri = projectUri;
+		this.memberName = memberName;
+		this.signature = signature;
+		this.documentFormat = documentFormat;
+	}
+
+	/**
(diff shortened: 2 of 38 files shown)


qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/QuteJavadocParams.java
```diff
@@ -0,0 +1,92 @@
...
+*******************************************************************************/
+package com.redhat.qute.commons;
+
+/**
+ * Represents the parameters needed to fetch the javadocs from the Java language
+ * server component.
+ * 
+ * @author datho7561
+ */
+public class QuteJavadocParams {
```
angelozerr, line 20: It should be nice to use the same name than QuteDefinitionParams. For instance using sourceType instead of className.

QuteJavadocParams could extends QuteDefinitionParams?

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/QuteSupportForTemplate.java
```diff
@@ -566,5 +562,122 @@ public WorkspaceEdit generateMissingJavaMember(GenerateMissingJavaMemberParams p
...
+		} else {
+			// ex : String implements CharSequence, ....
+			ITypeHierarchy typeHierarchy = type.newSupertypeHierarchy(monitor);
+			IType[] allSuperTypes = typeHierarchy.getSupertypes(type);
+			extendedTypes = Stream.of(allSuperTypes) //
+					.map(superType -> superType.getFullyQualifiedName()) //
+					.collect(Collectors.toList());
+		}
+
+		if (extendedTypes != null) {
```
angelozerr, line 660: Instead of computing a list of String extendedTypes  computed from IType[], it should be perhaps better to work only with IType to avoid using getTypeFromParams although the IType is already known.

The signature could look like this:

`public String getJavadoc(Type type, QuteJavadocParams params, IJDTUtils utils, IProgressMonitor monitor, Set<String> visited) throws JavaModelException {`

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/QuteJavadocParams.java
datho7561: I don't think there is enough overlap in the types for one to extend the other, but I'll rename the `className` to `sourceType` to be consistent

datho7561:
I still need to clean up and document the code, but the tests are written now

datho7561:
I've fixed the merge conflicts, and all the tests should pass, however I need to fix hover documentation for methods (and maybe fields?) for types with generics.

datho7561:
Hover documentation should now work for methods and values of objects of classes that have type parameters, but I need to write tests for this use case still.

angelozerr:
I play with this PR and it looks really promising. Great stuff @datho7561 !

I wonder why we have not the same renderer than vscode-java. For instance in Qute template I have:

image

image

In vscode-java, `clear`  is higlighted although in Qute template it is not highlighted, do you know why?

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/commons/DocumentFormat.java
```diff
@@ -0,0 +1,42 @@
+/*******************************************************************************
+* Copyright (c) 2020 Red Hat Inc. and others.
```
angelozerr, line 2: 2022

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/QuteSupportForTemplate.java
```diff
@@ -477,4 +482,126 @@ public WorkspaceEdit generateMissingJavaMember(GenerateMissingJavaMemberParams p
...
+					return javadoc;
+				}
+			}
+		}
+
+		return null;
+
+	}
+
+	private IType getTypeFromParams(String projectUri, String typeName, IProgressMonitor monitor) throws JavaModelException {
```
angelozerr, line 598: I will change the signature with 

`getTypeFromParams(String typeName, String projectUri, IProgressMonitor monitor) `

to be consistent with other method.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/commons/DocumentFormat.java
```diff
@@ -0,0 +1,42 @@
+/*******************************************************************************
+* Copyright (c) 2020 Red Hat Inc. and others.
```
angelozerr, line 2: 2022

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteHover.java
```diff
@@ -354,28 +426,31 @@ private CompletableFuture<Hover> doHoverForPropertyPart(Part part, String projec
...
 		JavaMemberInfo member = javaCache.findMember(part, resolvedType, projectUri);
 		if (member == null) {
-			return null;
+			return NO_HOVER;
 		}
 		boolean hasMarkdown = hoverRequest.canSupportMarkupKind(MarkupKind.MARKDOWN);
-		MarkupContent content = DocumentationUtils.getDocumentation(member, iterableOfResolvedType, hasMarkdown);
-		Range range = QutePositionUtility.createRange(part);
-		return new Hover(content, range);
+		CompletableFuture<Void> fetchDocsFuture = CompletableFuture.completedFuture(null);
```
angelozerr, line 445: If member has no Javadoc, it means that JTD LS javadoc command will be called again for a second hover? If yes we should perhaps add a flag in member which says that Javadoc has been resolved.

datho7561: Okay, it should only fetch the documentation on the first call now.

datho7561:
> I wonder why we have not the same renderer than vscode-java.

I copy/pasted the implementation for collecting the Javadoc from lsp4mp. I'll take a look at eclipse.jdt.ls, figure out what it's doing, and see if I can replicate the behaviour in qute-ls

datho7561:
> I wonder why we have not the same renderer than vscode-java.

Alright. It should be fixed! I'll open a bug on lsp4mp to use JavadocContentAccess2 to get the better Markdown formatting for Javadoc

angelozerr:
It works like a charm, great job @datho7561 !

PR #757, opened 2022-10-27
Fix tests by taking into account new property from lsp4mp

### LINKED ISSUE #756: Build fails on `master`
As of 27 October 2022, the tests on the latest commit fail. This is probably due to an upstream change in lsp4mp, maybe related to new properties being introduced

datho7561:
Closes #756


angelozerr:
Thanks @datho7561 !

PR #774, opened 2023-01-20
Prevent infinite loop in code action

### LINKED ISSUE #712: Qute generate missing member cannot generate new template extension when `TemplateExtensions` and `TemplateExtensions0` exist
Say you have a Qute template that references a property that doesn't exist:

```html
{myObj.asdf}
```

... and you have two classes in the default package named `TemplateExtensions` and `TemplateExtensions0`.

If you apply the code action to generate a new template extensions class, then nothing happens.

datho7561:
Fixes #712

### DIFF
--- a/qute.jdt/com.redhat.qute.jdt.test/src/main/java/com/redhat/qute/jdt/template/TemplateGenerateMissingJavaMemberTest.java
+++ b/qute.jdt/com.redhat.qute.jdt.test/src/main/java/com/redhat/qute/jdt/template/TemplateGenerateMissingJavaMemberTest.java
@@ -19,7 +19,10 @@ import java.util.regex.Pattern;
 
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.jdt.core.ICompilationUnit;
 import org.eclipse.jdt.core.IJavaProject;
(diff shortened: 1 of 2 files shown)


quarkus.ls.ext/com.redhat.quarkus.ls/src/test/java/com/redhat/quarkus/ls/JavaTextDocumentSnippetRegistryTest.java
```diff
@@ -23,6 +23,7 @@
 import org.eclipse.lsp4mp.ls.commons.snippets.ISnippetContext;
 import org.eclipse.lsp4mp.ls.commons.snippets.Snippet;
 import org.eclipse.lsp4mp.ls.commons.snippets.SnippetRegistry;
+import org.eclipse.lsp4mp.snippets.JavaSnippetCompletionContext;
```
angelozerr, line 26: Is it this change is part of this PR?

datho7561: No, but the tests were failing locally if I didn't have this. However, it fails on GitHub Actions if I make this change, so I'll remove it

PR #790, opened 2023-02-10
Definition for user tag parameter

### LINKED ISSUE #789: Definition for user tag parameter
Given this user tag `foo.html`

`{bar}`

It should be nice to go to the declaration from the template which declares the user tag:

`{#foo b|ar="" }`

The definition on bar should jump to the bar object part of foo.html.

Here a demo:

QuteUserTagDefinition

angelozerr:
Definition for user tag parameter

Fixes #789

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/tags/UserTag.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/tags/UserTag.java
@@ -165,6 +165,24 @@ public abstract class UserTag extends Snippet {
 	}
 
 	/**
+	 * Returns the parameter which have the given name <code>parameterName</code>
+	 * and null otherwise.
+	 * 
+	 * @param parameterName the parameter name.
+	 * 
+	 * @return the parameter which have the given name <code>parameterName</code>
+	 *         and null otherwise.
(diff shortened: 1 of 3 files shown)


angelozerr:
Here a demo:

QuteUserTagDefinition

datho7561:
What should happen if the user tag references the parameter many times?

angelozerr:
> What should happen if the user tag references the parameter many times?

It go to the first definition.

datho7561:
Hmm. I think it would be nice if we could show all the places it's used in the custom tag, but maybe that's more suited for "references".

Also, if you have a type declaration of the parameter at the top of the custom tag template, it would be nice if "go to definition" went to that.

angelozerr:
> Hmm. I think it would be nice if we could show all the places it's used in the custom tag

Ok I close this PR, and I created a new one at https://github.com/redhat-developer/quarkus-ls/pull/792 but it is in draft because it is linked to https://github.com/redhat-developer/quarkus-ls/pull/791 (please review it before).

angelozerr:
> Hmm. I think it would be nice if we could show all the places it's used in the custom tag, but maybe that's more suited for "references".

Please see the new PR https://github.com/redhat-developer/quarkus-ls/pull/794

PR #794, opened 2023-02-13
Definition for user tag parameter

### LINKED ISSUE #789: Definition for user tag parameter

angelozerr:
Definition for user tag parameter

Fixes #789


datho7561:
IMO it makes more sense for this to be `references` instead of `definition`. What do you think?

datho7561:
Okay

datho7561 (approved):
Looks good. Thanks, Angelo!

datho7561:
(the test are failing for an unrelated issue that I'll look into addressing, so I'll merge this for now).

PR #803, opened 2023-02-23
Support linked editing range for start/end tag section

### LINKED ISSUE #802: Support linked editing range for start/end tag section
To support correctly [surround with tag section](https://github.com/redhat-developer/vscode-quarkus/issues/581), we need to support:

 * linked editing range for start/end tag section
 * when linked editing range is not activated we need to update end tag section when completion is applied for section

@JessicaJHee I'm working on this issue.

angelozerr:
Support linked editing range for start/end tag section

Fixes #802


angelozerr:
After playing with linked editing range for start/end section, I think it is a bad idea because when you have this template:

```
{#for item in items}
{\for}
```

You dont want to replace for with if because there are not the same parameters between for and if.

PR #806, opened 2023-02-27
Update orphan end tag section when completion is applied

### LINKED ISSUE #805: Completion should generate only start section when empty end section is found
To support correctly [surround with tag section](https://github.com/redhat-developer/vscode-quarkus/issues/581), we need generate only start section when empty end section is found.

Given this template

```html
{#}
```

completion shows for, etc, after for item apply, it generates:

```html
{#for item in items}
  {item.name}
{/for}
```

Given this template which contains an empty orphan end section:

```
{#}
{\}
```

completion shows for, etc, after for item apply, it SHOULD generate:

```
{#for item in items}
{\}
```

@JessicaJHee I'm working on this issue.

angelozerr:
Update orphan end tag section when completion is applied

Fixes #805

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/AbstractTextDocumentService.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/AbstractTextDocumentService.java
@@ -44,12 +44,14 @@ import org.eclipse.lsp4j.ReferenceParams;
 import org.eclipse.lsp4j.RenameParams;
 import org.eclipse.lsp4j.SymbolInformation;
 import org.eclipse.lsp4j.TextDocumentClientCapabilities;
+import org.eclipse.lsp4j.TextDocumentIdentifier;
 import org.eclipse.lsp4j.TextEdit;
 import org.eclipse.lsp4j.WorkspaceEdit;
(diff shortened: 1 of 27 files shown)


angelozerr:
It should start working:

UpdateEndTag

@JessicaJHee could you play with this PR please to see ifit could work with surround.

I need to clean my code and write some again tests.

JessicaJHee:
This works well for the parent section! But it seems to have some issues when I try with the child section. For example, completion with the when and if sections has unexpected behavior when I surround the child expression. Depending on how many character I enter before hitting tab to complete, the result on the end tag is different. 

Peek 2023-02-28 11-19

Peek 2023-02-28 11-51

angelozerr:
Many thanks @JessicaJHee for your great feedback. It is a really bad news and it seems we have the same problem with XML.

I think the compute of additionnal text edit is computed after the insert of character and provide a wrong result. I have no idea for the moment how to fix that -(

angelozerr:
Accordding to https://github.com/microsoft/vscode-languageserver-node/issues/1198#issuecomment-1449773256 it seems that it is not possible to update correctly end tag section.

I close this PR for a new PR https://github.com/redhat-developer/quarkus-ls/pull/807 which doesn't update the end tag section

PR #808, opened 2023-03-06
Add .gitleaks.toml to ignore false positive leaks

datho7561:


rgrunber (approved):
LGTM

PR #823, opened 2023-03-21
Prevent resolving code actions that run commands

datho7561:
Some clients, like VS Code, will attempt to resolve a code action if `edit` is missing, even if a command is provided. In order to prevent this from happening, pass an empty `WorkspaceEdit` object as the `edit` for code actions that run commands.

Fixes redhat-developer/vscode-quarkus#598


angelozerr:
Is it a proper fix? I suggest that you explain the usecase (code action without edit but just with command) in https://github.com/microsoft/vscode-languageserver-node/issues to know if we need to set an empty workspace edit.

My fear is that some LSP client could throw an error if WorkspaceEdit has empty changes.

datho7561:
I opened https://github.com/microsoft/vscode-languageserver-node/issues/1216. See the other issue I linked there; it looks like this behaviour is intentional, but I would like a more concrete confirmation.

datho7561:
tests are flaky ;_;

JessicaJHee (commented):
Solves the problem with the code action failing, but the `UriNamespaceResolverSupportTest` test fails for me

edit: nvm tests are failing on master as well :(

angelozerr:
Many  thanks @datho7561 !

PR #834, opened 2023-03-28
Locate spot to generate `TemplateExtensions.java` better

### LINKED ISSUE #831: `TemplateExtension.java` class generated in `src/main/resources`
1. Create a template in Quarkus project with qute:

```qute-html
{@java.lang.String identifier}

{identifier.sandw|ich}
```

2. Apply the quick fix to resolve the error at the `|`: "Create template extension sandwich() in new template extensions class"

Expected: a new file `TemplateExtensions.java` is generated in `src/main/java` (the default package)

Actual: the file is generated in `src/main/resources`

datho7561:
When locating the folder to generate `TemplateExtension.java` into
during the "generate missing member" Qute quick fix:
- Filter out classpath entries that are in the output (`target`) folder
- Filter out classpath entries that represent test code
- Filter out classpath entries that represent resource folders (folders
  with no Java source files)

Fixes #831

### DIFF
--- a/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/TemplateGenerateMissingJavaMember.java
+++ b/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/TemplateGenerateMissingJavaMember.java
@@ -11,16 +11,22 @@ package com.redhat.qute.jdt.internal.template;
 
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
 
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.jdt.core.Flags;
+import org.eclipse.jdt.core.IClasspathEntry;
 import org.eclipse.jdt.core.ICompilationUnit;
 import org.eclipse.jdt.core.IField;
+import org.eclipse.jdt.core.IJavaElement;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.IPackageFragment;
 import org.eclipse.jdt.core.IPackageFragmentRoot;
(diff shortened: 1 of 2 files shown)


qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/template/TemplateGenerateMissingJavaMember.java
```diff
@@ -378,9 +384,25 @@ private static WorkspaceEdit createNewTemplateExtensionFile(GenerateMissingJavaM
 
 		IPackageFragment destPackage = null;
 		try {
+			IPath sourceFilesPath = project.getResource().getProjectRelativePath().append("src/main/java");
```
angelozerr, line 387: ²I wonder ifyou use IJavaProject#getRawClasspath should be easier? The Eclipse IDE wizard of new class initialize the source folder 

image


with this code

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/7caebb6cbb2167d1b5e33f4f8ba31273c47f75a5/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/wizards/NewContainerWizardPage.java#L251

datho7561: Unfortunately, this has the issue that it sometimes returns `src/main/resources` instead of `src/main/java`

angelozerr: After playing with the .classpath file, the Eclipse IDE wizard shows src/main/resources when src/main/resources is declared in the classpath before the src/main/java but is it a commons usecase?

fbricon: well you can iterate over javaProject.getRawClasspath(), filter on CPE_SOURCE, and return the source where  path == src/main/java

datho7561:
Okay. It should work now: it should avoid generating into a "generated sources" folder, test folder, and resource folder, with both maven and gradle, and in cases where you change where the source folder is.

(However, Qute template support is buggy in the last case since some features hard code `src/main/java`)

angelozerr:
The code looks very good since you don(t hard code src/main/java. Nice!

> (However, Qute template support is buggy in the last case since some features hard code src/main/java)

Sorry I don't understand what you mean?

Test alrady exists and you update the order of classpath entries, very good idea.

datho7561:
The code lens for type safe templates also points to the wrong folder when the resources folder is `src/resources`.

angelozerr:
Oh I see, the folder is hard coded at https://github.com/redhat-developer/quarkus-ls/blob/e6114a12e3c4c169c91424c6b75eacabd611a0cf/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/utils/JDTQuteProjectUtils.java#L42

We need to improve that

datho7561:
@angelozerr do you think this is okay to merge?

angelozerr:
> @angelozerr do you think this is okay to merge?

Yes but please create issue for codelens and other feature which doesn't work like you have noticed.

PR #835, opened 2023-03-29
{if !inject:beanName}{/if} is marked as invalid

### LINKED ISSUE #828: `{if !inject:beanName}{/if}` is marked as invalid
Open https://github.com/FroMage/quarkus-renarde-todo/blob/7784cba6de7f31e937726eec6725df3c95eca1f6/src/main/resources/templates/main.html#L31 in vscode-quarkus, and it will be marked with the warning "No namespace resolver found for: `!inject`". However, this code runs fine. It would be nice if we didn't mark it as invalid.

angelozerr:
{if !inject:beanName}{/if} is marked as invalid

Fixes #828

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/expression/NamespacePart.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/expression/NamespacePart.java
@@ -13,6 +13,9 @@ package com.redhat.qute.parser.expression;
 
 import com.redhat.qute.parser.expression.Parts.PartKind;
 import com.redhat.qute.parser.template.ASTVisitor;
+import com.redhat.qute.parser.template.Parameter;
+import com.redhat.qute.parser.template.Section;
+import com.redhat.qute.parser.template.SectionKind;
 
 /**
  * Namespace part.
@@ -28,6 +31,8 @@ public class NamespacePart extends Part {
 
 	public static final String DATA_NAMESPACE = "data";
 
+	private int startName = -1;
+	
 	public NamespacePart(int start, int end) {
 		super(start, end);
 	}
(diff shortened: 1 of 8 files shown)


angelozerr:
> But there is a problem with completion.

It should be fixed now.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/expression/NamespacePart.java
```diff
@@ -36,6 +41,27 @@ public NamespacePart(int start, int end) {
...
+		startName = super.getStartName();
+		Parameter parameter = getOwnerParameter();
+		if (parameter != null) {
+			Section section = parameter.getOwnerSection();
+			if (section != null && section.getSectionKind() == SectionKind.IF) {
+				String text = getOwnerTemplate().getText();
+				if (text.charAt(startName) == '!') {
+					// ex : !inject:
+					// !inject --> inject
+					startName++;
```
datho7561, line 59: You can have a space after the `!` and it will still work.

eg.

```qute-html
{#if  ! inject:vertxRequest.getParam('sandwich')??}
no sandwich ;-;
{/}
```

angelozerr: Could we see this usecase in an another PR because this usecase is used in renarde and I would like to integrate this fix in the release. We have supported this usacse for object part before and I copy/paste this code for namespace part.

datho7561: Okay. I opened https://github.com/redhat-developer/quarkus-ls/issues/839 to address this issue. It doesn't make sense to do the same fix for object part, since `{!` marks the start of a comment.

datho7561 (approved):
This works okay for me, and I filed a bug to improve the corner case I ran into.

PR #846, opened 2023-04-03
Remove unexpected nested-content parameter inserted in Quick Fix

### LINKED ISSUE #841: Quick fix to insert all missing attributes for user tag in Qute file adds a `nested-content` attribute
Let's say you have a user tag: (this is one from quarkus-renarde-todo)

`formElement.html`:
```qute-html
<label class="form-label" for="{name}">{label}</label>
<div class="mb-3 {class??}">
    {nested-content}
{#ifError name}
    <span class="invalid-feedback">​{#error name/}</span>​
{/ifError}
</div>
```

and you reference it from another Qute template like this:

```qute-html
{#formElement}
  <h3>Hello!</h3>
{/}
```

Then there is a quick fix that appears on `#formElement` to add all the missing attributes to the user tag. However, this quick fix adds an attribute called `nested-content`. However, `nested-content` [is a special key that gets replace with the content of the user tag](https://quarkus.io/guides/qute-reference#user_tags) (in this case `<h3>Hello!</h3>`). It doesn't makes sense to add it as an attribute with the quick fix.

JessicaJHee:
Fixes #841

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/tags/UserTag.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/tags/UserTag.java
@@ -174,6 +174,7 @@ public abstract class UserTag extends Snippet {
 		return getParameters().stream()
 				.filter(UserTagParameter::isRequired)
 				.map(UserTagParameter::getName)
+				.filter(paramName -> !paramName.equals(UserTagUtils.NESTED_CONTENT_OBJECT_PART_NAME))
 				.collect(Collectors.toList());
 	}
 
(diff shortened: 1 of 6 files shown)


datho7561:
Looks pretty good, but there is a related corner case:

If you have a user tag `myTag.html`:

```
{it}
```

and you reference it:
```
{#myTag}
{/myTag}
```

- The old behaviour was to generate `{#myTag it="it"}`
- The new behaviour (with this PR) is that the code action does nothing and the warning doesn't go away when you apply the code action
- The correct behaviour is {#myTag ""}, since `it` refers to the first parameter without a name (see https://quarkus.io/guides/qute-reference#user_tags)

I'm wondering if you want to address this in this PR or should I file an issue so that someone will eventually get to it?

JessicaJHee:
> * The correct behaviour is {#myTag ""}, since `it` refers to the first parameter without a name (see https://quarkus.io/guides/qute-reference#user_tags)

Thank you for pointing this case out! I've made some changes and added a test case, please let me know if I've understood the problem correctly.

datho7561:
PR seems to be good, but there are test failures: it seems like adding a new user tag to one of the test projects affected a test that counts the number of completion items.

datho7561 (approved):
Looks good. Thanks, Jessica!

PR #862, opened 2023-04-20
Ignore 'UndefinedObject' + 'UndefinedNamespace' problem code action should appear only if LSP client can support update configuration.

angelozerr:
Ignore 'UndefinedObject' problem code action should appear only if LSP client can support update configuration.

In IJ Quarkus, it doesn't support the capability to update configuration on client side. In otherwise the `Ignore 'UndefinedObject' problem` should not appear:

image

This PR checks that LSP client can support the command which update LSP client settings.


fbricon (approved):

angelozerr:
> Once the tests are fixed, I think this will be good to merge

Indeed I need to fix tests.

angelozerr:
@datho7561 tests should be fixed and I missed to do the same thing for UndefinedNamespace.

angelozerr:
@datho7561 it seems that there is a problem with snapshot https://github.com/redhat-developer/quarkus-ls/actions/runs/4757288906/jobs/8453919530#step:7:443

datho7561:
> it seems that there is a problem with snapshot

Right, we need to point to lsp4mp `0.8.0-SNAPSHOT`, since the snapshots for `0.7.0` are deleted after a certain amount of time.

datho7561 (approved):
This PR is good. Thanks, Angelo! I'll raise a quick PR to fix the issue with the snapshots

PR #871, opened 2023-05-22
Don't implement by default custom language client API

### LINKED ISSUE #870: Don't implement by default custom language client API
When a LSP client consumes the last version of Qute LS, it doesn't know which new API methods that it needs to implement because all API are implemented by default which return a CompletableFuture.completableFuture(null).

We need to remove default implementation to force all LSP client (like IJ Quarkus, JBoss Tools Quarkus) to implement new methods.

//cc @sbouchet

angelozerr:
Don't implement by default custom language client API

Fixes #870

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/QuteLanguageServer.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/QuteLanguageServer.java
@@ -61,7 +61,6 @@ import com.redhat.qute.ls.api.QuteLanguageClientAPI;
 import com.redhat.qute.ls.api.QuteLanguageServerAPI;
 import com.redhat.qute.ls.api.QuteProjectInfoProvider;
 import com.redhat.qute.ls.api.QuteResolvedJavaTypeProvider;
-import com.redhat.qute.ls.api.QuteTemplateJavaTextEditProvider;
 import com.redhat.qute.ls.api.QuteTemplateProvider;
 import com.redhat.qute.ls.api.QuteUserTagProvider;
 import com.redhat.qute.ls.commons.ParentProcessWatcher.ProcessLanguageServer;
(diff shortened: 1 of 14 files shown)


rgrunber:
I sort of wonder whether this is just swapping one issue for another. Yes, this means that when clients update they'll more easily see what they have to implement, but are the methods that were implemented as default mandatory ?

If a method is required for a client to work with that version of the language server, then yes it should not be marked as default with `return CompletableFuture.completedFuture(null);`, but if it isn't mandatory, then leaving it empty creates potentially extra work.

@datho7561 thoughts ?

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/QuteLanguageServer.java
```diff
@@ -89,7 +88,7 @@
  */
 public class QuteLanguageServer implements LanguageServer, ProcessLanguageServer, QuteLanguageServerAPI,
 		QuteProjectInfoProvider, QuteJavaTypesProvider, QuteResolvedJavaTypeProvider, QuteJavaDefinitionProvider,
-		QuteDataModelProjectProvider, QuteUserTagProvider, QuteTemplateJavaTextEditProvider, QuteJavadocProvider,
```
datho7561, line 92: Any reason why this interface is no longer needed?

datho7561 (changes requested):

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/QuteLanguageServer.java
angelozerr: It was an error, QuteTemplateJavaTextEditProvider is a method which must b eimplemented on client side and not on server side. As QuteTemplateJavaTextEditProvider has been implemented with default, we don't see this problem.

It is again a good point to remove default implementation.

PR #874, opened 2023-05-25
Update eclipse.jdt.ls to 1.24.0-SNAPSHOT

rgrunber:
See https://github.com/eclipse/eclipse.jdt.ls/pull/2666 . m2e 2.3.0 snapshot repo is gone so the old JDT-LS TP won't build. Need to use a newer one.


datho7561 (approved):
Looks good to me!

PR #895, opened 2023-07-05
Update lsp4mp to 0.8.0.

datho7561:
This is needed to fix the build, since the 0.8.0-SNAPSHOT have been automatically deleted.


rgrunber:
https://repo.eclipse.org/content/repositories/lsp4mp/org/eclipse/lsp4mp/org.eclipse.lsp4mp.ls/0.9.0-SNAPSHOT/ also doesn't exist since a build was never done.

datho7561:
Okay, I'll update this PR to use the released version instead of the SNAPSHOT

datho7561:
@rgrunber okay it's updated

rgrunber (approved):
It builds again, but the verification tests are running into OOM, so should probably have a look at that.

PR #898, opened 2023-07-11
Use DetectVMInstallationsJob.disabled=true property from JDT Debug

rgrunber:
- Automatic JVM detection in JDT Debug seems to have introduced a regression in JDT indexing causing java.lang.OutOfMemory
- Use Eclipse Temurin for JDK distribution.

See https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/231 for (I believe) the root cause.

Note : The property is just a workaround, and will only apply for the tests, so that they can run. I really suspect this would hit people at runtime. I've imported quarkus projects in vscode-quarkus (with vscode-java) and seen the OOM error. :neutral_face:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,11 +16,11 @@ jobs:
         with:
           path: ~/.m2/wrapper
           key: maven-wrapper-${{ hashFiles('**/mvnw') }}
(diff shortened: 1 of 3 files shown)


datho7561 (approved):
Looks good and works locally for me. Thanks, Roland!

PR #903, opened 2023-07-20
Fix NPE with Qute resolve completion when data is not filled.

angelozerr:
Fix NPE with Qute resolve completion when data is not filled.


angelozerr:
As it is a simple fix and it it the same issue than https://github.com/eclipse/lsp4mp/pull/412, I merge the PR.

PR #905, opened 2023-07-20
Support ? in parameter let section

### LINKED ISSUE #904: Support ? in parameter let section
If you open https://github.com/ia3andy/quarkus-blast/blob/aa969f34616ec63be0268fee20a40097d2ec1ba4/src/main/resources/templates/layout/page.html#L7 you will have an error which says that name is required.

image

It is because the user tag defines name https://github.com/quarkiverse/quarkus-web-bundler/blob/3603361f4db3f6d45f90de8b6667505afc4218af/runtime/src/main/resources/templates/tags/bundleStyle.html#L2

But name is defined here https://github.com/quarkiverse/quarkus-web-bundler/blob/3603361f4db3f6d45f90de8b6667505afc4218af/runtime/src/main/resources/templates/tags/bundleStyle.html#L1 and uses `?`. As we don't support `?` the let parameter is `name?` instead of `name` and in this case we consider that name is a required parameter instead of an optional parameter.

angelozerr:
Support ? in parameter let section

Fixes #904

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/template/Parameter.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/parser/template/Parameter.java
@@ -22,7 +22,6 @@ public class Parameter extends Node implements JavaTypeInfoProvider {
 	private int startName;
 
 	private int endName;
-
 	private int startValue = -1;
 
 	private int endValue = -1;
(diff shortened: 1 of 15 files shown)


angelozerr:
This PR should fix the error with name but you will notice that https://github.com/ia3andy/quarkus-blast/blob/aa969f34616ec63be0268fee20a40097d2ec1ba4/src/main/resources/templates/layout/page.html#L7 will have again a warning which says that bundle must b edeclared.

I have seen that with @ia3andy and it will fix the problem by setting bundle with @Global annotation in order to we can detect that bundle must be injected in the template data model.

angelozerr:
CI build is working again

datho7561 (approved):
Looks good. Thanks, Angelo!

angelozerr:
Thanks @datho7561 for your review!

PR #909, opened 2023-07-21
Don't report error when javaType is null

### LINKED ISSUE #908: Don't report error when javaType is null
Create a user tag (html file) `src/main/resources/templates/tags/mytag.html` with the following content:

`{foo.or(10)}`

It will report the error:

image

This error should not exists.

This usecase has been discovered in https://github.com/ia3andy/quarkus-blast/blob/aa969f34616ec63be0268fee20a40097d2ec1ba4/src/main/resources/templates/tags/user.html#L6

//cc @ia3andy

angelozerr:
Don't report error when javaType is null

Fixes #908


JessicaJHee (approved):
works as expected, LGTM!

angelozerr:
Thanks so much @JessicaJHee !

PR #918, opened 2023-07-27
Support itemDefaults for Qute parameter declaration

### LINKED ISSUE #900: Support itemDefaults/textEditor for Qute parameter declaration
Completion for Java classes / packages in parameter declaration can return a lot of Java classes completionitems which have the same text edit.

It should be nice to support itemDefaults/textEdit, see https://github.com/redhat-developer/quarkus-ls/blob/master/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForParameterDeclaration.java#L94

JessicaJHee:
Fixes #900 
* moves `editRange` and `insertTextFormat` to `itemDefaults` when supported

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForParameterDeclaration.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/completions/QuteCompletionsForParameterDeclaration.java
@@ -16,6 +16,7 @@ import java.util.List;
 import java.util.concurrent.CompletableFuture;
 
 import org.eclipse.lsp4j.CompletionItem;
+import org.eclipse.lsp4j.CompletionItemDefaults;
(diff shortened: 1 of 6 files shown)


datho7561:
I looked through the extension trace, and this doesn't seem to be working for me. (I only see `itemDefaults` in the trace when the client sends its capabilities to the server) Maybe I'm testing the wrong thing, though. This was the example I was trying:

```
{@java.util.List<String> strings}
{strings.|}
```

JessicaJHee:
@datho7561 the case I covered in this PR is:
```
{@ja|
```

which I think is the case Angelo is referring to in the issue if I understand correctly?

datho7561:
Yes, my bad, I read it incorrectly. I'll test it out again.

datho7561 (approved):
Looks good to me. Thanks, Jessica!

PR #921, opened 2023-08-03
Changelog for 0.16.0

JessicaJHee:


angelozerr:
Please update pom.xml version too

PR #922, opened 2023-08-08
Upversion to 0.17.0

datho7561:


JessicaJHee (approved):
LGTM :+1:

datho7561:
Good catch, I'll investigate. There should have been a lsp4mp snapshot build, but it looks like the build pipeline failed. I'll rerun it, then rerun this test suite.

datho7561:
I've run it twice and the tests have failed both times. I'll investigate if I can reproduce the test failure.

datho7561:
I can reproduce the test failure locally, but it only happens occasionally.

datho7561:
I reran the lsp4mp snapshot build and skipped tests to ensure it released (I confirmed a while back that the tests for lsp4mp were passing locally). Hopefully the build works here.

PR #936, opened 2023-09-14
Update eclipse.jdt.ls to 1.28.0-SNAPSHOT

rgrunber:
- React to org.apache.commons.io -> org.apache.commons.commons-io

This will require a snapshot build of https://github.com/eclipse/lsp4mp/pull/422 .


quarkus.jdt.ext/pom.xml
```diff
@@ -14,7 +14,7 @@
     <tycho.extras.version>${tycho.version}</tycho.extras.version>
```
rgrunber, line 14: I might as well bump Tycho to 3.0.5 and update the Maven wrapper to use 3.9.3.

Issue #939, opened 2023-09-16
Properties should be available recursively


gbourant:
Assume you have 3 tags:

1. `inputLabelText.html`
```html
       {#inputLabel /}
       {#inputText  /}
```

2. `inputLabel.html`
```html
<label {#if for??}for="{for}"{/if}>
        {title??}
</label>
```
3. `inputText.html`
```html
<input {#if value??}value="{value}"{/if} />
```

So when the `inputLabelText.html` is used `{#inputLabelText /}` the properties of the other tags are not available for auto completion.

angelozerr:
@gbourant I moved your issue in quarkus-ls github project since your 2 issues comes from Qute LS.

mkouba:
> @mkouba is it an expected behavior?

I'm not quite sure what behavior do you mean? 

I don't think that args/parameters of a user tag should be validated at all because tags that are not executed in isolation can access data from the parent context. 

Note that in [Quarkus 3.5 there's a breaking change](https://github.com/quarkusio/quarkus/wiki/Migration-Guide-3.5#qute) - a user tag is now executed as an isolated template by default (unlike in previous versions).

PR #940, opened 2023-09-19
Qute support for multi module project

### LINKED ISSUE #930: Qute support for multi module project
When project contains multi-module with some projects which are Qute project, it doesn't work, because today we consider that a workspace folder is a project which is wrong, because a workspace folder could contains several maven module which could be a Qute project.

To fix that we need to:

 * not use workspace folder to load Qute project
 * having on JDT / IJ side a new services:
   * java/qute/projects request -> return all Qute projects from rootUri
   * java/qute/projectAdded / java/qute/projectRemoved notifications to notify the language server that a new Qute project has been added/removed from the workspace.

angelozerr:
Qute support for multi module project

Fixes #930

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/api/QuteProjectInfoProvider.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/api/QuteProjectInfoProvider.java
@@ -11,8 +11,10 @@
 *******************************************************************************/
 package com.redhat.qute.ls.api;
 
+import java.util.Collection;
 import java.util.concurrent.CompletableFuture;
 
+import org.eclipse.lsp4j.jsonrpc.services.JsonNotification;
 import org.eclipse.lsp4j.jsonrpc.services.JsonRequest;
 
 import com.redhat.qute.commons.ProjectInfo;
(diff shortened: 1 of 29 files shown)


angelozerr:
This PR requires vscode-quarkus https://github.com/redhat-developer/vscode-quarkus/pull/627 PR if you wish to test it in vscode-quarkus

angelozerr:
To test this PR, open https://github.com/gbourant/multi-module-template-test. The use case is that [index.html](https://github.com/gbourant/multi-module-template-test/blob/master/project-b/src/main/resources/templates/index.html) from [project-b](https://github.com/gbourant/multi-module-template-test/blob/master/project-b) include the template [root.html](https://github.com/gbourant/multi-module-template-test/blob/master/project-a/src/main/resources/templates/root.html) from the [project-a](https://github.com/gbourant/multi-module-template-test/blob/master/project-a) (where [project-b have a dependency to project-a](https://github.com/gbourant/multi-module-template-test/blob/a2fd9b8412b1b362073f1a7428d647772157e590/project-b/pom.xml#L15)).

With this PR:

 * include root doesn't appear as an error
 * Ctrl+click on root jump to the root.html file
 * completion on #include should show root.

image

angelozerr:
I need to clean my code and write tests but @fbricon you can play with the PR (please see the limitations that I have listed)

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/ls/api/QuteProjectInfoProvider.java
```diff
@@ -26,6 +28,38 @@
  */
 public interface QuteProjectInfoProvider {
 
+	/**
+	 * Returns all Qute projects from the workspace.
+	 * 
+	 * @return all Qute projects from the workspace.
+	 */
+	@JsonRequest("qute/template/projects")
```
fbricon, line 36: why "qute/template/projects" instead of "qute/projects"? Project changes will affect both template and java support

angelozerr: I did that because all qute/template/* service are managed with `QuteSupportForTemplate`. More  "qute/template/projects" are used by only Qute template support. Java files are managed by Java external component IJ.Eclipseide

fbricon:
seems to for for includes, but not for user tags. If you create a https://github.com/gbourant/multi-module-template-test/tree/master/project-a/src/main/resources/templates /tags/link.html, then you have no completion, nor navigation for {#link} in https://github.com/gbourant/multi-module-template-test/blob/master/project-b/src/main/resources/templates/index.html

fbricon:
actually if you put any user tags in project a, then all user tags in project b disappear from rendering. @mkouba is that intended? I then don't understand how Renarde works, since it also contributes user tags

angelozerr:
> actually if you put any user tags in project a, then all user tags in project b disappear from rendering. @mkouba is that intended? I then don't understand how Renarde works, since it also contributes user tags

@gbourant we are using your sample (thanks for that), have you tried this usecase?

gbourant:
> > actually if you put any user tags in project a, then all user tags in project b disappear from rendering. @mkouba is that intended? I then don't understand how Renarde works, since it also contributes user tags
> 
> @gbourant we are using your sample (thanks for that), have you tried this usecase?

I can confirm that this happened to me too. 

That's why i moved the tags to project-b when i reported the [Properties should be available recursively #939](https://github.com/redhat-developer/quarkus-ls/issues/939)

angelozerr:
The PR contains now tests and the code should be cleaned, @fbricon please review it.

fbricon:
thanks @angelozerr

PR #952, opened 2024-02-05
build: generate unique p2 repo for quarkus & qute

fbricon:
Signed-off-by: Fred Bricon <fbricon@gmail.com>


rgrunber:
Could we create an identical structure for maven repos ? I think as long as it follows ${gid}/${aid}/${version}/ it should work. We could create a tag called "maven" so it's available as "https://github.com/redhat-developer/vscode-java/releases/download/maven/..." and always update it with the artifacts.

rgrunber:
I have included and updated this PR in https://github.com/redhat-developer/quarkus-ls/pull/1039 .

PR #967, opened 2024-07-09
fix: Qute highlight errors due to formatting

### LINKED ISSUE #966: Qute highlight errors due to formatting
The following image is the file `angelozerr.html`. As you can see, based on the formatting it highlights that the code has some errors.

image

To remove the error highlight i did the following, i changed

`ifElementHasError(errorFields, errorKeys)}` with
`ifElementHasError(errorFields,errorKeys) }`, I removed the space after the comma and added a space before the last curly bracket.

`getErrorMessages(errorFields,errorKeys)}` with
`getErrorMessages(errorFields, errorKeys)}` (added a space after the command and added a space before the last curly bracket)

After you fix those two errors you get a new one as shown in the image. So i changed `{errorMessage}` to `{ errorMessage}` (added a space after the first curly bracket)

image

Reproducer: https://github.com/gbourant/qute-bugs

angelozerr:
fix: Qute highlight errors due to formatting

Fixes #966

This PR fixes the 2 problem of #966 issue:

 * avoid having false positive error (when a method in for or each section is used with parameters, those parameters was validated to check if the parameter type was iterable):
image
 * provides thacapability to write spaces inside methode parameters.

@fbricon I had to do a big change in parameter parser. Hopefully we had alo of tests about that, I have written a lot of tests, but I hope I have not boken something.

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
@@ -52,6 +52,7 @@ import com.redhat.qute.parser.expression.Parts.PartKind;
 import com.redhat.qute.parser.expression.PropertyPart;
 import com.redhat.qute.parser.template.CaseOperator;
 import com.redhat.qute.parser.template.Expression;
+import com.redhat.qute.parser.template.ExpressionParameter;
 import com.redhat.qute.parser.template.JavaTypeInfoProvider;
 import com.redhat.qute.parser.template.LiteralSupport;
 import com.redhat.qute.parser.template.Node;
@@ -256,6 +257,11 @@ class QuteDiagnostics {
 								switch (section.getSectionKind()) {
 									case FOR:
 									case EACH:
+										Part lastPart = expression.getLastPart();
+										if (result != null) {
+											result = validateIterable(lastPart, section, result, result.getSignature(),
+													diagnostics);
+										}
 										String alias = ((LoopSection) section).getAlias();
 										currentContext.put(alias, result);
 										break;
@@ -461,10 +467,11 @@ class QuteDiagnostics {
 	 * Validate #include section.
 	 *
 	 * @param includeSection the include section
-	 * @param project 
+	 * @param project
 	 * @param diagnostics    the diagnostics to fill.
 	 */
-	private static void validateIncludeSection(IncludeSection includeSection, QuteProject project, List<Diagnostic> diagnostics) {
+	private static void validateIncludeSection(IncludeSection includeSection, QuteProject project,
+			List<Diagnostic> diagnostics) {
 		Parameter templateParameter = includeSection.getTemplateParameter();
 		if (templateParameter != null) {
 			// Validate template id, only if project exists.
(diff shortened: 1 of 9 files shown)


qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
```diff
@@ -232,7 +233,8 @@ private void validateDataModel(Node parent, Template template, QuteValidationSet
 					// validate expression parameters
 					boolean checkValidOperator = section.getSectionKind() == SectionKind.IF;
 					boolean shouldBeAnOperator = false;
-					for (Parameter parameter : parameters) {
+					for (int i = 0; i < parameters.size(); i++) {
```
fbricon, line 236: why? seems you're not using the index, so you can keep the enhanced for-loop

angelozerr: Good catch, fixed.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
```diff
@@ -1232,26 +1238,24 @@ private ResolvedJavaTypeInfo validateJavaTypePart(Part part, Section ownerSectio
...
-				return null;
-			}
+		// The expression is declared inside an iterable section like #for, #each.
+		// Ex: {#for item in items}
+		if (!resolvedJavaType.isIterable() && !resolvedJavaType.isInteger()) {
+			// The Java class is not an iterable class like
+			// - java.util.List
+			// - object array
+			// - integer
+			String expression = part.getParent().getContent();
```
fbricon, line 1253: you previously had a null-check on part. is it not necessary anymore?

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
```diff
@@ -1232,26 +1238,24 @@ private ResolvedJavaTypeInfo validateJavaTypePart(Part part, Section ownerSectio
...
-				Range range = QutePositionUtility.createRange(part);
-				Diagnostic diagnostic = createDiagnostic(range, DiagnosticSeverity.Error, QuteErrorCode.IterationError,
-						expression, javaTypeToResolve);
-				diagnostics.add(diagnostic);
-				return null;
-			}
+		// The expression is declared inside an iterable section like #for, #each.
+		// Ex: {#for item in items}
+		if (!resolvedJavaType.isIterable() && !resolvedJavaType.isInteger()) {
+			// The Java class is not an iterable class like
```
fbricon, line 1249: The Java class is neither an integer nor an iterable class like

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
angelozerr: The iterable check is done now only when last part of a #for or #each section is scanned. In this case we are sure that part is never null and the section is an iterable section.

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java

fbricon:
thanks @angelozerr!

PR #968, opened 2024-07-10
fix: separate user tag parameters inlay hint

angelozerr:
fix: separate user tag parameters inlay hint

This PR uses the paddingRight property of inlay hint to separate user tag parameters.

Here the result of vscode:

image

We will have to support this property in LSP4IJ.


fbricon:
Thanks @angelozerr

PR #989, opened 2024-08-29
Qute - incorrect behavior for nested template records

### LINKED ISSUE #987: Qute - incorrect behavior for nested template records
Expected: a template record that is not annotated with `@CheckedTemplate` should be treated as if it was annotated with `@CheckedTemplate` with default values. That is a base path for a nested record should be the simple name of the enclosing class.

Current: IDE expects the base path as if it was a top-level class.

In the following example the template path should be `/src/main/resources/templates/HelloResource/Hello.html`.

image

angelozerr:
Qute - incorrect behavior for nested template records

Fixes #987


angelozerr:
Here the result to fix the Qute record support:

image

@mkouba is it OK for you? Do you see some another usecases?

mkouba:
> Here the result to fix the Qute record support:
> 
> image
> 
> @mkouba is it OK for you? Do you see some another usecases?

Looks much better ;-). Thanks! I don't see any other use cases...

PR #995, opened 2024-09-11
Update eclipse.jdt.ls to 1.40.0-SNAPSHOT

### LINKED ISSUE #994: i build got deleted; need to update
for future reference, it seems the cron job runs around 10 AM EST

datho7561:
Requires https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3271/files to be merged in jdt.ls to work properly, so we should wait to merge this.

Fixes #994


rgrunber:
LSP4MP probably needs the same : See https://github.com/eclipse/lsp4mp/pull/454 .

PR #1006, opened 2024-10-17
Improve the Target platform for JDT Extension components.

rgrunber:
- Use the JDT-LS repository, which now contains all dependencies to build the target platform
- Include o.e.platform.feature.group for the test runtime


datho7561:
It seems that this PR is needed to build quarkus-ls locally, so I think that we should merge it

PR #1017, opened 2024-10-28
Roq DataMapping support

angelozerr:
Roq DataMapping support

This PR should fix false positive

image

Here a demo:

DataMappingDemo

//cc @ia3andy


fbricon:
Is it really roq specific? Isn't qute able to inject ANY CDI bean into template? cc. @mkouba @ia3andy

ia3andy:
> Is it really roq specific? Isn't qute able to inject ANY CDI bean into a template? cc. @mkouba @ia3andy

yes, but we don't have a way to inform the plugin parser about Synthetic Beans AFAIK (which is the case here)

fbricon:
maybe quarkus could generate a list of synthetic beans in a metadata file, at build time, but that's another story.

PR #1019, opened 2024-11-04
Fix the qute.jdt tests

datho7561:
We were expecting one more newline than we actually got.


angelozerr:
If CI build is working lets merge.

angelozerr:
Thanks so much @datho7561 !

PR #1038, opened 2024-12-06
Copy over eclipse/lsp4mp#480 into qute.jdt

datho7561:
Use the secondary-type search for types in the project, and if that fails, use the non-secondary type search for types.


datho7561:
There is a compilation error. Maybe I should get more sleep or use an actual IDE to make these changes :facepalm:

datho7561:
This causes more test failures locally for me, which is concerning to say the least

datho7561:
I don't think this will improve test stability

rgrunber (approved):
I'm fine with merging for the sake of consistency.

PR #1044, opened 2025-04-09
Update snakeyaml to 2.4

datho7561:


rgrunber:
You need to update https://github.com/redhat-developer/quarkus-ls/blob/1e9595e4a6a8b433e91cb04ae517ec0c109206bf/quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/META-INF/MANIFEST.MF#L25 & https://github.com/redhat-developer/quarkus-ls/blob/1e9595e4a6a8b433e91cb04ae517ec0c109206bf/quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/build.properties#L6

rgrunber (approved):
Works for me. I tested in an `application.yaml`. Seems to work same as before for completion.

angelozerr:
> Works for me. I tested in an `application.yaml`. Seems to work same as before for completion.

@rgrunber snake yaml is not used for completion in application.yaml (to support that we generate a json schem1 from the properties computed in the classpath)

Snake yaml is used when you save your application.yaml and store the properties (as flat) in memory.

After that when you hover a Java @ConfigProperty which defines the property, it should show the value declari3d in the saved application.yaml

datho7561:
I double checked using the steps you provided @angelozerr , and it seems to be working properly.

PR #1049, opened 2025-04-11
Support for Qute Integer Numbers

angelozerr:
Fixes https://github.com/redhat-developer/vscode-quarkus/issues/974

Here a sample which doesn't report false positive error:

```
{@int foo}
{@int[] bar}

{foo mod 10 + 58 + foo.mod(10) - 9999 minus 7}

{bar.0 + 10 - 7}

{#for item in bar}
  {item + 7 + item_count - 7 minus 88}
{/for}
```

The fix is done to support correctly matchNames used in TemplateExtension.

This `+` is managed with https://github.com/quarkusio/quarkus/blob/91c8a5b95d900917039190985e9ed84e1af75eba/extensions/qute/runtime/src/main/java/io/quarkus/qute/runtime/extensions/NumberTemplateExtensions.java#L17

By fixing matchName support it allows to support features like diagnostic, completion, hover, definition, etc

Here a demo

QuteNumebrSupport

### DIFF
--- a/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
+++ b/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
@@ -1136,7 +1136,6 @@ class QuteDiagnostics {
 
 		boolean matchVirtualMethod = result.isMatchVirtualMethod();
 		if (!matchVirtualMethod) {
-
 			Range range = QutePositionUtility.createRange(methodPart);
 			Diagnostic diagnostic = createDiagnostic(range, DiagnosticSeverity.Error,
 					QuteErrorCode.InvalidVirtualMethod, //
@@ -1146,8 +1145,9 @@ class QuteDiagnostics {
 			return null;
 		}
 
+		List<JavaParameterInfo> applicableParamters = method.getApplicableParameters();
 		if (methodPart.isInfixNotation()) {
-			int nbParameters = method.getParameters().size() - (method.isVirtual() ? 1 : 0);
+			int nbParameters = applicableParamters.size();
 			if (nbParameters != 1) {
 				// infix notation,
 				// ex: String#codePointCount cannot be used with infix notation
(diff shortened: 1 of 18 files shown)


angelozerr:
@datho7561 tests fail but it should not be related to this PR.

datho7561:
Sorry got distracted with some performance stuff, taking a look now

qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/QuteDiagnostics.java
```diff
@@ -1146,8 +1145,9 @@ private ResolvedJavaTypeInfo validateMethodPart(MethodPart methodPart, Section o
 			return null;
 		}
 
+		List<JavaParameterInfo> applicableParamters = method.getApplicableParameters();
```
datho7561, line 1148: applicableParamters -> applicableParameters

qute.ls/com.redhat.qute.ls/src/test/java/sandbox/TestItCountQute.java
```diff
@@ -0,0 +1,47 @@
+package sandbox;
```
datho7561, line 1: Maybe accidentally committed?

qute.ls/com.redhat.qute.ls/src/test/java/sandbox/it_count.qute.html
```diff
@@ -0,0 +1,3 @@
+{#each items}
```
datho7561, line 1: Maybe accidentally committed?

datho7561 (changes requested):
I found a few things that might clean up the PR, but otherwise the code looks good and works well.

qute.ls/com.redhat.qute.ls/src/test/java/sandbox/TestItCountQute.java
angelozerr: No I use those templates in sandbox to test qute with the qute engine.

datho7561: I see my bad.

datho7561 (approved):
Looks good. Thanks, Angelo!

PR #1051, opened 2025-04-13
Improve Java data model cache eviction

### LINKED ISSUE #444: Improve Java data model cache eviction
In Qute LS side, there is a Java data model cache for each classes, methods used in templates. 

When Qute template requires a Java data model (for validation, completiuon, etc) the first time it consumes extrenal component (ex : JDT Qute) to get methods, fields for used Java classes in the template and cache the result.

This cache is evicted as soon as a Java file is saved.

More the cache contains all template data model (ex : CheckedTemplate) of the Qute project which is computed by scanning sources and dependencies of the project.

It means that each time we save a java file, it will require the compute of:

 * all Java classes used for the template
 * scanning of java sources + dependencies to compute templates data model.

angelozerr:
Improve Java data model cache eviction

Fixes #444


angelozerr:
To test this PR:

Create 2 classes like this:

 * Animal.java
 * 
```java
package foo;

// Animal.java
public class Animal {
    protected String name;

    public Animal(String name) {
        this.name = name;
    }

    public void speak() {
        System.out.println("The animal makes a sound.");
    }

    public String getName() {
        return name;
    }
}
```
 * Dog.java
 
```java
package foo;

// Dog.java
public class Dog extends Animal {
    private String breed;

    public Dog(String name, String breed) {
        super(name);
        this.breed = breed;
    }

    @Override
    public void speak() {
        System.out.println(name + " barks!");
    }

    public String getBreed() {
        return breed;
    }
}

```

qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/internal/JavaDataModelListenerManager.java
```diff
@@ -151,16 +151,47 @@ public boolean visit(IResourceDelta delta) throws CoreException {
 			case IResource.FILE:
 				IFile file = (IFile) resource;
 				if (isJavaFile(file) && isFileContentChanged(delta)) {
-					// A Java file has been saved
-					JavaDataModelChangeEvent event = new JavaDataModelChangeEvent();
-					event.setProjectURIs(new HashSet<String>());
-					event.getProjectURIs().add(JDTQuteProjectUtils.getProjectURI(file.getProject()));
-					fireAsyncEvent(event);
+					// A Java file has been updated and saved
```
angelozerr, line 154: After some investigation, I wonder if it should be better to use ElementChangedEvent to track the update, delete, and create of Java file. Let me investigate that.

angelozerr: @rgrunber do you think it is possible to use JDT ElementChangedEvent to track change of Java file but only on save?

rgrunber: It should be possible. The workspace would broadcast a delta for any file that is changed. What do you mean by "save". Do you want to listen to any attempt to save the file, or any attempt to save when the file is dirty ?

angelozerr: I would like to avoid sending notification when user is updating the Java editor, but send a notification when user do a save.

angelozerr: I am using IResourceChangeListener for now.

datho7561 (approved):
I think this is a good improvement as-is. It works well for me; in the logs it goes from over 20 requests to resolvedJavaType to 1.

Looking in to a better way of detecting changes in Java files sounds like a good idea long term, and we should handle detecting file deletions as well. Given the severity of this performance problem when autosave is enabled (as you have described it to me), I think it's okay to merge this and continue to look for a better solution.

angelozerr:
@datho7561 I have started to implement my idea with ElementChangedEvent and it seems it is working.

I don't need to use IResourceChangeListener and it seems that ElementChangedEvent  is called only when ICompilationUnit is saved, deleted and created.

More I think it is better to use ElementChangedEvent because I need ICompilationUnit to get the qualified name in the sources.

My only problem is when ICompilationUnit is deleted, I cannot compute the qualified name.

rgrunber:
For me, it went from around ~50 `qute/template/resolvedJavaType` to about 3. :+1:

datho7561:
Angelo told me he started looking into why it was 3 instead of 1 as it was in the earlier change, so I think it's worth holding off on that, but this change dramatically improves things as is.

angelozerr:
> Angelo told me he started looking into why it was 3 instead of 1 as it was in the earlier change, so I think it's worth holding off on that, but this change dramatically improves things as is.

Please retry it should work now and you should have just one resolvedJavaType request and response

datho7561 (changes requested):
Deleting files no longer seems to work as of the latest change

angelozerr:
@datho7561 it should work now.

datho7561:
Realized I haven't run the tests yet and CI is failing on the quarkus.jdt tests, so I'm quickly running them locally

PR #1054, opened 2025-04-16
Run test suites as separate jobs

### LINKED ISSUE #1053: Set up "fail at end" for CI
Currently, if one of the CI steps fails, then the rest aren't run.

This is bad, because 2 of the test suites are flaky (quarkus.jdt and qute.jdt), and the flaky tests might hide legitimate failures in the other suites.

If would be nice if we always ran and reported the results of all the test suites.

There are a few ways we could accomplish this:
1. split up the job into 4 separate jobs, one for each subproject (I think I prefer this method since it allows them to be run concurrently)
2. use an if clause with `success() || failure()` after each test run
3. maybe other methods?

datho7561:
This lets be run in parallel. This means that we can check their failures individually, and we will probably get results sooner.

Fix #1053

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,33 +1,55 @@
 name: Tests
 on: [push, pull_request]
 jobs:
-  tests:
+  quarkus-jdt-tests:
     runs-on: ubuntu-latest
     steps:
       - name: Check out repository code
         uses: actions/checkout@v4
-      - name: Cache .m2 repository
-        uses: actions/cache@v4
-        with:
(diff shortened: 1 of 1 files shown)


datho7561:
I removed the .m2 and ./mvnw caching; I think those would still be helpful but also would take a bit of time to set up. Let me know if I should add them in this change.

angelozerr:
It is a very nice improvement! Now we can see that ls tests are working. Nice!

I think we have the same problem with lsp4mp, if yes could you do that please.

PR #1059, opened 2025-05-02
Do not generate property for `<clinit>`

datho7561:
- [x] Needs a test

Fixes redhat-developer/vscode-quarkus#1019


datho7561:
@fbricon this fix should work, but several lsp4mp releases got deleted off nexus due to the storage failure. So it will only build if you have those artifacts locally or build them.

CI is going to be failing until we can restore those releases / make new SNAPSHOTs / a z release / something else? In order to do that though, we'll need the lsp4mp Jenkins CI back up.

datho7561:
@angelozerr if you have time, could you please take a look?

For your information, it's intended behaviour that `IType.getChildren()` returns `<clinit>` as one of the `IMethod` for binary types. `<clinit>` is the method that the compiler generates to set the value of the static members.

quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus.test/src/main/java/com/redhat/microprofile/jdt/quarkus/QuarkusMavenProjectName.java
```diff
@@ -20,6 +20,6 @@ public class QuarkusMavenProjectName extends MicroProfileMavenProjectName {
 	public static String scheduler_quickstart = "scheduler-quickstart";
 	public static String scheduler_diagnostic = "scheduler-diagnostic";
 	public static final String quarkus_route = "quarkus-route";
-
+	public static final String clinit = "clinit";
```
angelozerr, line 23: It seems that you don't use this constant, right?

datho7561: I use it here https://github.com/redhat-developer/quarkus-ls/pull/1059/files#diff-5a589c56b38b81bf238303c13413a59cd85109cb5bd81389b2980f7d41852db8R298

angelozerr:
LGTP, thanks @datho7561 !

PR #1076, opened 2025-12-01
Smart iterable Qute completion support

angelozerr:
Smart iterable Qute completion support

Here a demo:

SmartListCompletionDemo2


angelozerr:
Thanks @datho7561 for your review!

PR #1080, opened 2025-12-03
Adapt to changes in jdt.ls

datho7561:
There was a method removed from `JavadocContentAccess2` that we were using. I've updated this project to use the new method.


datho7561:
Tests are timing out, not a great sign.

angelozerr:
> I adjusted the timeout to 15 minutes. It seems like most of the time is spent downloading dependencies. Maybe we should have a caching strategy? But we need to use `-U` to get the latest jdt.ls

It would be very nice if we could improve those tests 'which are flacky and takes long times). Any idea are welcome!

angelozerr:
Thanks so much @datho7561 !

Issue #1001, opened 2024-10-10
false-positive for `it` parameter in Qute template


datho7561:
See https://github.com/FroMage/quarkus-renarde-todo/blob/main/src/main/resources/templates/tags/user.html 

The validator treats `it` as a `java.lang.Object`, so it marks all the method invocations as errors.

It makes more sense to skip validating method invocations and field accesses on this object, since its actual type is not know.

angelozerr:
In this usecase , the user tag is called here https://github.com/FroMage/quarkus-renarde-todo/blob/d03d71f6b41b14b228fe8ad782b32800dcb9b6f0/src/main/resources/templates/main.html#L19

`{#user inject:user img=true size=20/}`

As I have implemented Qute closed template, my plan is to provide a smart inference to say that it must be an User object instead of Object only.

angelozerr:
See issue https://github.com/redhat-developer/quarkus-ls/issues/913

Issue #931, opened 2023-09-14
Qute LS freezes when LSP client doesn't support `workspace/didChangeWatchedFiles` and have no templates folder


angelozerr:
When LSP client cannot support `workspace/didChangeWatchedFiles`, the Qute LS uses FileSystemWatcher and if the project have no src/main/resources/templates, it blocks the language server https://github.com/redhat-developer/quarkus-ls/blob/2be29a3628711e829aa99fd8fcf9a67432a888c9/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/WatchDir.java#L102

We need to poll the existing of src/main/resources/templates in  Thread.

Please note that:

 * in vscode we have not the problem because vscode supports  `workspace/didChangeWatchedFiles`
 * in IJ I'm implementing  `workspace/didChangeWatchedFiles` https://github.com/redhat-developer/intellij-quarkus/pull/1163
 * in Eclipse IDE I created the issue to support  `workspace/didChangeWatchedFiles` https://github.com/eclipse/lsp4e/issues/807 //cc @sbouchet

PR #1088, opened 2025-12-31
Smart inference of Java type for user tag parameter

### LINKED ISSUE #913: Smart inference of Java type for user tag parameter
Given this user tag `src/main/resources/tags/myUser.html`:

```
Login: {user.login}
```

At this step, we cannot know what is user

Given a Qute template which consumes the user tag

```
{@org.acme.User userFromJava}
{#myUser user=userFromJava /}
```

At this step we can inject the `org.acme.User` Java type inside the `myUser.html` which will provide completion, validation, hover, defintion,etc

This feature is very important to support for instance https://github.com/FroMage/RivieraDEV-Quarkus/blob/main/src/main/resources/templates/tags/organisers.html

angelozerr:
Smart inference of Java type for user tag parameter

Fixes #913

It should also fix https://github.com/redhat-developer/quarkus-ls/issues/1001

This PR implements references for parameter of user tag:

image

and manages usages of parameter to guess the java type:

image

//cc @FroMage


FroMage:
So, where does it get the type info from?

angelozerr:
> So, where does it get the type info from?

Here a demo in order to you understand more:

UserTagDemoInferType

angelozerr:
In this PR I removed the watcher implementation with file system because today LSP4E, vscode and LSP4IJ supports `workspace/didChangeWatchedFiles`. It will fix https://github.com/redhat-developer/quarkus-ls/issues/931

PR #1095, opened 2026-01-20
Support for Roq Data files

angelozerr:
Support for Roq Data files

To implement this PR I am using https://github.com/myfear/ejq_substack_articles/tree/main/roq-author-website usecase.

To test the PR:

 * open  https://github.com/myfear/ejq_substack_articles/tree/main/roq-author-website usecase.
 * create a foo.html file in the content folder

```
{#for b in inject:books.list}
    {b.buy}
{/for}
```

you should have no error. Here a demo which show the support in action to write this content (completion, validation, hover and definition)

InjectDataDemo2

This PR tracks also changes of data files (stored in data folder https://github.com/myfear/ejq_substack_articles/tree/main/roq-author-website/data . It mean that if you create, delete or update yaml files, the validation of template must be retrrigered and completion etc should be refreshed also.

InjectDataDemo3

I need to:

 * improve hover of property coming from yaml files
 * improve inlay hint

Definition (ctrl+Click) open the file on the first line of the files and not where the property is defined. I will not support that in this PR but in a future PR because this PR is already big.


datho7561:
To test the code, I've been writing a test case, then debugging to look at the AST that's generated.

